PR #5, opened 2015-02-20
  Expose cors headers so clients subject to it can access it.

Natim:
Replace https://github.com/mozilla-services/readinglist/pull/129


cliquet/tests/testappviews.py
```diff
@@ -0,0 +1,11 @@
+from cliquet.resource import BaseResource, ResourceSchema, crud
```
ametaireau: I believe we should put that in a separate folder, exactly like we would do for a real app.

cliquet/tests/testappviews.py
```diff
@@ -0,0 +1,11 @@
...
+import colander
+
+
+class MushroomSchema(ResourceSchema):
+    name = colander.SchemaNode(colander.String())
+
+
+@crud()
+class Mushroom(BaseResource):
+    mapping = MushroomSchema()
```
leplatrem: Couldn't we factorize with this ? https://github.com/mozilla-services/cliquet/blob/exposed-cors-headers/cliquet/tests/resource/test_views.py#L16-L22

cliquet/tests/test_cors.py
```diff
@@ -0,0 +1,36 @@
+from .support import BaseWebTest, unittest
+
+
+def get_available_headers(headers):
+    access_control_headers = headers['Access-Control-Expose-Headers']
+    return [x.strip() for x in access_control_headers.split(',')]
+
+
+class CORSTest(BaseWebTest, unittest.TestCase):
+    def test_preflight_headers_for_default(self):
```
ametaireau: Use sentences in your functions.

Natim:
Updated.

Natim:
It fails with py34

cliquet/tests/test_cors.py
```diff
@@ -0,0 +1,34 @@
+import json
+from .support import BaseWebTest, unittest
+
+
+def get_available_headers(headers):
```
ametaireau: get_exposed_headers

PR #8, opened 2015-02-20
Postgres backend (fixes #7)

### LINKED ISSUE #7: Postgres backend
Revamp of https://github.com/mozilla-services/readinglist/issues/131

leplatrem:
Revamp of https://github.com/mozilla-services/readinglist/pull/122


cliquet/session/redis.py
```diff
@@ -43,7 +45,7 @@ def delete(self, key):
 
 def load_from_config(config):
     settings = config.registry.settings
-    host = settings.get('session_redis.host', 'localhost')
-    port = settings.get('session_redis.port', 6379)
-    db = settings.get('session_redis.db', 1)
-    return RedisSessionStorage(host=host, port=port, db=db)
+    uri = settings.get('cliquet.session_url', '')
```
Natim: Should it be ``redis://localhost:6349/1`` by default.

Natim: Or even db index `0`

README.rst
```diff
@@ -33,11 +36,66 @@ Once Redis is installed:
...
+are performed in memory.
+
+In configuration::
+
+    cliquet.storage_backend = cliquet.storage.redis
+
+*(Optional)* Instance location URI can be customized::
 
-    cliquet.storage_backend = cliquet.storage.memory
+    cliquet.storage_url = localhost:6379/0
```
Natim: I think we should add a scheme instead, something like: ``redis://localhost:6379/0``

leplatrem: +1

leplatrem:
@ametaireau @Natim r?

cliquet/session/redis.py
```diff
@@ -43,7 +45,10 @@ def delete(self, key):
...
+    uri = urlparse.urlparse(uri)
+
+    return RedisSessionStorage(host=uri.hostname or 'localhost',
```
Natim: I think it would be better to directely put this default into the ``uri = settings.get`` default value

cliquet/storage/redis.py
```diff
@@ -192,24 +193,22 @@ def get_all(self, resource, user_id, filters=None, sorting=None,
...
+        return records, count
 
 
 def load_from_config(config):
     settings = config.registry.settings
-    host = settings.get('redis.host', 'localhost')
-    port = settings.get('redis.port', 6379)
-    db = settings.get('redis.db', 0)
-    return Redis(host=host, port=port, db=db)
+    uri = settings.get('cliquet.storage_url', '')
```
Natim: same here.

cliquet/session/redis.py
leplatrem: But why ?

This way, if you don't specify the port in the URL, it uses the default for example.

Natim: You mean ``redis://my-host/`` should default to ``redis://my-host:6739/0`` well why not.

leplatrem: I don't know if it should, but it's the current behaviour. If we have some counter-arguments, I'm willing to change it :)

Natim: No you are right, I think it makes sense.

cliquet/session/redis.py
ametaireau: yes, we should default it here.

cliquet/session/redis.py
ametaireau: Having it here allows to have different default values depending on the backend we use. For instance, the port of MySQL can differ from the port of Redis, so it's rather hard to put the defaults in the `.get()` call.

cliquet/storage/postgresql/__init__.py
```diff
@@ -0,0 +1,458 @@
...
+            if conn and not conn.closed:
+                conn.close()
+
+    def _init_schema(self):
+        """Create PostgreSQL tables, only if not exists.
+
+        :note:
+            Relies on JSON fields, available in recent versions of PostgreSQL.
+        """
+        # Since indices cannot be created with IF NOT EXISTS, inspect:
```
ametaireau: nit: The `:` at the end of the sentence doesn't bring anything, replace with a ".";

cliquet/storage/postgresql/__init__.py
```diff
@@ -0,0 +1,458 @@
...
+        try:
+            inspect_tables = "SELECT * FROM records LIMIT 0;"
```
ametaireau: I expected postgresql to come with a way to know if a table existed, by looking this up in a meta table somewhere, isn't it possible?

ametaireau: If not, we should probably read the "table name" from the configuration rather than having in in clear here.

cliquet/storage/postgresql/__init__.py
```diff
@@ -0,0 +1,458 @@
...
+        with self.connect() as cursor:
+            cursor.execute(schema)
+        logger.info('Created PostgreSQL storage tables')
+
+    def flush(self):
+        """Delete records from tables without destroying schema. Mainly used
+        in tests suites.
+        """
+        query = """
+        DELETE FROM deleted;
```
ametaireau: Same here: we should read the name of the table from the configuration.

cliquet/storage/postgresql/__init__.py
```diff
@@ -0,0 +1,458 @@
...
+        DELETE FROM records;
+        """
+        with self.connect() as cursor:
+            cursor.execute(query)
+        logger.debug('Flushed PostgreSQL storage tables')
+
+    def ping(self):
+        try:
+            with self.connect() as cursor:
+                cursor.execute("SELECT now();")
```
ametaireau: Does this check we can write on the database or just tests the connectivity?

cliquet/storage/postgresql/__init__.py
```diff
@@ -0,0 +1,458 @@
...
+        with self.connect() as cursor:
+            cursor.execute(query, placeholders)
+            if cursor.rowcount == 0:
+                raise exceptions.RecordNotFoundError(record_id)
+            else:
+                result = cursor.fetchone()
+
+        record = result['data']
+        record[resource.id_field] = record_id
+        record[resource.modified_field] = result['last_modified']
```
ametaireau: I wonder if we should factorize this out of this database, since we're duplicating (I think) these assignments across our storage implementations

cliquet/storage/postgresql/__init__.py
```diff
@@ -0,0 +1,458 @@
...
+        RETURNING last_modified::BIGINT
+        """
+        resource_name = classname(resource)
+        placeholders = dict(record_id=record_id,
+                            user_id=user_id,
+                            resource_name=resource_name,
+                            data=json.dumps(record))
+
+        with self.connect() as cursor:
+            self.check_unicity(resource, user_id, record)
```
ametaireau: Oh, so we need to actually do a GET before updating (or creating) the value.

I suppose this means it's not possible to ad constraints on JSON fields.

cliquet/storage/postgresql/__init__.py
```diff
@@ -0,0 +1,458 @@
...
+
+        :note:
+            All rules are combined using OR.
+
+        :note:
+            Field names are escaped as they come from HTTP API.
+
+        :returns: A SQL string with placeholders, and a dict mapping
+            placeholders to actual values.
+        :rtype: tuple
```
ametaireau: Interesting, it's the first time I see this, we should be consistent and put if everywhere if we start to.

ametaireau:
LGTM

cliquet/storage/postgresql/__init__.py
leplatrem: +1 to look up pg catalog (if no specific permission required)

cliquet/storage/postgresql/__init__.py
leplatrem: -1 for putting table names in configuration

cliquet/storage/postgresql/__init__.py
leplatrem: +1 to update the metadata table for example

cliquet/storage/postgresql/__init__.py
leplatrem: please open an issue to address this specifically :)

cliquet/storage/postgresql/__init__.py
leplatrem: Well this implementation is not ideal, but it allowed me to reuse a piece of code.
We could open an issue to substitute this with table check constraints (JSON level)

leplatrem:
Rebased.

PR #10, opened 2015-02-20
Use the RedisSession as a cache backend for PyFxA.

Natim:


cliquet/authentication.py
```diff
@@ -57,6 +59,7 @@ def forget(self, request):
     def _get_credentials(self, request):
         authorization = request.headers.get('Authorization', '')
         settings = request.registry.settings
+        ttl = int(settings.get('fxa-oauth.cache_ttl_seconds', 5 * 60))
```
leplatrem: Move 5 * 60 to global DEFAULT_OAUTH_CACHE_SECONDS ?

cliquet/__init__.py
```diff
@@ -54,7 +54,8 @@ def set_auth(config):
     """Define the authentication and authorization policies.
     """
     policies = [
-        authentication.Oauth2AuthenticationPolicy(),
+        authentication.Oauth2AuthenticationPolicy(
+            cache=config.registry.session),
```
leplatrem: you provide cache here, and never use it

cliquet/authentication.py
```diff
@@ -67,7 +70,9 @@ def _get_credentials(self, request):
         server_url = settings['fxa-oauth.oauth_uri']
         scope = settings['fxa-oauth.scope']
 
-        auth_client = OAuthClient(server_url=server_url)
+        auth_client = OAuthClient(
+            server_url=server_url,
+            cache=SessionCache(request.registry.session, ttl=ttl))
```
leplatrem: should it be self.cache ?

cliquet/__init__.py
```diff
@@ -53,8 +56,12 @@ def _redirect_to_version_view(request):
 def set_auth(config):
     """Define the authentication and authorization policies.
     """
+    ttl = int(config.registry.settings.get('fxa-oauth.cache_ttl_seconds',
```
ametaireau: Having this named "ttl" can cause confusion in case we have other ttls in there. I would name it with the name of the config, or `oauth_cache_ttl`.

cliquet/session/__init__.py
```diff
@@ -22,3 +22,19 @@ def get(self, key):
...
+    def __init__(self, session, ttl):
+        self.ttl = ttl
+        self.cache = session
+
+    def get(self, key):
+        return self.cache.get(key)
+
+    def set(self, key, value):
+        self.cache.set(key, value)
+        self.cache.expire(key, self.ttl)
```
ametaireau: Can we have the expire passed to the set directly?

ametaireau:
This looks good to me but needs some minor tweaking. We also lack tests ;)

PR #11, opened 2015-02-23
Refactor identification of resources (fixes #3)

### LINKED ISSUE #3: Resource id should be a classproperty
Instead of using classname() in backend code

```python
class classproperty(object):
    def __init__(self, getter):
        self.getter = getter
 
    def __get__(self, instance, owner):
        return self.getter(owner)

class Resource(object):
    @classproperty
    def resource_id(cls):
        return cls.__name__
```

leplatrem:


ametaireau:
I love that.

PR #14, opened 2015-02-23
Automatically add the API_PREFIX if missing — Fixes #113

Natim:


cliquet/tests/test_views_batch.py
```diff
@@ -81,6 +71,16 @@ def test_batch_cannot_be_recursive(self):
         resp = self.app.post_json('/batch', body, status=400)
         self.assertIn('Recursive', resp.json['message'])
 
+    def test_responses_are_resolved_with_api_with_prefix(self):
```
ametaireau: `are_resolved_with_api_prefix` ?

ametaireau: or even, `test_batch_adds_api_prefix_if_not_present`

cliquet/views/batch.py
```diff
@@ -116,7 +116,10 @@ def build_request(original, dict_obj):
     :param original: the original batch request.
     :param dict_obj: a dict object with the sub-request specifications.
     """
+    api_prefix = '/'.join(original.upath_info.split('/')[:2])
```
ametaireau: If we have path_info = "/v1/toto/tata", then the split returns `['', 'v1', 'toto', 'tata']`

and then we join it again . We should do a format with a `"/{prefix}".format(prefix=original.upath_info.split('/')[1])`

cliquet/views/batch.py
```diff
@@ -116,7 +116,10 @@ def build_request(original, dict_obj):
...
     path = dict_obj['path']
+    if not path.startswith(api_prefix):
+        path = '%s%s' % (api_prefix, path)
```
leplatrem: path = api_prefix + path  ?

cliquet/views/batch.py
leplatrem: Can we put this comment in the code ?

cliquet/views/batch.py
ametaireau: +1

PR #15, opened 2015-02-24
Docstrings for storage backend

leplatrem:


Natim:
:)

cliquet/storage/__init__.py
```diff
@@ -11,28 +11,167 @@ def __init__(self, id_generator=None, *args, **kwargs):
...
+
+        :returns: the record object.
+        :rtype: dict
+        """
         raise NotImplementedError
 
     def update(self, resource, user_id, record_id, record):
+        """Overwrite the record with the specified id.
+
+        If the specified id is not found, the record is created with the
```
Natim: It should raise a RecordNotFound IMHO

cliquet/storage/__init__.py
```diff
@@ -11,28 +11,167 @@ def __init__(self, id_generator=None, *args, **kwargs):
...
+        """
+
+        :param resource: the record associated resource
+        :type resource: cliquer.resource.BaseResource
+
+        :param user_id: the owner of the record
+        :type user_id: unicode
+
+        :param filters: Optionally filter the records by their attribute.
+            Each filter in this list refers to a field, a value and an
```
Natim: a tuple of a field, value and comparison.

cliquet/storage/__init__.py
```diff
@@ -11,28 +11,167 @@ def __init__(self, id_generator=None, *args, **kwargs):
...
+
+        :param limit: Optionnally limit the number of records to be retrieved.
+        :type limit: integer
+
+        :param include_deleted: Optionnally include the deleted records that
+            match the filters.
+        :type include_deleted: boolean
+
+        :returns: the limited list of records, and the total number of
+            matching records.
```
Natim: in the collection (without the deleted ones.)

cliquet/storage/__init__.py
leplatrem: Please open an issue

Natim: Actually maybe do we need this to work like that to handle the PUT endpoint.

PR #20, opened 2015-02-24
Refactor memory and SQL backends

leplatrem:
* [x] Move every piece of code related to memory records to MemoryBasedStorage
* [x] Move MemoryBasedStorage to memory module
* [x] Implement unicity check in pure SQL (and within the same transaction snapshot)
* [x] Move record filtering and sorting as MemoryBasedStorage methods


Natim:
:+1:

Natim:
Rebase needed apparently.

PR #24, opened 2015-02-25
Add Apache v2 license

leplatrem:


setup.py
```diff
@@ -25,6 +25,7 @@
       version='0.3.dev0',
       description='cliquet',
       long_description=README,
+      license='APLv2.0',
       classifiers=[
           "Programming Language :: Python",
           "Topic :: Internet :: WWW/HTTP",
```
Natim: ``License :: OSI Approved :: Apache Software License``

CONTRIBUTORS.rst
```diff
@@ -0,0 +1,8 @@
+Contributors
+############
+
+* Alexis Metaireau <alexis@mozilla.com>
+* Mathieu Leplatre <mathieu@mozilla.com>
+* Nicolas Perriault <nperriault@mozilla.com>
+* Rémy Hubscher <natim@mozilla.com>
```
Natim: ``rhubscher@mozilla.com``

setup.py
ametaireau: This is version 1 I believe, and I don't see the version 2 [in the list of trove classifiers](https://pypi.python.org/pypi?%3Aaction=list_classifiers). Maybe we should ask them to add one and in the meantime just add a "v2" at the end.

ametaireau:
LGTM.

setup.py
leplatrem: Just looked at https://pypi.python.org/pypi/apache-libcloud :) simply mentions apache license (2.0)

PR #27, opened 2015-02-25
Make every configuration optional (fixes #42)

### LINKED ISSUE #42: Refactor settings with defaults
* [x] Centralize default values for all settings
* [x] Provide way to easy add custom default values
* [x] Override defaults with environment variables ``CLIQUET_STORAGE_URL``

For example : 

```python
from cliquet import initialize_cliquet, DEFAULT_CONFIG

app_defaults = DEFAULT_CONFIG.copy()
app_defaults['captain.age']  = 42

initialize_cliquet(config, __version__, defaults=app_defaults)
```

Use ``registry.setting('setting_name')``

```python
def read_setting(settings, defaults):
    def with_default(name):
         env_name = name.replace('.', '_').upper()
         setted = settings.get(name, defaults.get(name, ''))
         value = os.getenv(env_name, setted)
         if not value:
                  warnings.warn('Value not set for %s' % name)
    return lamdba name: with_default(name)

registry.setting = read_setting(settings)
```

leplatrem:
fixes #42 

While wrinting the documentation, I realize that configuration could have default everywhere


cliquet/__init__.py
```diff
@@ -154,14 +156,16 @@ def includeme(config):
...
     config.registry.storage = storage.load_from_config(config)
 
-    session = config.maybe_dotted(settings['cliquet.session_backend'])
+    session = config.maybe_dotted(settings.get('cliquet.session_backend',
+                                               DEFAULT_SESSION_BACKEND))
     config.registry.session = session.load_from_config(config)
 
-    config.registry.project_name = settings['cliquet.project_name']
-    config.registry.project_docs = settings['cliquet.project_docs']
+    config.registry.project_name = settings.get('cliquet.project_name', '')
```
Natim: Not sure for this one.

cliquet/__init__.py
```diff
@@ -19,6 +19,8 @@
 from cliquet.utils import msec_time
 
 DEFAULT_OAUTH_CACHE_SECONDS = 5 * 60
+DEFAULT_STORAGE_BACKEND = 'cliquet.storage.postgresql'
```
Natim: Not sure if the default should be postgresql

Natim:
Refs https://github.com/mozilla-services/readinglist/issues/81

cliquet/__init__.py
Natim: I mean the project_name configuration should be mandatory and not default to empty string.

cliquet/__init__.py
leplatrem: ah yes, and that would mean it should  not be in extra_require then !

cliquet/__init__.py
leplatrem: Why ? it is used in hello view only !

Default to cliquet ?

ametaireau: Aim is to fail if it's not defined so that cliquet users define it. We can also fallback to a default value but in this case issuing a warning would help.

Natim: Maybe we should do the same thing as we do with version for the default value, use the default package name.

ametaireau: We don't do that anymore in cliquet: cliquet doesn't make any assumptions about the version, and the client must provide it explicitely.

Natim: I mean passing it as we do for version rather than reading it from the conf

ametaireau: Ah okay, I understand your proposal I think. Let me rephrase so I'm sure we're on the same page here.

Since we're already passing the version information to cliquet in `initialize_cliquet`, we should reuse this version to display it on the home page.

If that's it, I concur :)

Natim: > we should reuse this version to display it on the home page.
Isn't it already the case?

ametaireau: Hmm, I'm almost all wrong. We're already doing that (I wrote the patch for it, meh).

Your proposal is to do the same thing for the name. If yes, then I believe we should use that for default, but we should let the possibility to the user to configure another name (for instance,  "cloud storage" is a project name and "kinto" is the codename. They aren't the same).

Natim: > Your proposal is to do the same thing for the name. If yes, then I believe we should use that for default

Totally agree :+1:

leplatrem:
r?

leplatrem:
If you agree, I update this PR to include #42 and implement what you suggest about project name

Natim:
Yes ok.

Natim:
A warning is not exactly what I suggested about project name :)

leplatrem:
Yes I know

leplatrem:
@ametaireau @Natim r?

Here is what I did: 
* Read settings from environment
* Centralize defaults
* Cleaned tests settings to use defaults
* Got rid of ``settings.get`` since I assume we have defaults for everything
* Got rid of registry usage for hello view value (*thought it has better consistency*)
* Added project_name to initialize_cliquet (*not a fan of this btw, I would have liked to keep this method very very stupid*)
* Warn if no project name
* Raise error is no project version (can also be specified in settings)

PR #29, opened 2015-02-25
Resource api

leplatrem:
While writing the docs, I realized too much stuff was public in resource class

### DIFF
--- a/cliquet/resource.py
+++ b/cliquet/resource.py
@@ -80,15 +80,6 @@ class BaseResource(object):
 
         return CorniceSchema.from_colander(colander_schema)
 
-    def raise_invalid(self, location='body', **kwargs):
-        """Helper to raise a validation error.
-
-        :raises: :class:`pyramid.httpexceptions.HTTPBadRequest`
-        """
-        self.request.errors.add(location, **kwargs)
-        response = errors.json_error(self.request.errors)
-        raise response
-
     #
     # End-points
     #
@@ -98,6 +89,7 @@ class BaseResource(object):
         cors_headers=('Next-Page', 'Total-Records', 'Last-Modified')
     )
     def collection_get(self):
+        """Collection `GET` endpoint."""
         self._add_timestamp_header(self.request.response)
         self._raise_304_if_not_modified()
         self._raise_412_if_modified()
@@ -118,6 +110,7 @@ class BaseResource(object):
 
     @resource.view(permission='readwrite')
     def collection_post(self):
+        """Collection `POST` endpoint."""
         self._raise_412_if_modified()
 
         new_record = self.process_record(self.request.validated)
@@ -127,6 +120,7 @@ class BaseResource(object):
 
     @resource.view(permission='readwrite')
     def collection_delete(self):
+        """Collection `DELETE` endpoint."""
         settings = self.request.registry.settings
         enabled = settings.get('cliquet.delete_collection_enabled', 'true')
         if not native_value(enabled):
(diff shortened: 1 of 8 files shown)


leplatrem:
@ametaireau @Natim r?

Natim:
Le retour des underscores :)

cliquet/resource.py
```diff
@@ -120,30 +100,36 @@ def fetch_record(self):
...
-                self.raise_invalid(name=field, description=error)
+                self._raise_invalid(name=field, description=error)
 
         updated = record.copy()
         updated.update(**changes)
-        return self.validate(updated)
+        return self._validate(updated)
 
-    def validate(self, record):
+    def _validate(self, record):
```
ametaireau: Should this stay public, in case a library user would like to define its own validation logic?

ametaireau:
I'm a little fuzzy about what we're using private / public for here.

Since this is a base class, I read it like that: from inside the class, everything prefixed with an underscore is to be accessed by the base class, and not by the subclasses.

If this is the case, I believe the helpers to send errors should be public, since someone defining a subclass could make good use of such helpers (and I believe if they see them they will start to use them anyway, even if they start with an underscore, so we better put them public from the start).

Natim:
> they will start to use them anyway, even if they start with an underscore, so we better put them public from the start

Exactly my point.

cliquet/resource.py
```diff
@@ -154,13 +140,13 @@ def validate(self, record):
                 self.request.errors.add('body', name=field, description=error)
             raise errors.json_error(self.request.errors)
 
-    def add_timestamp_header(self, response):
+    def _add_timestamp_header(self, response):
```
Natim: Should this stay public, in case a library user would like to add it to other endpoints?

cliquet/resource.py
```diff
@@ -154,13 +140,13 @@ def validate(self, record):
...
         """Add current timestamp in response headers, when request comes in.
         """
         timestamp = six.text_type(self.timestamp).encode('utf-8')
         response.headers['Last-Modified'] = timestamp
 
-    def raise_304_if_not_modified(self, record=None):
+    def _raise_304_if_not_modified(self, record=None):
```
Natim: Should this stay public, in case a library user would like to add it to other endpoints?

cliquet/resource.py
```diff
@@ -176,10 +162,10 @@ def raise_304_if_not_modified(self, record=None):
 
             if current_timestamp <= modified_since:
                 response = HTTPNotModified()
-                self.add_timestamp_header(response)
+                self._add_timestamp_header(response)
                 raise response
 
-    def raise_412_if_modified(self, record=None):
+    def _raise_412_if_modified(self, record=None):
```
Natim: Should this stay public, in case a library user would like to add it to other endpoints?

cliquet/resource.py
```diff
@@ -202,10 +188,16 @@ def raise_412_if_modified(self, record=None):
                         error=HTTPPreconditionFailed.title,
                         message=error_msg),
                     content_type='application/json')
-                self.add_timestamp_header(response)
+                self._add_timestamp_header(response)
                 raise response
 
-    def raise_conflict(self, exception):
+    def _raise_invalid(self, location='body', **kwargs):
```
Natim: Should this stay public, in case a library user would like to add it to other endpoints?

cliquet/resource.py
```diff
@@ -202,10 +188,16 @@ def raise_412_if_modified(self, record=None):
...
+        """Helper to raise a validation error."""
+        self.request.errors.add(location, **kwargs)
+        response = errors.json_error(self.request.errors)
+        raise response
+
+    def _raise_conflict(self, exception):
```
Natim: Should this stay public, in case a library user would like to add it to other endpoints?

Natim:
I am ok with the other changes.

ametaireau:
Thanks for updating the pull request, @Natim. I'm okay to sign-off this merge, but since @leplatrem started it I'll let him review and merge.

Natim:
This apparently needs a rebase.

PR #28, opened 2015-02-25
Cliquet docs

leplatrem:

PR #33, opened 2015-02-26
Add CHANGELOG in documentation.

Natim:
Refs mozilla-services/readinglist#114

r? @leplatrem


leplatrem:
It is done in #28 already

PR #36, opened 2015-02-26
Add public methods and hide protocol

leplatrem:
Just a try: what do you think ?

### DIFF
--- a/cliquet/resource.py
+++ b/cliquet/resource.py
@@ -262,7 +262,8 @@ class BaseResource(object):
             sorting=sorting,
             pagination_rules=pagination_rules,
             limit=limit,
-            include_deleted=include_deleted)
+            include_deleted=include_deleted,
+            **self.db_kwargs)
 
         next_page = None
         if limit and len(records) == limit and total_records > limit:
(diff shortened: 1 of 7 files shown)


ametaireau:
+1 for me.

leplatrem:
We should problably reorder the methods to put endpoints on top

cliquet/resource.py
```diff
@@ -123,7 +124,167 @@ def raise_invalid(self, location='body', **kwargs):
...
+            the total number of records in the result set, and the next page
+            url.
+        :rtype: tuple
+        """
+        filters = self._extract_filters()
+        sorting = self._extract_sorting()
+        pagination_rules, limit = self._extract_pagination_rules_from_token(
+            sorting)
+
+        include_deleted = self.modified_field in [f[0] for f in filters]
```
leplatrem: I'm not totally convinced of having filters extraction here. If we mean to keep protocol stuff out of customization, this should probably go back to the endpoint code.

```python
def get_records(self, filters, sorting, pagination, limit, include_deleted):
       # Customize : disable sorting on certain fields
       sorting.remove(('age', 1))
       return super(..., ...).get_records(...)
```

leplatrem:
Well actually @Natim you can merge this, since it doesn't target master. We'll continue the discussion in #29

PR #43, opened 2015-02-26
Improve docs about authentication

leplatrem:


docs/api/authentication.rst
```diff
@@ -4,15 +4,40 @@ Authentication
...
+the HTTP method to authenticate requests may differ.
 
-Firefox Account OAuth Bearer token
-==================================
+A policy based on *OAuth2 bearer tokens* is recommended, but not mandatory.
+
+A *Basic Auth* can also be enabled in :ref:`configuration` for the convenience
+of clients or testing.
+
+By default, we profite a setup using :term:`Firefox Accounts`, that verifies
```
Natim: ``profite`` is not an english word, do you mean **propose** here?

docs/api/authentication.rst
```diff
@@ -4,15 +4,40 @@ Authentication
...
+::
+
+    Authorization: Basic <basic_token>
+
+
+The token is built using this formula ``base64("username:password")``.
+
+:notes:
+
+    If not enabled (**default**) this will result in a ``401`` error response.
```
Natim: We should add a link to the configuration documentation here.

docs/configuration.rst
```diff
@@ -49,11 +49,65 @@ See :ref:`storage backend documentation <storage>` for more details.
...
+Internally, Cliquet relies on Pyramid ``authenticated_userid`` request
+attribute to associate users to records.
+
+
+.. code-block :: python
+
+    def main(global_config, **settings):
+        config = Configurator(settings=settings)
+
+        cliquet.initialize_cliquet(config, __version__)
```
Natim: We should add the definition of __version__ here.

docs/api/authentication.rst
leplatrem: hehe :) provide :)

docs/configuration.rst
leplatrem: This was supposed to be done in get started section, but I call recall it

Natim: Yep ok

PR #44, opened 2015-02-27
Add PostgreSQL session backend

leplatrem:


cliquet/session/postgresql.py
```diff
@@ -0,0 +1,104 @@
...
+        query = """
+        DELETE FROM session;
+        """
+        with self.connect() as cursor:
+            cursor.execute(query)
+        logger.debug('Flushed PostgreSQL session tables')
+
+    def ping(self):
+        try:
+            self.set('heartbeat', True)
```
Natim: ``return True`` missing.

cliquet/session/postgresql.py
```diff
@@ -0,0 +1,104 @@
...
+    def __init__(self, **kwargs):
+        self._conn_kwargs = kwargs
+        self._init_schema()
+
+    def _init_schema(self):
+        schema = """
+        CREATE TABLE IF NOT EXISTS session(
+            key VARCHAR(256) PRIMARY KEY,
+            value TEXT NOT NULL,
+            ttl INT4 DEFAULT NULL
```
Natim: We should put an index on ttl since we are using it for filtering.

cliquet/session/postgresql.py
leplatrem: :+1:

cliquet/session/postgresql.py
leplatrem: true!

PR #45, opened 2015-02-27
[WIP] Find a fix for PostgreSQL sequences after PUT forcing record id

leplatrem:
This would probably imply to use UUID instead of SERIAL for records ids.

### DIFF
--- a/cliquet/storage/postgresql/__init__.py
+++ b/cliquet/storage/postgresql/__init__.py
@@ -218,10 +218,19 @@ class PostgreSQL(StorageBase):
             # Create or update ?
             query = "SELECT id FROM records WHERE id = %s;"
             cursor.execute(query, (record_id,))
-            query = query_update if cursor.rowcount > 0 else query_create
+            if cursor.rowcount > 0:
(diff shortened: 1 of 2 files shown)


cliquet/storage/postgresql/__init__.py
```diff
@@ -218,10 +218,19 @@ def update(self, resource, user_id, record_id, record):
...
+            else:
+                cursor.execute(query_create, placeholders)
+                result = cursor.fetchone()
 
-            cursor.execute(query, placeholders)
-            result = cursor.fetchone()
+                cursor.execute('SELECT MAX(id) AS max FROM records;')
+                max_id = cursor.fetchone()['max']
+                bump_serial = """
+                ALTER SEQUENCE records_id_seq RESTART WITH %s;"
```
Natim: I would prefer to use UUID here.

leplatrem: yes this was late yesterday night, I hate this and this does not work at all :)

leplatrem:
Replaced by @ametaireau in #70

PR #50, opened 2015-03-02
Remove py2.6 support

ametaireau:
r? @Natim

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,6 @@ services: redis-server
 addons:
   postgresql: "9.3"
 env:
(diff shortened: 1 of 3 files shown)


Natim:
Then we should change ``set([])`` with ``{}`` in 

* cliquet/tests/resource/test_pagination.py: 61
* cliquet/tests/resource/test_pagination.py: 62

Natim:
./cliquet/utils.py:
    2 :     import simplejson as json
./tox.ini:
   13 :     simplejson

ametaireau:
(updated)

PR #51, opened 2015-03-02
Filter on deleted with since

leplatrem:
r? @ametaireau

### DIFF
--- a/cliquet/resource.py
+++ b/cliquet/resource.py
@@ -52,7 +52,7 @@ class BaseResource(object):
     modified_field = 'last_modified'
     """Name of *last modified* field in resource schema"""
 
-    deleted_mark = ('deleted', True)
+    deleted_mark = 'deleted'
     """Field and value of deleted status in records"""
 
     def __init__(self, request):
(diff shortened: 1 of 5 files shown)


cliquet/resource.py
```diff
@@ -52,7 +52,7 @@ class BaseResource(object):
     modified_field = 'last_modified'
     """Name of *last modified* field in resource schema"""
 
-    deleted_mark = ('deleted', True)
+    deleted_mark = 'deleted'
```
ametaireau: so, it's not a "mark" anymore but a "field", like the other ones.

cliquet/resource.py
```diff
@@ -60,7 +60,8 @@ def __init__(self, request):
         self.db = request.db
         self.db_kwargs = dict(resource=self,
                               user_id=request.authenticated_userid)
-        self.known_fields = [c.name for c in self.mapping.children]
+        self.known_fields = [c.name for c in self.mapping.children] + \
+                            [self.deleted_mark]
```
ametaireau: deleted_field.

PR #62, opened 2015-03-03
Update deleted_field support for postgres.

ametaireau:
Without this, cliquet doesn't actually delete anything from the database.

r? @leplatrem @Natim ?


Natim:
:+1:

PR #70, opened 2015-03-04
Postgres userid fix

ametaireau:
r? @leplatrem

### DIFF
--- a/cliquet/tests/test_storage.py
+++ b/cliquet/tests/test_storage.py
@@ -60,6 +60,7 @@ class BaseTestStorage(object):
         self.storage = self.backend.load_from_config(self._get_config())
         self.resource = TestResource()
         self.user_id = '1234'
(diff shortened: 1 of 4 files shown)


cliquet/tests/test_storage.py
```diff
@@ -571,10 +577,41 @@ def test_pagination_rules_on_last_modified_apply_to_deleted_records(self):
...
+
+    def test_users_cannot_update_other_users_record(self):
+        record = self.create_record()
+        new_record = {"another": "record"}
+        self.storage.update(self.resource, self.other_user_id, record['id'],
+                            new_record)
+        not_updated = self.storage.get(self.resource, self.user_id,
+                                       record['id'])
+
+        self.assertNotIn("another", not_updated)
```
Natim: We don't need to add a similar test in views ?

Natim:
r+

PR #77, opened 2015-03-10
Handle backend errors for 503 (fixes #21)

### LINKED ISSUE #21: Handle 503 — when the database fail, we should log the error and return a 503
Today it is a 500

leplatrem:
fixes #21 

* [x] Define BackendError exception
* [x] Make sure all storage client code is wrapped to catch errors
* [x] Make sure all session client code is wrapped to catch errors
* [x] Bind BackendError to 503 response
* [x] Test that backend errors end up as 503 in views


Natim:
Great :)

PR #79, opened 2015-03-10
Make sure pagination token is a dict (fixes #72)

### LINKED ISSUE #72: When a token is json valid but doesn't match the protocol it should not 500
``[[('last_modified', 0, '>')]]``

leplatrem:


Natim:
r+

PR #83, opened 2015-03-12
Bug 1142401 — Remove the redirect querystring parameter to the /fxa-oauth/login endpoint.

Natim:
https://bugzilla.mozilla.org/show_bug.cgi?id=1142401

### DIFF
--- a/cliquet/views/oauth.py
+++ b/cliquet/views/oauth.py
@@ -1,5 +1,5 @@
 import uuid
-import urlparse
+from six.moves.urllib.parse import urlparse
 from fnmatch import fnmatch
 
 from cornice import Service
@@ -55,7 +55,7 @@ def authorized_redirect(req):
     if 'redirect' not in req.validated:
         return True
 
-    domain = urlparse.urlparse(req.validated['redirect']).netloc
+    domain = urlparse(req.validated['redirect']).netloc
 
     if not any((fnmatch(domain, auth) for auth in authorized)):
         req.errors.add('querystring', 'redirect',
--- a/cliquet/views/oauth.py
+++ b/cliquet/views/oauth.py
@@ -1,4 +1,6 @@
 import uuid
+import urlparse
+from fnmatch import fnmatch
 
 from cornice import Service
 import colander
@@ -7,6 +9,7 @@ from fxa import errors as fxa_errors
 
 from pyramid import httpexceptions
 from pyramid.security import NO_PERMISSION_REQUIRED
+from pyramid.settings import aslist
 
 from cliquet import errors
 from cliquet.schema import URL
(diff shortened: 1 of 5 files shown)


cliquet/views/oauth.py
```diff
@@ -33,15 +37,15 @@ def persist_state(request):
     It will be compared when return from login page on OAuth server.
     """
     state = uuid.uuid4().hex
-    request.registry.session.set(state, request.validated['redirect'])
+    request.registry.session.set(state,
+                                 fxa_conf(request, 'webapp.redirect_url'))
+    request.registry.session.expire(
```
Natim: Need test.

cliquet/views/oauth.py
```diff
@@ -82,6 +86,9 @@ def fxa_oauth_token(request):
     # Require on-going session
     stored_redirect = request.registry.session.get(state)
 
+    # Make sure we cannot try twice with the same code
+    request.registry.session.delete(state)
```
Natim: Need a test.

cliquet/views/oauth.py
rfk: Since this is now a fixed config value, do you need to persist it in the state?

Natim: Well it is to stay future proof since we probably will need to go for a url whitelist at some point. Having the server linked to only one webclient is a temporary feature. Also we still need to store the state.

cliquet/views/oauth.py
```diff
@@ -23,9 +22,14 @@
                 path='/fxa-oauth/token',
                 error_handler=errors.json_error_handler)
 
+DEFAULT_STATE_EXPIRY_SECONDS = 3600  # 1 hour
```
leplatrem: 5 minutes would have been enough :)

rfk: I've been amazed how long the oauth dance can take if the user e.g. forgets their password or has to sign up for a new account, and then wait for the email to come through, and then fish it out of their spam folder, and then click the link but it opens in a different browser, etc etc....

leplatrem: Fascinating :)

cliquet/views/oauth.py
```diff
@@ -33,15 +37,15 @@ def persist_state(request):
     It will be compared when return from login page on OAuth server.
     """
     state = uuid.uuid4().hex
-    request.registry.session.set(state, request.validated['redirect'])
+    redirect_url = fxa_conf(request, 'webapp.redirect_url')
+    request.registry.session.set(state, redirect_url)
```
ametaireau: We need to make extra clear this is *not* an user session. E.g. anyone with a state valid (for all our users) would pass the check.

ametaireau: Otherwise a potential vector attack is to generate a lot of states and then try brute forcing.

ametaireau: Hmm, actually not after this PR since we're deleting them.

ametaireau:
We should add a pattern whitelist here rather than defaulting.

ametaireau:
Ready for review. r? @leplatrem

leplatrem:
I would suggest we had it to DEFAULT_SETTINGS too :)

ametaireau:
It's in the DEFAULT_SETTINGS already.

Natim:
Yes see https://github.com/mozilla-services/cliquet/blob/1142401-remove-redirect-login-parameter/cliquet/__init__.py#L41

PR #82, opened 2015-03-13
Consider storing the raw FxA uid, rather than hmacing it

Natim:
While poking around in service of Bug 1141166 [1] I noticed that you're hmacing the FxA uid rather than storing it directly:

  https://github.com/mozilla-services/cliquet/blob/master/cliquet/authentication.py#L75

I assume this is intended to obfuscate it for privay purposes?

We generally haven't been treating the account uid as private, and are encouraging reliers/providers to just use it directly as the primary identifier for the account.  You might consider saving a bit of code and ops complexity by just storing it directly.

Having the raw uid could also be useful if you e.g. you want to publish notification events for other reliers via the upcoming https://github.com/mozilla/fxa-notification-server.

[1] https://bugzilla.mozilla.org/show_bug.cgi?id=1141166

### DIFF
--- a/cliquet/authentication.py
+++ b/cliquet/authentication.py
@@ -71,10 +71,7 @@ class Oauth2AuthenticationPolicy(base_auth.CallbackAuthenticationPolicy):
         auth_client = OAuthClient(server_url=server_url, cache=self.cache)
         try:
             profile = auth_client.verify_token(token=auth, scope=scope)
-            hmac_secret = settings.get('cliquet.userid_hmac_secret')
-            user_id = hmac.new(hmac_secret.encode('utf-8'),
-                               profile['user'].encode('utf-8'),
-                               hashlib.sha256).hexdigest()
+            user_id = profile['user'].encode('utf-8')
         except fxa_errors.OutOfProtocolError:
             raise httpexceptions.HTTPServiceUnavailable()
         except (fxa_errors.InProtocolError, fxa_errors.TrustError):


Natim:
Here is a commit that remove the hmac of the FxA UID, also I am keeping the hmac_secret for the basic_auth Authentication backend.

ckarlof:
> I really don't understand the benefit of storing the UID directly

One intention of the FxA uid is to let us link user data across our services. If the privacy goal of hmacing the FxA uid is to avoid linking "user data in-between application that uses the FxA account", then our two goals are potentially working against each other.

However, suppose we decided that preventing linkage of data across services was really a goal. This mechanism on its own isn't effective in preventing it. FxA has a complete list of FxA uids, and given the HMAC secret (which we also control), Mozilla can link uids with hmac-ed uids at will. In other words, we can still do the linkage whenever we want to, and this mechanism just makes it more cumbersome and complex to do so.

ckarlof:
> IMHO, rollbacking an already implemented feature is a risk as well.

> We should probably have this broader discussion before rollbacking this feature because it will not be trivial to revert it later.

> Having a consistent ID size accross Authentication scheme available for the readinglist is something that reduce the risk at a database level.

I recognize that the proposed patch adds risk this late in the process. I'll leave it to you and @tarekziade to decide on the appropriate technical approach, but the high level requirement I'm making is that we have the capability to directly associate user data with the FxA uid on the RL server from day 1.

ametaireau:
Since we don't have anything released in prod right now, I believe it's the right timing to do the change.

We will issue a new tag and ask @Micheletto to use it rather than the already tagged version.

PR #87, opened 2015-03-16
Use batch instead of multiple requests in cloud storage backend

leplatrem:


cliquet/storage/cloud_storage.py
```diff
@@ -183,29 +183,51 @@ def get_all(self, resource, user_id, filters=None, sorting=None,
...
-                                        headers=self._build_headers(resource))
-                resp.raise_for_status()
-                for record in resp.json()['items']:
+                params_ = list(params)
+                params_ += [("%s%s" % (FILTERS[op], k), v)
+                              for k, v, op in filters]
+                querystring = '&'.join(['%s=%s' % (p, v) for p, v in params_])
+                batch_payload['requests'].append({
+                    'path': url + '?%s' % querystring,
+                })
```
ametaireau: Neat! I'm actually thinking we could add a context manager here and when we exit() from it we send the batch request.

leplatrem:
In order to fix the tests on Travis, we should make sure the underlying Kinto runs the cliquet version of the current branch.
This is not the case, it uses the last pypi version :(

leplatrem:
@Natim @ametaireau r?

Natim:
> In order to fix the tests on Travis - I think we should use only one venv for everything and finish the install by ``pip install -e .`` so that we will use the same version of cliquet for everything (kinto and tests).

PR #90, opened 2015-03-18
Set default limit for number of requests in batch

Natim:
Currently if the setting `cliquet.batch_max_requests` is not set, there is no limit.

We could set a default value (e.g. 25 ?). 

I would wait for #27 to be merged before changing this.

### DIFF
--- a/cliquet/tests/test_views_batch.py
+++ b/cliquet/tests/test_views_batch.py
@@ -298,15 +298,14 @@ class BatchServiceTest(unittest.TestCase):
         path = resp['responses'][0]['path'].decode('utf8')
         self.assertEqual(path, u'/v0/ð ®')
 
-    def test_number_of_requests_is_not_limited_by_default(self):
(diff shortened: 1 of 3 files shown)


leplatrem:
I would add a test specifying that number of request is limited by default (e.g. batch with 26 requests returning 400)

cliquet/tests/test_views_batch.py
```diff
@@ -298,15 +298,14 @@ def test_subrequests_responses_paths_are_url_decoded(self):
         path = resp['responses'][0]['path'].decode('utf8')
         self.assertEqual(path, u'/v0/ð ®')
 
-    def test_number_of_requests_is_not_limited_by_default(self):
+    def test_number_of_requests_is_not_limited_when_settings_set_to_none(self):
+        self.request.registry.settings['cliquet.batch_max_requests'] = None
         requests = {}
         for i in range(500):
```
leplatrem: :)

ametaireau:
LGTM.

PR #97, opened 2015-03-18
Add a pref to enable FxA-OAuth views

ametaireau:


cliquet/tests/test_views_oauth.py
```diff
@@ -5,6 +5,22 @@
...
+    def get_app_settings(self):
+        settings = super(DeactivatedViewsTest, self).get_app_settings()
+        settings['fxa-oauth.enabled'] = False
+        return settings
+
+    def test_login_view_is_not_available(self):
+        self.app.get('/fxa-oauth/login', status=404)
+
+    def test_params_view_is_not_available(self):
+        self.app.get('/fxa-oauth/params', status=404)
```
Natim: We want to keep this one, because we have the scope list in here, but we want to strip it to just the scope.

cliquet/__init__.py
```diff
@@ -40,6 +41,7 @@
     'fxa-oauth.scope': 'profile',
     'fxa-oauth.state.ttl_seconds': 3600,  # 1 hour
     'fxa-oauth.webapp.authorized_domains': '',
+    'fxa-oauth.enabled': False,
```
leplatrem: False by default, really ?

leplatrem:
I believe we should inspect the docs, since we mention FxA in quite a few places.

I really would have preferred a clean separate package, that is included as pyramid plugin in projects using cliquet. This way, it would make cliquet really authentication agnostic.

As for readinglist rush, couldn't we redefine the views there and raise a 404 ?

cliquet/__init__.py
Natim: It is the webapp client flow only not the full oauth flow.

It will not disable the fact that you can log using a bearer token.

I think we should work on the wording because here it is too ambiguous.

Natim: ``fxa-oauth.webapp.enabled``?

ametaireau: Yes. `fxa-oauth.relier.enabled` ?

cliquet/tests/test_views_oauth.py
ametaireau: What about providing it somewhere else (e.g. in `/` ?)

cliquet/__init__.py
Natim: Yes I think it works.

cliquet/tests/test_views_oauth.py
leplatrem: Why not just disable the others views except this one ?

Natim: I like the meaning of the URL.

ametaireau: Then I think I need to put it in a different file, otherwise it will be scanned and included.

ametaireau:
Updated.

ametaireau:
This is the preparational work for the split, and is the "fast" solution to have readinglist out of the way. 

However, this code has been thought to be easy to splut and will probably remain useful when it will be in a separate package.

cliquet/views/oauth/params.py
```diff
@@ -0,0 +1,19 @@
...
+params = Service(name='fxa-oauth-params',
+                 path='/fxa-oauth/params',
+                 error_handler=errors.json_error_handler)
+
+
+@params.get(permission=NO_PERMISSION_REQUIRED)
+def fxa_oauth_params(request):
+    """Helper to give Firefox Account configuration information."""
+    return {
+        'client_id': fxa_conf(request, 'client_id'),
```
Natim: We should remove this.

ametaireau: Well, it will return null, no?

Natim: No we should just remove it since no client needs it.

leplatrem:
I like it like that :notes:

Issue #40, opened 2015-02-26
Manage authentication from settings


leplatrem:
Currently, authentication method is setup with code in ``cliquet.initialize()``. [See docs](http://cliquet.readthedocs.org/en/latest/configuration.html#custom-authentication)

They could be specified from configuration files, just like storage backend: 

```
cliquet.authentication_policies  = storage.authentication.basic_auth,
                                   oauth2.providers.google
```

And we could mimic the ``load_from_config()`` to instantiate the class from potential settings

leplatrem:
If I'm not wrong, this could allow @ebrehault to run Kinto with *JSON Web Token* auth, just by changing default configuration.

Related #205.

leplatrem:
pyramid_multiauth can already do that https://github.com/mozilla-services/pyramid_multiauth#deployment-settings

PR #103, opened 2015-03-18
Move auth config out of init

leplatrem:
Make setup cleaner, and would be a first step for #40 

Reopening.

### DIFF
--- a/cliquet/authentication.py
+++ b/cliquet/authentication.py
@@ -9,6 +9,7 @@ from pyramid.interfaces import IAuthenticationPolicy, IAuthorizationPolicy
 from pyramid.security import Authenticated
 from zope.interface import implementer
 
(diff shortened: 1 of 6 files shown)


cliquet/authentication.py
```diff
@@ -44,11 +44,35 @@ def __init__(self, *args, **kwargs):
...
+        self.cache.set(key, value, self.ttl)
+
+    def delete(self, key):
+        self.cache.delete(key)
+
+
 @implementer(IAuthenticationPolicy)
 class Oauth2AuthenticationPolicy(base_auth.CallbackAuthenticationPolicy):
-    def __init__(self, realm='Realm', cache=True):
+    def __init__(self, config, realm='Realm'):
```
Natim: Actually this is an important bugfix.

PR #108, opened 2015-03-19
Use ujson

leplatrem:
* [x] Centralize json usage from utils
* [x] Register ujson for psycopg
* [x] Use ujson in requests https://github.com/kennethreitz/requests/issues/1595#issuecomment-24356026
* [x] Register ujson for JSON renderer in Cornice
* [x] Use ujson for request body deserialization
* [x] Register ujson for structlog

### DIFF
--- a/cliquet/__init__.py
+++ b/cliquet/__init__.py
@@ -11,9 +11,10 @@ import webob
 
 from pyramid.events import NewRequest, NewResponse
 from pyramid.httpexceptions import HTTPTemporaryRedirect, HTTPGone
-from pyramid_multiauth import MultiAuthenticationPolicy
+from pyramid.renderers import JSON as JSONRenderer
 from pyramid.security import NO_PERMISSION_REQUIRED
 from pyramid.settings import asbool
+from pyramid_multiauth import MultiAuthenticationPolicy
 
 # Main Cliquet logger.
 logger = structlog.get_logger()
@@ -183,6 +184,10 @@ def includeme(config):
     # Monkey patch to use ujson
     webob.request.json = utils.json
     requests.models.json = utils.json
+
+    # Override json renderer using ujson
+    renderer = JSONRenderer(serializer=lambda v, **kw: utils.json.dumps(v))
+    config.add_renderer('json', renderer)
 
     load_default_settings(config)
     settings = config.get_settings()
--- a/cliquet/__init__.py
+++ b/cliquet/__init__.py
@@ -5,6 +5,7 @@ import warnings
 
 from dateutil import parser as dateparser
 import pkg_resources
+import requests
 import structlog
 import webob
 
@@ -179,10 +180,9 @@ def includeme(config):
     Service.cors_origins = ('*',)
     Service.default_cors_headers = ('Backoff', 'Retry-After', 'Alert')
 
-    # Monkey patch webob to use ujson
+    # Monkey patch to use ujson
     webob.request.json = utils.json
-
-    # Monkey patch structlog use ujson
+    requests.models.json = utils.json
     structlog.processors.json = utils.json
 
     load_default_settings(config)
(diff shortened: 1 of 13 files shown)


leplatrem:
https://github.com/mozilla-services/cornice/blob/448a9c53e79420fbf092810e168a2ed2fbbde395/cornice/util.py#L65-L66

leplatrem:
@tarekziade Most of json usage is done in Cornice renderer. Shall we register a new JSON renderer then?

Otherwise we rely on Pyramid json for the PATCH request body only (https://github.com/Pylons/pyramid/blob/master/pyramid/request.py#L235-L237)

tarekziade:
@leplatrem yeah ideally any json operation should go to the same lib - if it's simple to plug it everywhere

leplatrem:
In Pyramid json_body overrides WebOb, but strangely in our resourse when we do self.request.json, we rely on ``webob/request.py", line 725, in _json_body__get`` https://github.com/Pylons/webob/blob/master/webob/request.py#L725)

(btw patching a record with empty body fails miserably there ``simplejson/decoder.py", line 400, in raw_decode...JSONDecodeError: Expecting value: line 1 column 1 (char 0);)``, /me will fix)

cliquet/__init__.py
```diff
@@ -179,6 +181,14 @@ def includeme(config):
     Service.cors_origins = ('*',)
     Service.default_cors_headers = ('Backoff', 'Retry-After', 'Alert')
 
+    # Monkey patch to use ujson
+    webob.request.json = utils.json
+    requests.models.json = utils.json
+
+    # Override json renderer using ujson
+    renderer = JSONRenderer(serializer=lambda v, **kw: utils.json.dumps(v))
```
leplatrem: Ignore default kwarg
https://github.com/Pylons/pyramid/blob/master/pyramid/renderers.py#L292

cliquet/logs.py
```diff
@@ -104,7 +104,7 @@ def __call__(self, logger, name, event_dict):
         return log_msg
 
 
-class MozillaHekaRenderer(structlog.processors.JSONRenderer):
+class MozillaHekaRenderer(object):
```
leplatrem: Not inherit from default JSON renderer because ujsons.dumps() does not support cls kwarg
https://github.com/hynek/structlog/blob/master/structlog/processors.py#L146

leplatrem:
@tarek r?

Failing test is instability for cloud storage backend test.

cliquet/storage/postgresql/__init__.py
```diff
@@ -35,6 +34,8 @@ def connect(self):
         cursor = None
         try:
             conn = psycopg2.connect(**self._conn_kwargs)
+            # Will use ujson
+            psycopg2.extras.register_json(conn, loads=json.loads)
```
Natim: Neat :+1:

Natim: Does it json.dumps at some point?

Natim:
r+

cliquet/__init__.py
```diff
@@ -179,6 +181,14 @@ def includeme(config):
     Service.cors_origins = ('*',)
     Service.default_cors_headers = ('Backoff', 'Retry-After', 'Alert')
 
+    # Monkey patch to use ujson
```
ametaireau: We should put this logic in a "patch_json" function.

ametaireau:
`FAIL: test_timestamps_are_unique (cliquet.tests.test_storage.CloudStorageTest)`, so I think we're good to merge.

cliquet/storage/postgresql/__init__.py
leplatrem: We do the dumps when interpolating queries, so this should be enough !

PR #110, opened 2015-03-19
Bringing Sentry configuration.

Natim:
See http://raven.readthedocs.org/en/latest/config/index.html

### DIFF
--- a/cliquet/__init__.py
+++ b/cliquet/__init__.py
@@ -206,11 +206,16 @@ def includeme(config):
     config.add_tween("cliquet.end_of_life_tween_factory")
 
     if settings['cliquet.sentry_dsn']:
-        raven.Client(
+        raven_client = raven.Client(
             settings['cliquet.sentry_dsn'],
             include_paths=['cornice', 'cliquet'] +
             settings['cliquet.sentry_projects'].split(','),
             release=settings['cliquet.project_version'])
+
+        raven_client.captureMessage("%s %s starting." % (
+            settings['cliquet.project_name'],
+            settings['cliquet.project_version']
+        ))
 
     storage = config.maybe_dotted(settings['cliquet.storage_backend'])
     config.registry.storage = storage.load_from_config(config)
--- a/cliquet/__init__.py
+++ b/cliquet/__init__.py
@@ -8,7 +8,7 @@ import pkg_resources
 import requests
 import structlog
 import webob
-from raven import Client
+import raven
 
 from pyramid.events import NewRequest, NewResponse
 from pyramid.httpexceptions import HTTPTemporaryRedirect, HTTPGone
@@ -65,6 +65,7 @@ DEFAULT_SETTINGS = {
     'cliquet.storage_max_fetch_size': 10000,
     'cliquet.userid_hmac_secret': '',
     'cliquet.sentry_dsn': None,
+    'cliquet.sentry_projects': '',
 }
 
 
@@ -204,6 +205,13 @@ def includeme(config):
     handle_api_redirection(config)
     config.add_tween("cliquet.end_of_life_tween_factory")
 
+    if settings.get('cliquet.sentry_dsn') is not None:
+        raven.Client(
+            settings['cliquet.sentry_dsn'],
+            include_paths=['cornice', 'cliquet'] +
+            settings['cliquet.sentry_projects'].split(','),
+            release=settings['cliquet.project_version'])
+
     storage = config.maybe_dotted(settings['cliquet.storage_backend'])
     config.registry.storage = storage.load_from_config(config)
 
@@ -240,9 +248,6 @@ def initialize_cliquet(config, version=None, project_name=None):
     :type project_name: string
     """
     settings = config.registry.settings
-
-    if settings.get('cliquet.sentry_dsn') is not None:
-        client = Client(settings.get('cliquet.sentry_dsn'))
 
     # The API version is derivated from the module version.
     project_version = settings.get('cliquet.project_version') or version
--- a/cliquet/__init__.py
+++ b/cliquet/__init__.py
@@ -64,7 +64,7 @@ DEFAULT_SETTINGS = {
     'cliquet.storage_url': '',
     'cliquet.storage_max_fetch_size': 10000,
     'cliquet.userid_hmac_secret': '',
-    'cliquet.sentry_dsn': None,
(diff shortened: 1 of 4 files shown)


Natim:
Too simple to be true.
@leplatrem @ametaireau

ametaireau:
Neat. The page you linked specifies we should define some more options:

```python
    # inform the client which parts of code are yours
    # include_paths=['my.app']
    include_paths=[__name__.rsplit('.', 1)[0]],

    # pass along the version of your application
    # release='1.0.0'
    # release=raven.fetch_package_version('my-app')
    release=raven.fetch_git_sha(os.path.dirname(__file__)),
```

cliquet/__init__.py
```diff
@@ -7,6 +7,7 @@
 from dateutil import parser as dateparser
 import pkg_resources
 import structlog
+from raven import Client
```
leplatrem: just put import raven, otherwise we don't what Client is :)

cliquet/__init__.py
```diff
@@ -225,6 +227,9 @@ def initialize_cliquet(config, version=None, project_name=None):
     """
     settings = config.registry.settings
 
+    if settings.get('cliquet.sentry_dsn') is not None:
```
leplatrem: if settings['cliquet.sentry_dsn']:   # is recommended to disable via env var

cliquet/__init__.py
```diff
@@ -188,6 +191,13 @@ def includeme(config):
     handle_api_redirection(config)
     config.add_tween("cliquet.end_of_life_tween_factory")
 
+    if settings.get('cliquet.sentry_dsn') is not None:
```
leplatrem: What happens if I have ``cliquet.sentry_dsn = server.com`` in my .ini and if run with ``CLIQUET_SENTRY_DSN= make serve`` ?

leplatrem: I would expect it to temporarily disable sentry :)

Natim: I wouldn't :)

ametaireau:
Great, waiting on travis to merge.

Natim:
Should we add the ping at startup as we did for loop?

ametaireau:
Ah good idea I wasn't thinking about that. That's a good way to check connectivity, so yes!

Natim:
Ok adding http://raven.readthedocs.org/en/latest/usage.html#raven.base.Client.captureMessage

cliquet/__init__.py
```diff
@@ -188,6 +191,18 @@ def includeme(config):
...
     config.add_tween("cliquet.end_of_life_tween_factory")
 
+    if settings['cliquet.sentry_dsn']:
+        raven_client = raven.Client(
+            settings['cliquet.sentry_dsn'],
+            include_paths=['cornice', 'cliquet'] +
+            settings['cliquet.sentry_projects'].split(','),
+            release=settings['cliquet.project_version'])
+
+        raven_client.captureMessage("%s %s starting." % (
```
ametaireau: I think we should ensure we test it's running when the setting is set and not running when it's not. (yeah, me again about tests :p)

cliquet/__init__.py
```diff
@@ -61,6 +62,8 @@
     'cliquet.storage_url': '',
     'cliquet.storage_max_fetch_size': 10000,
     'cliquet.userid_hmac_secret': '',
+    'cliquet.sentry_dsn': None,
+    'cliquet.sentry_projects': '',
```
leplatrem: This deserves a mention in the docs

cliquet/__init__.py
```diff
@@ -188,6 +191,18 @@ def includeme(config):
     handle_api_redirection(config)
     config.add_tween("cliquet.end_of_life_tween_factory")
 
+    if settings['cliquet.sentry_dsn']:
+        raven_client = raven.Client(
+            settings['cliquet.sentry_dsn'],
+            include_paths=['cornice', 'cliquet'] +
+            settings['cliquet.sentry_projects'].split(','),
```
leplatrem: please use intermediary variable

cliquet/__init__.py
```diff
@@ -188,6 +191,18 @@ def includeme(config):
...
+            release=settings['cliquet.project_version'])
+
+        raven_client.captureMessage("%s %s starting." % (
+            settings['cliquet.project_name'],
+            settings['cliquet.project_version']
+        ))
```
leplatrem: this deserves a test in test_initialization :)

docs/configuration.rst
```diff
@@ -104,6 +104,17 @@ redirected to standard output:
...
 
 
+Handling exceptions with Sentry
+:::::::::::::::::::::::::::::::
+
+Sentry logging can be enabled (disabled by default):
+
+.. code-block :: ini
+
+    cliquet.sentry_dsn = your dsn here
```
leplatrem: http://public:secret@example.com/1 ?

docs/configuration.rst
```diff
@@ -104,6 +104,17 @@ redirected to standard output:
...
+    cliquet.sentry_project = sentry project name
```
leplatrem: sentry_projects (comma separated string ?)

leplatrem: cliquet.sentry_projects = readinglist,prod

docs/configuration.rst
```diff
@@ -104,6 +104,17 @@ redirected to standard output:
...
+
+
```
leplatrem: > A basic message will be sent when application starts.

leplatrem:
Appart from docs nits :) r+ !

cliquet/__init__.py
```diff
@@ -61,6 +62,8 @@
     'cliquet.storage_url': '',
     'cliquet.storage_max_fetch_size': 10000,
     'cliquet.userid_hmac_secret': '',
+    'cliquet.sentry_dsn': None,
```
leplatrem: Since we have ``storage_url`` and ``cache_url``, maybe we should be use ``sentry_url`` too ? ... I know that storage_dsn and cache_dsn would have been better... :(
We'll also have ``statsd_<dsn|url>`` :)

cliquet/__init__.py
leplatrem: So you don't want to allow disabling of Sentry via env ?

PR #114, opened 2015-03-19
Add some statsd timers on the storage and cache backends.

Natim:


leplatrem:
To be honest, I find this statsd very intrusive

I would have use the methods (``get_records``, etc.) that we have in resource instead no ?

leplatrem:
Or something like in the resource code: 
````
with statsd.counter('blala'):
     self.request.storage.get_all(...)
```

Natim:
Yes I choose the decorator but I can choose the contextmanager if you guys prefer, just a matter of taste.

I also like the fact that we can display the name of the backend in the statsd key.

leplatrem:
Yes sure!

It's mostly a matter of keeping this code confined in one place...

cliquet/statsd.py
```diff
@@ -0,0 +1,32 @@
...
+
+def dummy_decorator(f):
+    """Just call the given function."""
+    @wraps(f)
+    def wrapper(*args, **kwargs):
+        return f(*args, **kwargs)
+    return wrapper
+
+
+class StatsdClient(object):
```
ametaireau: I don't understand why this is a class since it only has class methods. A function would seem more appropriate.

Natim: Well it could be at the module level or at the class level, I prefer to have global variable on a class.

cliquet/statsd.py
```diff
@@ -0,0 +1,33 @@
...
+    global statsd
+    if statsd:
+        return statsd.timer(key_name)
+    return dummy_decorator
+
+
+def incr(key_name):
+    global statsd
+    if statsd:
+        return statsd.incr(key_name)
```
leplatrem: Shouldn't it be this way : 

```
class Client():
      ...

statsd = Client()
```

cliquet/storage/cloud_storage.py
```diff
@@ -81,11 +81,13 @@ def _build_headers(self, resource):
         }
 
     @wrap_http_error
+    @StatsdClient.timer('storage.cloud_storage.flush')
     def flush(self):
```
leplatrem: You may be able to able the decorated class and method names from the decorator

ametaireau: Oh yeah that's correct. Neat.

Natim:
Going forward with a metaclass.

Natim:
Here is what is timed:

```
Timing cache.CacheBase.set
Timing cache.CacheBase.get
Timing cache.CacheBase.flush
Timing cache.CacheBase.ping
Timing cache.CacheBase.expire
Timing cache.CacheBase.ttl
Timing cache.CacheBase.delete
Timing storage.StorageBase.get
Timing storage.StorageBase.get_all
Timing storage.StorageBase.collection_timestamp
Timing storage.StorageBase.create
Timing storage.StorageBase.delete_all
Timing storage.StorageBase.ping
Timing storage.StorageBase.update
Timing storage.StorageBase.flush
Timing storage.StorageBase.delete
```

cliquet/statsd.py
```diff
@@ -0,0 +1,54 @@
...
+        return f(*args, **kwargs)
+    return wrapper
+
+
+class Client(object):
+    statsd = None
+
+    @classmethod
+    def setup_client(cls, settings):
+        if settings['cliquet.statsd_endpoint'] is not None:
```
leplatrem: Maybe this test should be outside, before calling ``setup_client`` cliquet/init

cliquet/cache/postgresql/__init__.py
```diff
@@ -40,6 +40,9 @@ class PostgreSQL(PostgreSQLClient, CacheBase):
         boost performances and bound memory usage (*work_mem per connection*).
 
     """
+
+    __metaclass__ = statsd.CacheTimer
```
leplatrem: Instead of requiring every cache/storage backend to have this metaclass, couldn't we do something directly in ``statsd.setup()`` ?

Like : 

```
def decorate_methods(backend, prefix=''):
      members = dir(backend)
      for key, value in members.items():
          ...
          setattr(backend, key, decorated)

statds.decorate_methods(config.registry.storage, prefix='storage')
statds.decorate_methods(config.registry.cache, prefix='cache')
```

This way we set it up only if activated, and we don't have to put on every backend class.

What do you think ?

cliquet/__init__.py
```diff
@@ -61,6 +62,7 @@
     'cliquet.storage_url': '',
     'cliquet.storage_max_fetch_size': 10000,
     'cliquet.userid_hmac_secret': '',
+    'cliquet.statsd_endpoint': None,
```
leplatrem: ``statsd_url`` ? _dsn ?

leplatrem: To be mentionned in docs too :)

cliquet/__init__.py
```diff
@@ -191,17 +193,25 @@ def eos_tween(request):
...
+        msg = "{cliquet.project_name} {cliquet.project_version} starting."
+        raven_client.captureMessage(msg % settings)
+
+
+def handle_statsd(config):
+    settings = config.get_settings()
+
+    if settings['cliquet.statsd_url']:
+        client = statsd.load_from_config(config)
+        client.watch_execution_time(config.registry.cache, prefix='cache')
```
Natim: GG good idea.

cliquet/storage/postgresql/__init__.py
```diff
@@ -138,6 +139,7 @@ def flush(self):
         query = """
         DELETE FROM deleted;
         DELETE FROM records;
+        DELETE FROM metadata;
```
Natim: What is this for?

PR #121, opened 2015-03-20
Optional monitoring requirements ?

leplatrem:
/me wondering...

Since we are running into deployment problems because of Raven code base under python 2.7...
https://github.com/getsentry/raven-python/issues/546


ametaireau:
Cutting down optional dependencies seems like a good idea, thanks!

leplatrem:
We'll have to be careful with readinglist deps on upgrade!

Issue #124, opened 2015-03-20
postgresql for kinto and redis for cliquet ?


leplatrem:
* We have redis by default in settings
* We have postgresql in kinto test config

Is that on purpose ?

Natim:
So the decision made about it seems to be:
- Redis backend by default for cliquet
- Postgresql backend by default for kinto

PR #127, opened 2015-03-20
Add cache in memory for convenience

leplatrem:


ametaireau:
I actually would like to challenge the usage of in-memory backends. I tend to think this is more code to maintain for something no-one will ever need (appart tests maybe, but running from redis will be even faster).

leplatrem:
I like the idea of frontend developpers, cloning the project and running it locally with a simple ``make serve`` (btw see #124)

You may be right, the solution could be to use https://pypi.python.org/pypi/redislite instead ?

cliquet/tests/test_authentication.py
```diff
@@ -59,7 +59,7 @@ def test_views_are_forbidden_if_oauth2_scope_mismatch(self):
 
 class TokenVerificationCacheTest(unittest.TestCase):
     def setUp(self):
-        cache = redis_backend.Redis(max_connections=1)
+        cache = memory_backend.Memory(max_connections=1)
```
Natim: You do not need the ``max_connections`` here

ametaireau:
That sounds less code to maintain at least!

ametaireau:
Are we okay to close this?

Natim:
I'd rather merge it.

Natim:
Lets merge it.

PR #143, opened 2015-03-23
Setting paginate_by overrides limit (fixes #129)

### LINKED ISSUE #129: If paginate_by is set, limit should be ignored if superior
Currently, I can have ``paginate_by=100`` in configuration, and still fetch with ``_limit=100000``

leplatrem:


leplatrem:
@Natim r?

Natim:
LGTM

PR #145, opened 2015-03-23
Fix basic auth ofuscation and prefix (fixes #128)

### LINKED ISSUE #128: With Basic auth user ids are not prefixed with ``basicauth_`` as expected
```
http POST http://localhost:8000/v1/articles title=toto added_by=toto url=http://mozilla.com --auth mat:
```

```
postgres=# SELECT user_id FROM records;
 user_id 
---------
 mat
(1 row)
```

leplatrem:


cliquet/tests/test_authentication.py
```diff
@@ -79,3 +80,28 @@ def test_set_expires_the_value(self):
...
+
+class BasicAuthenticationPolicyTest(unittest.TestCase):
+    def setUp(self):
+        self.policy = authentication.BasicAuthAuthenticationPolicy()
+        self.request = DummyRequest()
+        self.request.headers['Authorization'] = 'Basic bWF0Og=='
+        self.request.registry.settings['cliquet.basic_auth_enabled'] = True
+
+    def test_prefixes_users_with_basicauth(self):
+        user_id = self.policy.unauthenticated_userid(self.request)
```
Natim: what about ``authenticated_userid``?

leplatrem: you mean the attribute on the request ?

It is set elsewhere in pyramid I guess....

Natim:
ok looks good to me.

cliquet/authentication.py
```diff
@@ -12,37 +12,41 @@
...
-
-class BasicAuthAuthenticationPolicy(base_auth.BasicAuthAuthenticationPolicy):
     def __init__(self, *args, **kwargs):
-        super(BasicAuthAuthenticationPolicy, self).__init__(check_credentials,
+        noop_check = lambda *a: [Authenticated]  # NOQA
+        super(BasicAuthAuthenticationPolicy, self).__init__(noop_check,
                                                             *args,
                                                             **kwargs)
 
+    def unauthenticated_userid(self, request):
```
ametaireau: `unauthenticated_userid` should be called only when we've went trough the authentication procedure already, and it's supposed to returned a cached version of the user id.

PR #150, opened 2015-03-25
Rely on Pyramid API to build pagination Next-Url (really fixes #147)

### LINKED ISSUE #147: Server port is lost in Next-Page URL
https://github.com/mozilla-services/readinglist/pull/120#discussion_r24584706

leplatrem:
really fixes #147

### DIFF
--- a/cliquet/utils.py
+++ b/cliquet/utils.py
@@ -109,11 +109,21 @@ def reapply_cors(request, response):
     recreating the response from scratch.
 
     """
+    service = current_service(request)
+    if service:
+        request.info['cors_checked'] = False
+        response = cors.ensure_origin(service, request, response)
+    return response
+
+
+def current_service(request):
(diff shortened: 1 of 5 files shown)


cliquet/utils.py
```diff
@@ -109,11 +109,21 @@ def reapply_cors(request, response):
...
+    return response
+
+
+def current_service(request):
+    """Return the Cornice service matching the specified request.
+
+    :returns: the service or None if unmatched.
+    :rtype: cornice.Service
+    """
     if request.matched_route:
```
ametaireau: I wonder if this shouldn't leave in cornice instead?

ametaireau:
I like the approach taken here, and the produced code is more readable ;)

cliquet/utils.py
leplatrem: that looked like a safety check, I left it :)

PR #151, opened 2015-03-25
Get rid of useless http_scheme setting

leplatrem:
I want to remove the ``cliquet.http_scheme`` setting for the following reasons:

* Less is more
* Tests did not break when I took it out (*untested code does not exist.*)
* It is *cliquet* specific, and gives the impression that we do some magic with it, whereas WSGI and Pyramid already manage this stuff
* It is unclear whether changing requests attributes will truely override Pyramid internal machinery on URLs (which partly relies on WSGI environs)
* It was not at the right place (``attach_http_objects``)

I suggest we had some documentation and highlight pitfalls of WSGI apps proxifying instead... such as for Nginx: 

```
        proxy_set_header        Host $http_host;
        proxy_set_header        X-Real-IP $remote_addr;
        proxy_set_header        X-Forwarded-For $proxy_add_x_forwarded_for;
        proxy_set_header        X-Forwarded-Proto $scheme;
```

### DIFF
--- a/cliquet/__init__.py
+++ b/cliquet/__init__.py
@@ -55,7 +55,6 @@ DEFAULT_SETTINGS = {
     'cliquet.eos_message': None,
     'cliquet.eos_url': None,
     'cliquet.logging_renderer': 'cliquet.logs.ClassicLogRenderer',
-    'cliquet.http_scheme': None,
     'cliquet.paginate_by': None,
     'cliquet.project_docs': '',
     'cliquet.project_name': '',
@@ -144,11 +143,6 @@ def attach_http_objects(config):
         # Attach objects on requests for easier access.
         event.request.db = config.registry.storage
         event.request.cache = config.registry.cache
-
-        # Force request scheme from settings.
-        http_scheme = config.registry.settings['cliquet.http_scheme']
-        if http_scheme:
-            event.request.scheme = http_scheme
 
     config.add_subscriber(on_new_request, NewRequest)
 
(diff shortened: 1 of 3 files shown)


ametaireau:
In general (not for the scheme only), experience is telling me that having this defined in a configuration somewhere works better than having it guessed by the app, because there are a number of ways this discovery can fail. 

Proxies can set headers in an incorrect way (it happened already), and sometimes it's done by code you don't control (think AWS).

Also, I believe explicit is better than implicit, so we could try to guess the correct behaviour, but having a way to ensure it's doing the right thing is IMO a good practice.

leplatrem:
I won't discuss your experience :) But I truely prefer "refer to Pyramid documentation" instead of "we are doing some black magic for you".

Anyway, this piece of code has no place in our codebase if it's not asserted that it truely forces the attributes. Especially if incoming headers or OS/wsgi environment variables are set.

ametaireau:
Oh, experience is relative, I don't hold the only one truth (too bad) ;)

You're correct, though about the fact this code isn't leaving at the right place. We should replace it with proper configuration of the service scheme / url.

Merging for now, and opening another issue for this.

Natim:
Actually with this PR, and since we are using an ELB, the WSGI stack will not detect the https in production nor stage. You can validate in dev that the home endpoint doesn't give you the right url anymore.

leplatrem:
@Natim that's why #159 was opened

Natim:
This was already our fix for #159

Natim:
Ok let's see if your fix #165 works

PR #155, opened 2015-03-25
Implement a simple schema migration system for PostgreSQL (fixes #139)

### LINKED ISSUE #139: Basic schema migration implementation in PG storage backend
I propose that we store a ``schema_version`` (simple integer) in the ``metadata`` table.

Then on startup, instead of checking the existance of the ``records`` like we do currently, we compare the schema versions.

For each missing intermediary version, we run the script ``migration_N_N+1.sql``. Each migration script would update the ``schema_version`` value.

The ``schema.sql`` would always contain the last version of the schema and end up by setting the ``schema_version`` value.

leplatrem:


ametaireau:
And of course I commented on the wrong one :p

leplatrem:
@Natim @ametaireau r?

Natim:
LGTM

cliquet/scripts/cliquet.py
```diff
@@ -2,10 +2,17 @@
 import argparse
 import sys
 import textwrap
+import warnings
 
 from pyramid.paster import bootstrap
 
 
+def deprecated_init(env):
+    message = '"cliquet init" is deprecated. Use "cliquet migrate" instead.'
```
ametaireau: migrate sounds weird for an initializer. I would have expected to have two commands: init and migrate (not related)

cliquet/storage/postgresql/__init__.py
```diff
@@ -135,27 +137,43 @@ def __init__(self, *args, **kwargs):
 
     def _execute_sql_file(self, filepath):
         here = os.path.abspath(os.path.dirname(__file__))
-        schema = open(os.path.join(here, 'schema.sql')).read()
+        schema = open(os.path.join(here, filepath)).read()
```
ametaireau: good catch!

ametaireau:
Neat.

Having a migration script is indeed something we want. I'm not sure about the rename of `init` in `migrate`.
What do you think about having two commands `cliquet initdb` and `cliquet migratedb`, the former could accept an optional ``--migrate`` option.

cliquet/scripts/cliquet.py
leplatrem: That's what Django has. ``manage.py migrate`` for everything :)

Do you suggest to separate the init process from the migration too ?

ametaireau:
I'll let you rebase and then you can merge :)

PR #156, opened 2015-03-25
[WIP] Use a materialized view for collection timestamps (fixes #138)

### LINKED ISSUE #138: Optimize collection timestamps in PG backend
Currently, the current collection timestamp is obtained using a ``SELECT`` on the records and deleted tables. It is performed for each request on the resource, and those tables can be pretty huge.

In order to optimize this (*since most requests will read and not write*), is to use materialized views of postgres.

```sql
CREATE MATERIALIZED VIEW collection_timestamp AS (
    -- see current SELECT in python code for collection_timestamp
    ...
);
CREATE UNIQUE INDEX idx_collection_timestamp ON collection_timestamp (user_id, resource_name);
```

In python code, we read a cached version, with a lot less records (only one per user per collection):

```sql
SELECT timestamp
   FROM collection_timestamp
   WHERE user_id = X AND resource_name = Y
```

In the ``bump_timestamp()`` SQL function, we refresh the materialized view:

```sql
REFRESH MATERIALIZED VIEW collection_timestamp;
```

leplatrem:
Currently facing a deadlock (?)

Depends on PR #155

### DIFF
--- a/cliquet/storage/postgresql/__init__.py
+++ b/cliquet/storage/postgresql/__init__.py
@@ -103,7 +103,7 @@ class PostgreSQL(PostgreSQLClient, StorageBase):
 
     """
 
-    schema_version = 1
+    schema_version = 2
 
     def __init__(self, *args, **kwargs):
         self._max_fetch_size = kwargs.pop('max_fetch_size')
--- a/cliquet/storage/postgresql/__init__.py
+++ b/cliquet/storage/postgresql/__init__.py
@@ -103,31 +103,46 @@ class PostgreSQL(PostgreSQLClient, StorageBase):
 
     """
 
+    schema_version = 1
+
     def __init__(self, *args, **kwargs):
         self._max_fetch_size = kwargs.pop('max_fetch_size')
         super(PostgreSQL, self).__init__(*args, **kwargs)
         self._init_schema()
 
+    def _execute_sql_file(self, filepath):
+        here = os.path.abspath(os.path.dirname(__file__))
+        schema = open(os.path.join(here, 'schema.sql')).read()
+        with self.connect() as cursor:
+            cursor.execute(schema)
+
     def _init_schema(self):
-        """Create PostgreSQL tables, only if not exists.
+        """Create PostgreSQL tables, and run necessary schema migrations.
 
         :note:
             Relies on JSON fields, available in recent versions of PostgreSQL.
         """
-        # Since indices cannot be created with IF NOT EXISTS, inspect.
-        query = """
-        SELECT *
-          FROM pg_tables
-         WHERE tablename = 'records';
-        """
-        with self.connect() as cursor:
-            cursor.execute(query)
-            exists = cursor.rowcount > 0
-
-        if exists:
-            logger.debug('Detected PostgreSQL storage tables')
+        version = self._get_installed_version()
+        if not version:
+            # Create full schema.
+            self._check_database_encoding()
+            self._execute_sql_file('schema.sql')
+            logger.info('Created PostgreSQL storage tables.')
             return
 
+        logger.debug('Detected PostgreSQL schema version %s.' % version)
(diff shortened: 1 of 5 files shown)


cliquet/storage/postgresql/__init__.py
```diff
@@ -103,31 +103,46 @@ class PostgreSQL(PostgreSQLClient, StorageBase):
...
-            logger.debug('Detected PostgreSQL storage tables')
+        version = self._get_installed_version()
+        if not version:
+            # Create full schema.
+            self._check_database_encoding()
+            self._execute_sql_file('schema.sql')
+            logger.info('Created PostgreSQL storage tables.')
             return
 
+        logger.debug('Detected PostgreSQL schema version %s.' % version)
```
ametaireau: Can we put this logic in a different method?

cliquet/storage/postgresql/migration_1_2.sql
```diff
@@ -0,0 +1,19 @@
+ALTER FUNCTION as_epoch(TIMESTAMP) IMMUTABLE;
```
ametaireau: I think it's wise to actually put these in another level (a schema folder ?)

ametaireau:
That's pretty cool to see schema migration hapening. I would love to have inputs from @andymckay and @magopian on this, since they were using it for AMO / Marketplace and I think they had some issues regarding this kind of migrations?

magopian:
From our experience using a custom tool ([schematic](https://github.com/mozilla/schematic)), i would advise against it.

The main problems we encountered were:
- migrations are written manually, and are thus error prone
- no easy/automatic way to revert a migration (django south provides one for example)
- it depends on the file name (and the number in it) to keep track of the latest migration that ran (which is a problem if you have migrations in different PRs that you don't merge in the correct order, or that run in the wrong order)

There are a lot of corner cases and hard problems to fix while creating such a migration tool, so even though I don't know what solutions are already existing (appart from [nashvegas](https://github.com/paltman/nashvegas) and [south now in django's core](https://docs.djangoproject.com/en/dev/topics/migrations/), both specific to django), I would highly recommend not implementing your own system.

cliquet/storage/postgresql/__init__.py
```diff
@@ -103,31 +103,46 @@ class PostgreSQL(PostgreSQLClient, StorageBase):
...
+        migrations = [(v, v + 1) for v in range(version, self.schema_version)]
+        if not migrations:
+            logger.info('Schema is up-to-date.')
+
+        for migration in migrations:
+            logger.info('Migrate schema from version %s to %s.' % migration)
+            filepath = 'migration_%s_%s.sql' % migration
```
leplatrem: Pad numbers ? ``001`` ?

cliquet/storage/postgresql/__init__.py
```diff
@@ -103,31 +103,46 @@ class PostgreSQL(PostgreSQLClient, StorageBase):
...
+            self._execute_sql_file(filepath)
```
leplatrem: As @magopian pointed out, adding an assertion on the current version would be useful, to prevent skipping numbers

cliquet/storage/postgresql/migration_2_3.sql
```diff
@@ -0,0 +1,58 @@
+CREATE OR REPLACE FUNCTION resource_timestamp(uid VARCHAR, resource VARCHAR)
```
leplatrem: Assert here that version is 2

cliquet/storage/postgresql/__init__.py
ametaireau: +1 that's easier to look at the migration scripts when in a FS.

leplatrem:
Closing this since it adds complexity and indirections.

Also, it looks like some fairly good results can be obtained on collection timestamps using an ordered index on last modified http://www.postgresql.org/docs/8.3/static/indexes-ordering.html

PR #162, opened 2015-03-26
Add a prefix to statsd requests

ametaireau:


cliquet/__init__.py
```diff
@@ -71,6 +71,7 @@
     'cliquet.sentry_url': None,
     'cliquet.sentry_projects': '',
     'cliquet.statsd_url': None,
+    'cliquet.statsd_prefix': 'cliquet'
```
Natim: Do you want to add the env here directly or can we use it automatically after the prefix?

Shoudn't we use the project_name directly?

ametaireau: I'm not sure for the env, since I believe we have different stacks depending the env used.

ametaireau: We can fallback on project name if statsd_prefix isn't defined, good idea.

Natim: ok

docs/configuration.rst
```diff
@@ -128,6 +128,7 @@ StatsD metrics can be enabled (disabled by default):
 .. code-block :: ini
 
     cliquet.statsd_url = udp://localhost:8125
+    cliquet.statsd_prefix = yourproject
```
leplatrem: I would mention here that this is useful only if different than project_name :)

PR #170, opened 2015-03-27
Postgres optimizations

leplatrem:


leplatrem:
Tests fail because nosetests instantiates many tests cases. Since the storage test case instantiates a storage object, a great of connections is opened on the database.

Any idea how to control how many instances are created by nose ?

Natim:
It is working locally.

PR #174, opened 2015-03-27
Use the raven client to captureException on the 500 error page.

Natim:


cliquet/views/errors.py
```diff
@@ -82,6 +82,10 @@ def error(context, request):
         logger.critical(context.original, exc_info=True)
         return service_unavailable(context, request)
 
+    if hasattr(request, 'raven'):
+        ident = request.raven.get_ident(request.raven.captureException())
+        print("Exception captured in event number: %s" % ident)
```
leplatrem: with this we won't catch storage backend errors (ex. DB constraints etc.)

leplatrem: instead of print(), use logger.info() ?

cliquet/__init__.py
```diff
@@ -214,13 +214,19 @@ def eos_tween(request):
...
-            settings['cliquet.sentry_url'],
+            dsn=settings['cliquet.sentry_dsn'],
             include_paths=['cornice', 'cliquet'] + extra_projects,
             release=settings['cliquet.project_version'])
 
+        def on_new_request(event):
+            # Attach objects on requests for easier access.
+            event.request.raven = raven_client
+
+        config.add_subscriber(on_new_request, NewRequest)
```
leplatrem: Add a small test in test_initialization please

cliquet/__init__.py
```diff
@@ -71,7 +71,7 @@
     'cliquet.storage_pool_maxconn': 50,
     'cliquet.storage_max_fetch_size': 10000,
     'cliquet.userid_hmac_secret': '',
-    'cliquet.sentry_url': None,
+    'cliquet.sentry_dsn': None,
```
leplatrem: update docs then

(I personnally prefer _url everywhere)

cliquet/views/errors.py
Natim: Actually we catch everything that is raised in a view, since it is a webserver we should catch everything that is not related to server startup.

leplatrem: If a DB constraint is violated, it will be catched in the backend and a BackendError will be raised. And sentry will not be aware of it, because it is returned 4 LOC above

Natim: Ok so I can leave this but put it 4 LOC before is that right?

Natim: I think it would be much better to actually configure it with: http://raven.readthedocs.org/en/latest/integrations/logging.html or http://raven.readthedocs.org/en/latest/integrations/pyramid.html

I will rework that.

leplatrem:
You left it on purpose in monitoring ?


Otherwise, the documentation should probably be updated too :)

leplatrem:
Before merge : 
* remove SentryConfigurationTest``
* add a logger.info("...Application starting...") like before with raven client
* make sure we want to leave sentry in setup.py for monitoring deps

PR #179, opened 2015-03-30
Rename maxconn to size, and reduce from 50 to 10 (fixes #176)

### LINKED ISSUE #176: pool size too big
The current pool size goes up to 50x2.  postgresql RDS default max_connections is < 300, so we're busting it with 3 processes.

We should have lower default values.

leplatrem:
@ametaireau @Natim r?

### DIFF
--- a/cliquet/storage/postgresql/__init__.py
+++ b/cliquet/storage/postgresql/__init__.py
@@ -1,5 +1,6 @@
 import contextlib
 import os
+import warnings
 from collections import defaultdict
 
 import psycopg2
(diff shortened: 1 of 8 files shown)


cliquet/storage/postgresql/__init__.py
```diff
@@ -614,10 +623,10 @@ def load_from_config(config):
     settings = config.get_settings()
 
     max_fetch_size = settings['cliquet.storage_max_fetch_size']
-    pool_maxconn = int(settings['cliquet.storage_pool_maxconn'])
+    pool_size = int(settings['cliquet.storage_pool_size'])
     uri = settings['cliquet.storage_url']
     uri = urlparse.urlparse(uri)
-    conn_kwargs = dict(max_connections=pool_maxconn,
+    conn_kwargs = dict(pool_size=pool_size,
```
ametaireau: More understandable with pool_size.

Issue #177, opened 2015-03-28
"tuple concurrently updated" when using several processes


tarekziade:
looks like concurrent ALTER ROLE during initiatization of workers are not working well

```
{"EnvVersion":"2.0","Severity":7,"Timestamp":1427501395940000000,"Hostname":"xxx",
"Pid":10952,"Fields":{},"Logger":"readinglist","Type":"ALTER ROLE rainbo SET TIME ZONE 'UTC';"}
Mar 28 00:09:55 localhost uwsgi: 2015-03-28 00:09:55,942 ERROR [cliquet.storage.postgresql][uWSGIWorker1Core0] {"EnvVersion":"2.0","Severity":2,"Timestamp":1427501395940000000,"Hostname":"ip-172-31-55-178","Pid":10952,"Fields":{},"Logger":"readinglist","Type":{"args":["tuple concurrently updated\n"],"cursor":{"arraysize":1,"binary_types":null,"closed":false,"connection":
```

PR #180, opened 2015-03-30
Warn if PostgreSQL user is not UTC (ref #177)

leplatrem:
Instead of altering user on each startup, check and warns during first schema creation.

Contrary to UTF8 setting, this is not a mandatory setting, but a recommandation.


Natim:
:)

leplatrem:
@Natim @ametaireau r?

PR #181, opened 2015-03-30
Add command to initialize schema (fixes #178)

### LINKED ISSUE #178: Separate backend initialization from storage.
Currently, the backend initialization phase is taking place when the backend starts. This causes a problem when we're running in a multiple workers environment, because they all try to create a schema, and this creates a deadlock.

One way to solve this is to split-out the backend initialization phase from the starting of the backend.

leplatrem:
* [x] Add method initialize_schema on each backend
* [x] Add command ``init_schema`` that calls initialize_schema on configured cache and storage backend
* [x] Prevent tween declaration to bind errno value on logger


setup.py
```diff
@@ -38,6 +38,13 @@
     'statsd',
 ]
 
+ENTRY_POINTS = {
+    'console_scripts': [
+        'init_schema = cliquet.scripts.init_schema:main'
```
ametaireau: I believe this will create a `init_schema` command, where I think it should be prefixed (at least) with `cliquet`.

cliquet/scripts/init_schema.py
```diff
@@ -0,0 +1,37 @@
...
+    """
+    logging.basicConfig(level=logging.DEBUG)
+    logger = logging.getLogger(__name__)
+
+    usage = "usage: %prog config_uri"
+    parser = optparse.OptionParser(usage=usage,
+                                   description=textwrap.dedent(description))
+    options, args = parser.parse_args(sys.argv[1:])
+
+    if not len(args) >= 1:
```
ametaireau: Should we use argparse instead? It could be handy with the subparsers so we can have multiple sub commands, and do something like `cliquet init`, `cliquet migrate` etc.

ametaireau:
Cool, thanks.

PR #184, opened 2015-03-31
Add heartbeat status for cache and oauth

Natim:
Refs mozilla-services/readinglist#229

### DIFF
--- a/cliquet/tests/test_views_heartbeat.py
+++ b/cliquet/tests/test_views_heartbeat.py
@@ -1,4 +1,5 @@
 import mock
+import requests
 
 from .support import BaseWebTest, unittest
 
@@ -9,6 +10,14 @@ class HeartBeatViewTest(BaseWebTest, unittest.TestCase):
         response = self.app.get('/__heartbeat__')
         self.assertEqual(response.json['database'], True)
 
+    def test_returns_cache_true_if_ok(self):
+        response = self.app.get('/__heartbeat__')
+        self.assertEqual(response.json['cache'], True)
+
+    def test_returns_oauth_true_if_ok(self):
+        response = self.app.get('/__heartbeat__')
+        self.assertEqual(response.json['oauth'], True)
+
     @mock.patch('cliquet.storage.memory.Memory.ping')
     @mock.patch('cliquet.storage.redis.Redis.ping')
     def test_returns_database_false_if_ko(self, *mocked):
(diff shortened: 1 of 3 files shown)


cliquet/views/heartbeat.py
```diff
@@ -1,5 +1,8 @@
+import requests
 from cornice import Service
 from pyramid.security import NO_PERMISSION_REQUIRED
+from six.moves.urllib.parse import urljoin
+from fxa.oauth import Client as OAuthClient
```
leplatrem: Move up to order imports alphabetically

cliquet/views/heartbeat.py
```diff
@@ -9,9 +12,27 @@
...
+        auth_client = OAuthClient(server_url=server_url)
+        server_url = auth_client.server_url
+        oauth = False
+
+        try:
+            r = requests.get(urljoin(server_url, '/__heartbeat__'), timeout=10)
+            r.raise_for_status()
+            oauth = True
+        except requests.exceptions.HTTPError:
+            pass
```
leplatrem: In order to keep this heartbeat agnostic, I would try to move this in the authentication module.

I'm not sure we can do something like this:

``authent = registry.authent_policy.ping()``

Where ``ping()`` would be a method on ``Oauth2AuthenticationPolicy`` class.


Otherwise, maybe just define a ``fxa_ping()`` method in ``cliquet.authentication`` to keep things all together there.

What do you think ?

Natim: Yep

cliquet/tests/test_views_heartbeat.py
```diff
@@ -16,3 +25,17 @@ def test_returns_database_false_if_ko(self, *mocked):
...
+            mock_instance.return_value = False
+        response = self.app.get('/__heartbeat__', status=503)
+        self.assertEqual(response.json['cache'], False)
+
+    @mock.patch('requests.get')
+    def test_returns_oauth_false_if_ko(self, *mocked):
+        for mock_instance in mocked:
+            mock_instance.side_effect = requests.exceptions.HTTPError()
+        response = self.app.get('/__heartbeat__', status=503)
+        self.assertEqual(response.json['oauth'], False)
```
leplatrem: I see one feature that you do not assert :) «returns none if not configured»

Natim: I added it here: https://github.com/mozilla-services/cliquet/pull/184/files#diff-f019564cfafcf39e7396d1e84e019e1dR43

PR #187, opened 2015-04-01
Rename cliquet.initialize_cliquet in cliquet.initialize

ametaireau:
r? @leplatrem @Natim

PR #189, opened 2015-04-01
Newrelic

ametaireau:
Based on #187

### DIFF
--- a/cliquet/__init__.py
+++ b/cliquet/__init__.py
@@ -9,6 +9,11 @@ import requests
 import structlog
 import webob
 
+try:
+    import newrelic.agent
+except ImportError:
+    pass  # NOQA
+
 from cornice import Service
 from pyramid.events import NewRequest, NewResponse
 from pyramid.httpexceptions import HTTPTemporaryRedirect, HTTPGone
@@ -316,3 +321,16 @@ def initialize(config, version=None, project_name=None):
     # Include cliquet views with the correct api version prefix.
     config.include("cliquet", route_prefix=api_version)
     config.route_prefix = api_version
+
+
+def wrap_app(app, settings):
+    "Wraps the passed wsgi application with wsgi middlewares."
+
+    # Setup new-relic.
+    if settings.get('cliquet.newrelic_config', False):
+        ini_file = settings.get('cliquet.newrelic_config')
+        env = settings.get('cliquet.newrelic_env', 'dev')
+        newrelic.agent.initialize(ini_file, env)
+        app = newrelic.agent.WSGIApplicationWrapper(app)
+
+    return app
--- a/cliquet/__init__.py
+++ b/cliquet/__init__.py
@@ -323,8 +323,8 @@ def initialize(config, version=None, project_name=None):
     config.route_prefix = api_version
 
 
-def wrap_app(app, settings):
-    "Wraps the passed wsgi application with wsgi middlewares."
+def install_middlewares(app, settings):
+    "Install a set of middlewares defined in the ini file on the given app."
 
     # Setup new-relic.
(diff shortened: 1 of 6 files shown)


cliquet/__init__.py
```diff
@@ -327,3 +340,16 @@ def initialize_cliquet(config, version=None, project_name=None):
     # Include cliquet views with the correct api version prefix.
     config.include("cliquet", route_prefix=api_version)
     config.route_prefix = api_version
+
+
+def wrap_app(app, settings):
```
Natim: I don't like too much the name of this method, should we name it ``plug_newrelic_if_activated``

cliquet/__init__.py
```diff
@@ -327,3 +340,16 @@ def initialize_cliquet(config, version=None, project_name=None):
...
+    "Wraps the passed wsgi application with wsgi middlewares."
+
+    # Setup new-relic.
+    if settings.get('cliquet.newrelic_config', False):
+        ini_file = settings.get('cliquet.newrelic_config')
+        env = settings.get('cliquet.newrelic_env', 'dev')
```
leplatrem: Don't we want default value for this setting ? Even None ?

cliquet/__init__.py
leplatrem: Yes, maybe something like ``wrap_newrelic_app`` ?

cliquet/__init__.py
Natim: Here the default is ``dev`` isn't it enough?

cliquet/__init__.py
ametaireau: The idea is to have this wrap other things, not only new relic, and have the wrappers defined in a configuration, so we need an app wrapper for all the middlewares.

cliquet/__init__.py
ametaireau: We could have a default value for `'cliquet.newrelic_config` set to None, yes.

cliquet/tests/test_initialization.py
```diff
@@ -85,6 +85,67 @@ def test_initialize_cliquet_is_deprecated(self):
...
+
+    @mock.patch('cliquet.newrelic.agent')
+    def test_newrelic_is_not_included_by_default(self, mocked_newrelic):
+        settings = {}
+        app = cliquet.install_middlewares(mock.sentinel.app, settings)
+        mocked_newrelic.initialize.assert_not_called()
+        self.assertEquals(app, mock.sentinel.app)
+
+
+class SentryConfigurationTest(unittest.TestCase):
```
Natim: Why is this present here? It has been removed in master.

cliquet/__init__.py
leplatrem: Well, I believe we should keep up with gathering all default settings in one place. It's also a very easy way to get a list of all settings handled by cliquet :)

ametaireau: Yes, dunno why I haven't done that. PR updated.

docs/quickstart.rst
```diff
@@ -35,8 +35,8 @@ just add some extra initialization code:
         config = Configurator(settings=settings)
 
         cliquet.initialize(config, __version__)
-
-        return config.make_wsgi_app()
+        app = config.make_wsgi_app()
+        return cliquet.install_middlewares(app)
```
leplatrem: Question: don't we want to leave the wrapping only in the monitoring section ?

I fear one thing : the first impression of cliquet might be : "this is too much initialization magic", "too far from vanilla pyramid"

If one day we have general purpose middleware we will add it here, in the general introduction docs.

Thoughts ?

ametaireau: makes sense.

ametaireau:
updated.

PR #190, opened 2015-04-01
Improve code coverage

leplatrem:
Previous PR was merged without reaching coverage of 100%.

This should reach 100%.

It also fixes a bug in PostgreSQL delete_all().

A mistake was misleading us : ``assert_called()`` does not exist on Mock object, and always returns True.

Some parts were marked as ``no cover`` since they depend on environment configuration.


Natim:
Tests are failing with Python 3.4

Natim:
I had the wrong version of PyFxA alles claar.

PR #192, opened 2015-04-02
Fix tests after code coverage improvement merge

leplatrem:


leplatrem:
I took some time to speed-up the tests.

Currently the ping on FxA oauth in Kinto takes quite some time. Should we disable OAuth in Kinto config for storage tests ?

ametaireau:
Definitely, yes.

Makefile
```diff
@@ -28,7 +28,7 @@ $(PYTHON):
 	$(VIRTUALENV) $(VENV)
 
 tests-once: install-dev need-kinto-running
-	$(VENV)/bin/nosetests -s --with-mocha-reporter --with-coverage --cover-package=cliquet
+	$(VENV)/bin/nosetests -s --with-mocha-reporter --with-coverage --cover-min-percentage=100 --cover-package=cliquet
```
ametaireau: yep, this!

cliquet/views/oauth/relier.py
```diff
@@ -33,7 +33,7 @@ def persist_state(request):
     """
     state = uuid.uuid4().hex
     redirect_url = request.validated['redirect']
-    expiration = int(fxa_conf(request, 'cache_ttl_seconds'))
+    expiration = float(fxa_conf(request, 'cache_ttl_seconds'))
```
ametaireau: thanks.

Makefile
leplatrem: it fails because 100% coverage was lost after werkzeug middleware :)

.travis.yml
```diff
@@ -15,6 +15,10 @@ before_install:
     - git clone https://github.com/mozilla-services/kinto.git kinto
     - pushd kinto
     - make install-dev
+    # Use this cliquet.
+    - pip install newrelic
```
leplatrem: Why the hell ?

cliquet/tests/support.py
```diff
@@ -87,12 +87,9 @@ class BaseWebTest(FakeAuthentMixin):
 
     api_prefix = "v0"
 
-    def get_test_app(self):
```
ametaireau: This is useful to control what's the application used in the tests.

Is it the part that makes things slow? (eg. why do we want to remove this `get_test_app` utility?)

leplatrem: Before #193 it was not used and was executed by nose twice per test suite (because contains "test" in name)

I will restore it and prefix it with ``_``.

leplatrem:
Tests failed twice in a raw because : download error in docs, and resetted connection in cloud storage 

Can be merged.

PR #193, opened 2015-04-02
Convert settings booleans with pyramid.settings.asbool

ametaireau:
Related to https://github.com/mozilla-services/readinglist/issues/237

### DIFF
--- a/cliquet/tests/support.py
+++ b/cliquet/tests/support.py
@@ -90,22 +90,26 @@ class BaseWebTest(FakeAuthentMixin):
     def __init__(self, *args, **kwargs):
         super(BaseWebTest, self).__init__(*args, **kwargs)
         self.app = self._get_test_app()
-        self.app.RequestClass = get_request_class(self.api_prefix)
         self.db = self.app.app.registry.storage
         self.headers.update({
             'Content-Type': 'application/json',
         })
 
-    def _get_test_app(self):
-        return webtest.TestApp(testapp(self.get_app_settings()))
+    def _get_test_app(self, settings=None):
+        app = webtest.TestApp(testapp(self.get_app_settings(settings)))
+        app.RequestClass = get_request_class(self.api_prefix)
+        return app
 
-    def get_app_settings(self):
+    def get_app_settings(self, additional_settings=None):
         settings = DEFAULT_SETTINGS.copy()
         settings['cliquet.project_name'] = 'cliquet'
         settings['cliquet.project_docs'] = 'https://cliquet.rtfd.org/'
         settings['fxa-oauth.relier.enabled'] = True
         settings['fxa-oauth.oauth_uri'] = 'https://oauth-stable.dev.lcip.org'
         settings['fxa-oauth.webapp.authorized_domains'] = ['*.firefox.com', ]
+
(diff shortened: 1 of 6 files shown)


leplatrem:
Maybe we should assert that the Real header is not sent either ?

cliquet/tests/test_authentication.py
```diff
@@ -27,6 +27,26 @@ def test_basic_auth_is_accepted_if_enabled_in_settings(self):
...
+            self.app.get(self.sample_url, headers=headers, status=401)
+
+    def test_basic_auth_is_declined_if_unknown_value_in_settings(self):
+        auth_password = base64.b64encode('bob:secret'.encode('ascii'))
+        headers = {
+            'Authorization': 'Basic {0}'.format(auth_password.decode('ascii'))
+        }
+
+        with mock.patch.dict(self.app.app.registry.settings,
+                             [('cliquet.basic_auth_enabled', 'False')]):
```
ametaireau: It actually is.

ametaireau:
Failing test is `test_timestamps_are_unique`

cliquet/tests/test_authentication.py
Natim: http://docs.pylonsproject.org/docs/pyramid/en/latest/api/settings.html#pyramid.settings.asbool

Natim: Yes it is unknow in a sense that it is not ``True``

ametaireau:
updated.

Natim:
> Maybe we should assert that the ``Realm`` header is not sent either ?

As Alexis pointed out, we should probably not instanciate the BasicAuth authentication backend in case it is unactivated.

ametaireau:
I've changed how it's done and updated the tests. Hopefully ready to be merged!

ametaireau:
@Natim @leplatrem r?

cliquet/tests/support.py
```diff
@@ -86,25 +86,29 @@ class BaseWebTest(FakeAuthentMixin):
 
     api_prefix = "v0"
 
-    def get_test_app(self):
-        return webtest.TestApp(testapp(self.get_app_settings()))
+    def get_test_app(self, settings=None):
+        app = webtest.TestApp(testapp(self.get_app_settings(settings)))
+        app.RequestClass = get_request_class(self.api_prefix)
+        return app
```
leplatrem: The fact that this method has "test" in its name makes nose executing it as a test :(

Elsewhere we use patch.dict(), we should make it consistent to use the same method everywhere

I noticed that instantiating app was slow, and took it out in #192. This point makes this PR conflicting with #192 then

ametaireau: I though tests methods needed to start with *test_*. Interesting.

leplatrem: > I though tests methods needed to start with test_. Interesting.

So did I !

ametaireau: patch.dict() works in some cases but not in others.

For instance, if the settings are read at application startup, you'll need to create a new application, because patching will not work (the code you want to change the behaviour would have been called already).

I believe we don't want to remove the code I changed here in #192. I commented there.

PR #197, opened 2015-04-07
Bypass Colander schema binding on request

leplatrem:
I haven't felt very inspired by an API change in Cornice. So I did this quick and dirty fix.

Should it be done in readinglist instead ? Since only there we're sure that deferred will never be used. Or should it come optimized by default cliquet ?

@ametaireau @Natim Thoughts ?

### DIFF
--- a/cliquet/resource.py
+++ b/cliquet/resource.py
@@ -2,7 +2,7 @@ import re
 
 import colander
 from cornice import resource
-from cornice.schemas import CorniceSchema
+from cornice import schemas as cornice_schemas
 from pyramid.httpexceptions import (HTTPNotModified, HTTPPreconditionFailed,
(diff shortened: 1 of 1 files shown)


ametaireau:
I would think doing that in cornice would be the right place. Adding a specific argument that could be looked by cornice to deactivate the binding would do the optim trick (because of backwards we cannot change the current behaviour).

leplatrem:
That's this *"specific argument"* I couldn't figure out... On the Service class ? On the CorniceSchema class ?

leplatrem:
I have been reading cornice source code, and I see no obvious way to add this option :/ If you are inspired, please help me :)

leplatrem:
https://github.com/mozilla-services/cornice/pull/288

PR #202, opened 2015-04-08
Perform create/delete actions in heartbeat view

leplatrem:
ref mozilla-services/readinglist#229

@Natim @ametaireau @tarekziade r?

(*we do more with less code* \o/)

### DIFF
--- a/cliquet/cache/__init__.py
+++ b/cliquet/cache/__init__.py
@@ -3,7 +3,7 @@ import random
 
 _HEARTBEAT_DELETE_RATE = 0.5
 _HEARTBEAT_KEY = '__heartbeat__'
-_HEARTBEAT_TTL = 3600
+_HEARTBEAT_TTL_SECONDS = 3600
 
 
 class CacheBase(object):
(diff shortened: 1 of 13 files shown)


cliquet/cache/__init__.py
```diff
@@ -1,3 +1,11 @@
+import random
+
+
+_HEARTBEAT_DELETE_RATE = 0.5
+_HEARTBEAT_KEY = '__heartbeat__'
+_HEARTBEAT_TTL = 3600
```
Natim: ``_HEARTBEAT_TTL_SECONDS``

cliquet/cache/__init__.py
```diff
@@ -21,7 +29,14 @@ def ping(self):
         :returns: `True` is everything is ok, `False` otherwise.
         :rtype: boolean
         """
-        raise NotImplementedError
+        try:
+            if random.random() > _HEARTBEAT_DELETE_RATE:
```
Natim: ``random.random() < _HEARTBEAT_DELETE_RATE``

In case of 0.5 it is the same but not with 0.3

Natim:
LGTM

ametaireau:
I like the idea of letting backends decide what's the best way to "ping" them, and this let's it open as a possibility, via inheritance.

+1 from me.

PR #198, opened 2015-04-07
Use ordered index to improve PG performances (ref #138)

leplatrem:
ref #138

This changeset boosts insertion of records (x100 faster):

*before ~10ms per insert*

    testdb=# INSERT INTO records(user_id, resource_name, data) SELECT * FROM generate_records(10000, 1, 10);
    INSERT 0 100000
    Time: 1102111,033 ms

*after ~0.1 ms per insert*

    testdb=# INSERT INTO records(user_id, resource_name, data) SELECT * FROM generate_records(100000, 1, 10);
    INSERT 0 1000000
    Time: 109397,719 ms

@Natim @ametaireau could you give a shot locally and see if you notice the same (amazing) improvements ?

Original notes here : https://etherpad.mozilla.org/reading-list-perf-postgres

PR #208, opened 2015-04-09
Get rid of uuid in postgresql (fixes #71, fixes #152)

### LINKED ISSUE #71: Error out when doing a put with something else than a UUID formed id
Currently, we don't validate the id in passed in the URL (``/resource/<id>``). And it is used as it is given by the storage backend.

One approach would be to add a regexp and validate it at the resource level.

### LINKED ISSUE #152: Use an id_generator for records ids in cliquet.
Currently, we're generating UUIDs for the records in the postgres backend, but the exposed HTTP API isn't requiring UUIDs (For instance it's possible to do a PUT, letting the user agent chose the id).

Cliquet should let the developer chose the ID generator they want to use.

- [ ] Pass an `id_generator` argument to the storage engines;
- [ ] Create a `random_uuid` generator;
- [ ] Make it possible to configure the `id_generator` to use at the resource level (since storage engine will be passed the resource when called it's possible to have a different strategy per resource)
- [ ] Add documentation explaining how to create and plug an id generator.

Note: we should make sure to define validation logic for the IDs in the readinglist project.

leplatrem:
Based on #198, review will be easier through 2 last commits


cliquet/storage/__init__.py
```diff
@@ -1,7 +1,11 @@
 import random
+import re
 from collections import namedtuple
 
 
+RECORD_ID_REGEXP = re.compile(r'^[a-zA-Z0-9\-_\.]{1,36}$')
```
Natim: I am not sure ``_`` and ``.`` should be present here.

Natim:
Don't we have a better type than VARCHAR for the UUID? a BIN or INT maybe?

Natim:
http://simononsoftware.com/how-to-store-uuids-in-postgresql/

leplatrem:
http://blog.jonanin.com/2013/11/20/postgresql-char-varchar/

Natim:
Ok let's go for TEXT instead of VARCHAR then

Natim:
LGTM

cliquet/storage/generators.py
```diff
@@ -0,0 +1,59 @@
+import re
+from uuid import uuid4
+
+import six
+
+
+class Generator(object):
```
ametaireau: We should mark somewhere this is a base class.

cliquet/storage/redis.py
```diff
@@ -92,7 +91,7 @@ def create(self, resource, user_id, record):
         self.check_unicity(resource, user_id, record)
 
         record = record.copy()
-        _id = record[resource.id_field] = self.id_generator()
+        _id = record[resource.id_field] = resource.id_generator()
```
ametaireau: I like how we make the backend use the resource class for specific operations, it's a really nice extension point.

ametaireau:
This looks good to me also.

ametaireau:
(Soon, we will have an option to have nicely named ids ;))

PR #209, opened 2015-04-09
Get rid of redondant indices

leplatrem:
Based on #208, review should be easier on last commit only.

------

Indices were not restricted to those used in queries. Indexing an individual column if it is not used in queries is a waste. This PR removes redundants indices and make sure queries target the existing indices on tuples (id, user_id, resource_name) and (user_id, resource_name, last_modified).

In addition to cleaner schema, it improves creation significantly (~35%):

*before*

    testdb=# INSERT INTO records(id, user_id, resource_name, data) SELECT uuid_generate_v4(), * FROM generate_records(100000, 1, 10);
    INSERT 0 1000000
    Time: 114463,644 ms

*after*

    INSERT INTO records(id, user_id, resource_name, data) SELECT uuid_generate_v4(), * FROM generate_records(100000, 1, 10);
    INSERT 0 1000000
    Time: 73150,822 ms

### DIFF
--- a/cliquet/storage/postgresql/__init__.py
+++ b/cliquet/storage/postgresql/__init__.py
@@ -538,11 +538,19 @@ class PostgreSQL(PostgreSQLClient, StorageBase):
         holders = {}
         for i, filtr in enumerate(filters):
             value = filtr.value
+            sql_operator = operators.setdefault(filtr.operator, filtr.operator)
+
+            if filtr.field == resource.modified_field:
+                # Value is integer (asserted in resource).
+                assert isinstance(filtr.value, six.integer_types)
+                # Bypass field and value escaping in order to use a function.
+                cond = "last_modified %s from_epoch(%s)" % (sql_operator,
+                                                            filtr.value)
+                conditions.append(cond)
+                continue
 
             if filtr.field == resource.id_field:
                 sql_field = 'id'
-            elif filtr.field == resource.modified_field:
-                sql_field = 'as_epoch(last_modified)'
             else:
                 # Safely escape field name
                 field_holder = '%s_field_%s' % (prefix, i)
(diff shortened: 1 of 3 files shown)


leplatrem:
I reverted the approach I had to get rid of the ``as_epoch(last_modified)`` index. It was introducing instabilities because of milliseconds rounding : GT comparison was sometimes becoming inclusive.

If one day we want to get rid of this index, we'll have to switch the last_modified column to integer. Personnally I like some semantics in databases schemas, and timestamps is fine :)

leplatrem:
Before

```
Duration: 600.07 seconds
Hits: 42103
Approximate Average RPS: 70
Average request time: 0.28s
```

After

```
Duration: 600.07 seconds
Hits: 45737
Approximate Average RPS: 76
Average request time: 0.26s
```

leplatrem:
@Natim @ametaireau  r?

ametaireau:
So there is no way to have a precision of MS with using the epoch postgres semantic?

leplatrem:
yes of course there is. I don't understand your question.

(*the problem with the from_epoch() approach, converting integer to timestamp, is that precisely it is round to the closest millesecond, where as the timestamp precision goes beyond that*)

ametaireau:
Yep, sorry for the bad phrasing here. Question is more "isn't it possible to have something that goes beyond milliseconds when using the from_epoch() approach?"

leplatrem:
Currently the timestamp is an integer representing number of milliseconds since epoch.
The only way to go beyond milliseconds with from_epoch() is to switch to microseconds timestamps or to switch to milliseconds with decimals instead of integers.
Could be another ticket.

ametaireau:
Yes, let's merge this for now and change that later if we want.

PR #219, opened 2015-04-13
[WIP] Warn if database schema mismatch

leplatrem:
I made a try to add an assertion on startup but it does not work (event not received)


leplatrem:
Since this event is also triggered when running a command (via Pyramid bootstrap), I don't see we can achieve this cleanly... closing.

PR #221, opened 2015-04-13
Fix migration behaviour when metadata table was flushed.

leplatrem:
Reading List tests could not be run several times in a row because the metadata table was flushed, and migrations were failing while being applied on the last schema version.

We now consider the schema as the last version if metadata is empty.

### DIFF
--- a/cliquet/storage/postgresql/__init__.py
+++ b/cliquet/storage/postgresql/__init__.py
@@ -222,7 +222,16 @@ class PostgreSQL(PostgreSQLClient, StorageBase):
             if cursor.rowcount > 0:
                 return int(cursor.fetchone()['version'])
             else:
-                # In the first versions of cliquet, there was no migration.
+                # Guess current version.
+                query = "SELECT COUNT(*) FROM metadata;"
+                cursor.execute(query)
+                was_flushed = int(cursor.fetchone()[0]) == 0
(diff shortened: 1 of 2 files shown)


Natim:
Or maybe we should not empty the metadata:migration table rows?

leplatrem:
I had thought of that too and hesitated...
But in the end I prefered this fix since it does not introduce `` WHERE`` with specific value in ``flush()`` and will work in metadata table is flushed manually by mistake.

What do you think ?

Natim:
It feels strange to me to lose the migration state in the database. For me missing migration data means no migration at all not that all migration where done. Maybe our flushdb should remote the schema then?

leplatrem:
IHMO : No migration data means no information. Since applying migrations on the last schema version will fail, we consider that we have the last version and do nothing.

If ``flush()`` drops the tables, then it means that we have to recreate them all at each setUp()/tearDown(). It will be slow and useless since the schema is already there!

ametaireau:
I feel no migration data means we are on a clear state (no migration needs to be run). 
I also feel we're  dealing with this because the tests are flushing the database.

Flushing the database shouldn't happen that often in a real world scenario, and if it's the case, we should do it with a cliquet command, which could let the migration information in place.

PR #222, opened 2015-04-14
Disable Cornice schema request binding (fixes #172)

### LINKED ISSUE #172: Schema binding in Cornice is costly
When disabling request binding, the time spent in this part of the code drops from 25% to 6%.

See https://github.com/mozilla-services/cornice/blob/master/cornice/schemas.py#L34
And https://github.com/mozilla-services/cliquet/blob/master/cliquet/resource.py#L83

leplatrem:
Requires https://github.com/mozilla-services/cornice/pull/288

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -15,7 +15,7 @@ with open(os.path.join(here, 'CONTRIBUTORS.rst')) as f:
 
 REQUIREMENTS = [
     'colander',
-    'cornice >= 0.20',  # Fixes cors.
+    'cornice >= 1.0',  # Disable request binding.
     'python-dateutil',
     'pyfxa >= 0.0.6',  # Gevent support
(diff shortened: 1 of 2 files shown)


leplatrem:
@ametaireau r?

setup.py
```diff
@@ -15,7 +15,7 @@
 
 REQUIREMENTS = [
     'colander',
-    'cornice >= 0.20',  # Fixes cors.
+    'cornice >= 1.0',  # Disable request binding.
```
ametaireau: I believe it's 1.0.0 (not sure how the comparison works).

ametaireau:
LGTM.

ametaireau:
Build was failing for a dependency to structlog which actually exist, so it's a false negative. Merging.

PR #223, opened 2015-04-14
Move parts of protocol from utils to resource

leplatrem:
Remove one level of indirection and reduce set of utils.


Natim:
Ok why not r+ :)

ametaireau:
yes, if they're not used in any other place we don't have any reason to keep them.

PR #229, opened 2015-04-15
Change arrow direction in logo

leplatrem:
As suggested by @Natim


ametaireau:
I actually find this weird :) What does the arrow mean? Is it a force? Here if the ratchet goes to this direction, then the wheel can go two directions, which is what we don't want.

leplatrem:
I think I know the solution : make it both ways!

Natim:
Yes I like it :)

Natim:
On devrait l'afficher dans le README aussi.

PR #230, opened 2015-04-16
Fix regression of schema simplification

leplatrem:
Since id and last_modified are not part of the schema,
they should be explicitly allowed for sorting and
filtering.

(*I was seeing 400 errors in RL load tests*)

@Natim r?


Natim:
r+

Issue #215, opened 2015-04-13
Recover from DB IP change ?


leplatrem:
https://bugzilla.mozilla.org/show_bug.cgi?id=1153489

Changing the database IP could invalidate the opened connections and refresh the pool with new ones

leplatrem:
Implemented by SQLA engine (see #231)

PR #171, opened 2015-03-27
[wip] Add new setting to configure pool class

leplatrem:
* [x] Implement
* [ ] Write test after (booh)
* [ ] Add section in documentation

PR #231, opened 2015-04-16
Use SQLAlchemy for pools and connections instead of raw psycopg2 (fixes #186, #215, #171)

### LINKED ISSUE #186: pool overflow and other SQLA goodies
we should add overflow in our pool, plus more goodies from SQLA

- overflow
- max backlog while waiting for a connection from the pool.
- pool recycling

see http://docs.sqlalchemy.org/en/rel_0_9/core/pooling.html

leplatrem:
Fixes #186 and #215 (and #171)
* [x] Add sqlalchemy in deps
* [x] Wire-up slqa in existing code base
* [x] Add more pooling parameters
* [x] Add a note to make sure "postgresql" scheme now has a meaning
* [x] Check that changing pool class does not fail
* [x] Fix behaviour when error occurs (change mocks ?)
* [x] Replace low-level usage of cursor with sqla ~~high-level~~ engine object
* [ ] ~~Import sync code~~
* [ ] ~~Add some minimalistic pooling tests~~
* [x] Share engine instance among storage instances ?
* [x] Share pool instance between cache/storage ?
* [x] Mention in docs among "giants"
* [x] Document pooling parameters
* [x] Restore fsync=off fix


Natim:
@leplatrem What is the status of this PR, should we continue the work done here or close ?

leplatrem:
I believe we should continue.

elemoine:
Out of curiosity, is the goal of this work "just" to benefit from SQLA's pooling system? Or do you also plan to use SQLA's SQL Expression Language for queries (as the issue description seems to suggest)? I think it makes sense to go step and step, and only use the pooling system as the first step if using the pooling system is indeed your primary goal.

leplatrem:
Yes, the intention is to benefit from the pooling objects of SQLA.
Especially to cover the issues mentionned in the description.

This PR is stalled, but the work should go on in the next weeks when we'll put *Kinto* in production :)

> SQL Expression Language for queries [...] as the issue description seems to suggest

No I only had the ``engine`` objects in mind (necessary to recover database IP change for instance)

To be honest, I am not a fan of SQL abstractions when raw SQL is mixed-up (like for pgpsql functions).

elemoine:
Understood. Thanks for the answers.

leplatrem:
I am currently reconsidering this pull-request.

Especially to introduce transaction tight to the requests life-cycle.

Good summary here : https://metaclassical.com/what-the-zope-transaction-manager-means-to-me-and-you/

elemoine:
> Especially to introduce transaction tight to the requests life-cycle.

Yep, a very useful thing.

leplatrem:
This is now ready for review !

At first I was sceptic about SQLA. But I realize that the connect code is less hacky, and the connection pools are smarter (especially for invalidation, #215).

This PR **does not** introduce:

* SQL abstraction or highlevel ORM
* Heterogenous backends like SQLite or whatever

The provided backend is specific for PostgreSQL. But if someone needs to write a backend using those fancy SQL abstractions, to manipulate relations or advanced data model, there are available :)

**In a another PR**, we will plug SQLA sessions and **per-request transactions**. The good news is that with two-phase commits, we will be able to have transaction accross distinct backends/databases.

Addition of custom pool class with *max backlog* option can be done in another PR, or as a contribution to SQLA :)

@Natim @tarekziade r?

(... @elemoine if you're not too busy ;))

cliquet/storage/postgresql/client.py
```diff
@@ -1,106 +1,102 @@
...
+        return PostgreSQLClient(_ENGINES[uri])
+
+    # Initialize SQLAlchemy engine.
+    options = {
+        'pool_size': settings.get(prefix + 'pool_size'),
+        'max_overflow': settings.get(prefix + 'pool_maxoverflow'),
+        'pool_recycle': settings.get(prefix + 'pool_recycle'),
+        'pool_timeout': settings.get(prefix + 'pool_timeout'),
+    }
+    # Filter specified values only, to preserve SQLAchemy defaults.
```
elemoine: typo: SQLAchemy -> SQLAlchemy

cliquet/storage/postgresql/client.py
```diff
@@ -1,106 +1,102 @@
...
+    # When fsync setting is off, like on TravisCI or in during development,
+    # some storage tests fail because commits are not applied
+    # accross every opened connections.
+    # XXX: find a proper solution to support fsync off.
+    # Meanhwile, disable connection pooling to prevent test suite failures.
+    if os.getenv('TRAVIS', False):  # pragma: no cover
+        warnings.warn('Option fsync = off detected. Disable pooling.')
+        options = {'poolclass': sqlalchemy.pool.NullPool}
+
+    # Store one engine per URI.
```
leplatrem: use `sqlalchemy.engine_from_config(configuration, prefix='sqlalchemy.')`

cliquet/storage/postgresql/__init__.py
```diff
@@ -105,9 +112,9 @@ def _check_database_timezone(self):
         # Make sure database has UTC timezone.
         query = "SELECT current_setting('TIMEZONE') AS timezone;"
         with self.client.connect() as cursor:
```
elemoine: Would it make sense to rename `cursor` to `connection` (or `conn`)?

elemoine:
I did not do a thorough review but I think it looks great!

cliquet/storage/postgresql/__init__.py
Natim: :+1: I like ``conn`` and ``connection``

Natim:
It looks great thanks.

leplatrem:
Tests with Pypy are stalled. And we run into tests instabilities because of `fsync=off`. `#discouraged`

elemoine:
> Tests with Pypy are stalled. And we run into tests instabilities because of fsync=off. #discouraged

Have you made it work finally?

leplatrem:
Yes, I disabled fsync=off to make sure data integrity is preserved during tests, and used a StaticPool to limit number of opened transactions during tests. I guess it also comes from the way nose instantiates the tests, but to be honest I didn't find out how to corner this beast in a nicer way...

Issue #107, opened 2015-03-19
Add a flag to disable the automatic redirection


ametaireau:
Currently there is an automatic redirection of all the resources which aren't prefixed by a version number.

The redirection happens in `cliquet.__init__`. To resolve this it's as easy as adding a new flag in the configuration and do not calling the redirection handler in this file.

The flag could be something like ``cliquet.version_prefix_redirect_enabled`` ?

hints:
- use `asbool` to cast the value of the config
- be sure to add tests
- be sure to add documentation
- default behaviour should be to let it on.

hiromipaw:
Hey, if it is ok for you I'd like to pick this up... please let me know...

ametaireau:
Sure, you can go ahead with this Silvia!

leplatrem:
Merged

PR #232, opened 2015-04-23
Add a flag to disable automatic redirection

hiromipaw:
Addressing issue #107.
Currently cliquet redirects all the resources which aren't prefixed by a version number.

The redirection happens in cliquet.\_\_init__. 

Therefore a new flag, cliquet.version_prefix_redirect_enabled, has been added to the configuration to enable the redirection through the config file.


hiromipaw:
Was not sure if automatic redirection should just not redirect or it shall return an error or something.

leplatrem:
Thanks  @hiromipaw ! 

Tests fail because of a style error in test_hello_view.

cliquet/__init__.py
```diff
@@ -69,6 +69,7 @@
     'cliquet.project_docs': '',
     'cliquet.project_name': '',
     'cliquet.project_version': '',
+    'cliquet.version_prefix_redirect_enabled': True,
```
leplatrem: we sorted the settings alphabetically, please move below :)

cliquet/__init__.py
```diff
@@ -124,23 +125,31 @@ def load_default_settings(config):
...
     # Redirect to the current version of the API if the prefix isn't used.
-    config.add_route(name='redirect_to_version',
-                     pattern='/{path:(?!%s).*}' % route_prefix)
+    # Do not redirect if cliquet.version_prefix_redirect_enabled is set to
+    # False.
+    settings = config.get_settings()
+    version_prefix_redirection_enabled = asbool(
+        settings['cliquet.version_prefix_redirect_enabled'])
+
+    if version_prefix_redirection_enabled:
```
leplatrem: I suggest that we do that on top, and exit the function (return) if enabled is False. This will prevent to have a long indented bloc of code

cliquet/__init__.py
```diff
@@ -124,23 +125,31 @@ def load_default_settings(config):
 def handle_api_redirection(config):
     """Add a view which redirects to the current version of the API.
     """
-    # Disable the route prefix passed by the app.
-    route_prefix = config.route_prefix
-    config.route_prefix = None
 
     def _redirect_to_version_view(request):
         raise HTTPTemporaryRedirect(
             '/%s/%s' % (route_prefix, request.matchdict['path']))
```
leplatrem: we could define this below, only if enabled

cliquet/tests/test_initialization.py
```diff
@@ -59,6 +59,14 @@ def test_set_the_project_name_from_settings_even_if_specified(self):
         self.assertEqual(config.registry.settings['cliquet.project_name'],
                          'kinto')
 
+    def test_set_version_prefix_redirect_to_false(self):
+        config = Configurator(
+            settings={'cliquet.version_prefix_redirect_enabled': False})
+        cliquet.initialize(config, '0.0.1', 'kinto')
+        self.assertEqual(
+            config.registry.settings[
+                'cliquet.version_prefix_redirect_enabled'], False)
```
leplatrem: use an intermediary variable to avoid extra indentation

cliquet/tests/test_views_hello.py
```diff
@@ -46,3 +46,11 @@ def test_redirect_to_version(self):
...
         finally:
             self.app.RequestClass = original_request_class
+
+    def test_do_not_redirect_to_version(self):
+        # GET on the hello view.
+        with mock.patch.dict(self.app.app.registry.settings,
+                [('cliquet.version_prefix_redirect_enabled', False)]):
+            response = self.app.get('/')
+            self.assertEqual(response.status_int, 200)
+            self.assertEqual(response.location, None)
```
leplatrem: :+1:

cliquet/tests/test_initialization.py
leplatrem: or use assertFalse btw

cliquet/__init__.py
hiromipaw: Sorry :) This was because before it was only called redirect_enabled

hiromipaw:
Will do the mods :+1:

leplatrem:
Thanks for adjusting so quickly! 

I know that take the risk that you get discouraged, but I believe we still can improve this patch!
I'll write a few more comments :)

cliquet/__init__.py
```diff
@@ -124,23 +125,34 @@ def load_default_settings(config):
...
+                        permission=NO_PERMISSION_REQUIRED)
 
-    config.add_view(view=_redirect_to_version_view,
-                    route_name='redirect_to_version',
-                    permission=NO_PERMISSION_REQUIRED)
+        config.route_prefix = route_prefix
 
-    config.route_prefix = route_prefix
+    else:
+        return
```
leplatrem: This statement does not bring anything. What I had in mind when I mentionned to invert the condition was this:

Replace : 

```python
if condition:
    # very
    # long
    # piece
    # of
    # indented
    # code
```

by : 

```python
if not condiition:
    return

# very
# long
# piece
# of
# code
```

cliquet/tests/test_views_hello.py
```diff
@@ -46,3 +46,12 @@ def test_redirect_to_version(self):
                              'http://localhost/v0/mushrooms')
         finally:
             self.app.RequestClass = original_request_class
+
+    def test_do_not_redirect_to_version(self):
```
leplatrem: If we consider the name of test as a spec (see http://blog.mathieu-leplatre.info/your-tests-as-your-specs.html), then the name of this test would be "do not redirect to version if disabled in settings"

cliquet/tests/test_initialization.py
```diff
@@ -59,6 +59,14 @@ def test_set_the_project_name_from_settings_even_if_specified(self):
         self.assertEqual(config.registry.settings['cliquet.project_name'],
                          'kinto')
 
+    def test_set_version_prefix_redirect_to_false(self):
+        config = Configurator(
+            settings={'cliquet.version_prefix_redirect_enabled': False})
+        cliquet.initialize(config, '0.0.1', 'kinto')
+        redirect_enabled = config.registry.settings[
+            'cliquet.version_prefix_redirect_enabled']
+        self.assertFalse(redirect_enabled)
```
leplatrem: I was reading this again, and I think that this test can be removed. Indeed it only tests that Pyramid is storing settings in registry, which isn't something in the scope of the pull-request :)

hiromipaw:
Sure @leplatrem I am totally up for improving this :+1:  !
I will get back to you with mods soon :smile:

hiromipaw:
I have also removed the initialization test. Although I got your comment about it by email I am not able to find it anymore...

cliquet/__init__.py
```diff
@@ -124,17 +125,26 @@ def load_default_settings(config):
...
             '/%s/%s' % (route_prefix, request.matchdict['path']))
 
-    # Redirect to the current version of the API if the prefix isn't used.
-    config.add_route(name='redirect_to_version',
-                     pattern='/{path:(?!%s).*}' % route_prefix)
+    # Disable the route prefix passed by the app.
+    route_prefix = config.route_prefix
+    config.route_prefix = None
+
+    config.add_route(name='redirect_to_version', pattern='/{path:(?!%s).*}'
```
leplatrem: IMO the original line folding was better (one line per argument)

cliquet/tests/test_initialization.py
leplatrem: You restored the test because of coverage ?

Then it means the other test (disabling the redirect) is not reaching your code properly, otherwise coverage would be 100% without this test.

cliquet/tests/test_views_hello.py
```diff
@@ -46,3 +46,12 @@ def test_redirect_to_version(self):
                              'http://localhost/v0/mushrooms')
         finally:
             self.app.RequestClass = original_request_class
+
+    def test_do_not_redirect_to_version_if_disabled_in_settings(self):
+        # GET on the hello view.
+        with mock.patch.dict(
+                self.app.app.registry.settings,
+                [('cliquet.version_prefix_redirect_enabled', False)]):
+            response = self.app.get('/')
```
leplatrem: I realize that this is using ``self.app`` as defined in ``support.py``. In other words, this request is sent with a version prefix.

The mock patch dict is performed too late, the app was already initialized with redirection enabled, the mocked settings aren't read.

I suggest that you create another app object from scratch without prefixed request class: 

```python
app = self._get_test_app({''cliquet.version_prefix_redirect_enabled': False})
app.RequestClass = TestRequest  # disable redirect
response = app.get('/')
...
...
```

hiromipaw:
Hey. Thanks a lot. I'll get back to you after the weekend. :+1:

docs/contributing.rst
```diff
@@ -45,17 +45,20 @@ OS X
...
 ::
+
     export LC_ALL=en_US.UTF-8
     export LANG=en_US.UTF-8
 
 Then run:
 
 ::
+
     $ source ~/.bash_profile
```
hiromipaw: Not sure this should be included this. But I had just noticed this formatting error.

leplatrem: Thanks! No problem :)

leplatrem:
Thanks @hiromipaw for this contribution!

ametaireau:
thanks @hiromipaw !

PR #233, opened 2015-04-27
Provide details in error responses

leplatrem:
@ametaireau r?


ametaireau:
haha, here we go! Finally, we need this info. I feel we would benefit from having this exposed at the root (e.g. have a list of errors) rather than in a details argument, but for compatibility purposes, let's do that :)

ametaireau:
(needs to be rebased)

PR #235, opened 2015-04-27
Return existing record instead of raising 409 on POST (fixes #75)

### LINKED ISSUE #75: Move handling of conflict with 200 from readinglist to cliquet
Currently, cliquet raises 409 on conflict on POST, and 200 with existing record in cliquet.

Since this is part of the protocol we should behave differently.

This would consist in take the readinglist view code and move it here  (along with its tests)

leplatrem:
fixes #75

### DIFF
--- a/cliquet/resource.py
+++ b/cliquet/resource.py
@@ -162,6 +162,10 @@ class BaseResource(object):
     def collection_post(self):
         """Collection ``POST`` endpoint: create a record.
 
+        If the new record conflicts against a unique field constraint, the
+        posted record is ignored, and the existing record is returned, with
+        a ``200`` status.
+
         :raises:
             :exc:`~pyramid:pyramid.httpexceptions.HTTPPreconditionFailed` if
             ``If-Unmodified-Since`` header is provided and collection modified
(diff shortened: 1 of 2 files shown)


cliquet/resource.py
```diff
@@ -416,16 +425,10 @@ def create_record(self, record):
...
-            if a unique field constraint is violated.
-
         :returns: the newly created record.
         :rtype: dict
         """
-        try:
-            return self.db.create(record=record, **self.db_kwargs)
-        except storage_exceptions.UnicityError as e:
-            self._raise_conflict(e)
+        return self.db.create(record=record, **self.db_kwargs)
```
ametaireau: this was renamed in storage.

ametaireau:
LGTM.

cliquet/resource.py
leplatrem: I don't understand. What was renamed ?

ametaireau: I though the `self.db` renaming in `self.storage` was already merged, but it's not the case.

leplatrem: Ah no because it depends on a PR waiting review/merge in Kinto https://github.com/mozilla-services/kinto/pull/32,  itself waiting for https://github.com/mozilla-services/kinto/pull/27

PR #241, opened 2015-04-29
Do not hide FxA errors (fixes mozilla-services/readinglist#70)

leplatrem:


cliquet/authentication.py
```diff
@@ -107,9 +109,11 @@ def _get_credentials(self, request):
         try:
             profile = auth_client.verify_token(token=auth, scope=scope)
             user_id = profile['user']
-        except fxa_errors.OutOfProtocolError:
+        except fxa_errors.OutOfProtocolError as e:
+            logger.error(e)
```
ametaireau: Should we add a test for this?

Natim: In what I recall we should also probably improve the FxA error message.

leplatrem:
Created issue in **cliquet-fxa**.

PR #242, opened 2015-04-29
Add OAuth scopes to principals (fixes #239)

### LINKED ISSUE #239: OAuth scopes in pyramid principals
Currently the scopes obtain from OAuth token are not exposed outside the authentication class. 

I propose that we add them to ``effective_principals`` to allow the implementation of an authorization policy that would rely on those scopes.

leplatrem:


leplatrem:
@Natim r?

cliquet/authentication.py
```diff
@@ -79,8 +80,28 @@ def __init__(self, config, realm='Realm'):
...
-        return user_id
+        profile = self._get_credentials(request)
+        if profile is None:
+            return None
+
+        # Attach scopes to request.
+        setattr(request, '_fxa_scopes', profile['scope'])
+
+        # Return user-id with prefix.
+        return 'fxa_%s' % profile['user']
```
Natim: We should use ``fxa:%s``

cliquet/authentication.py
```diff
@@ -79,8 +80,28 @@ def __init__(self, config, realm='Realm'):
...
+
+    def effective_principals(self, request):
+        """Provide OAuth scopes in principals, in order to allow authorizations
+        policies to rely on them.
+        """
+        principals = super(Oauth2AuthenticationPolicy,
+                           self).effective_principals(request)
+
+        # unauthenticated_userid() is always called before this.
+        scopes = getattr(request, '_fxa_scopes', [])
```
Natim: Finally scopes should not be added as principals because they cannot be trusted.

We will have the following principals format:

 - ``fxa:<userid>``
 - ``basicauth:<userid>``
 - ``groups:<bucket's_group_id>``
 - ``system:Authenticated``
 - ``system:Everyone``

Basically the user id will be ``authorization_scheme:userid``

Natim: Also we will use scope to reduce the permission of a given token.

So if Bob have got ``storage:blog:articles:read`` scope we want to restreint the permission for this token to read on the ``articles`` collection of the ``blog`` bucket.

leplatrem:
I don't think we should keep *capital/uppercase* in ``Authenticated`` and ``Everyone``.

I don't think we should oblige authentication policies to prefix the userid with the authentication scheme. I believe we just should just have ``<userid>`` like Pyramid does natively.

> Also we will use scope to reduce the permission of a given token.
>

Then what ? Should we track scopes or not ?

leplatrem:
More generally speaking, I don't think we should have formatting requirements regarding principals.
If we want to keep authentication pluggable, we should make no assumptions on formatting.

For example, we should not expect ``system:everyone``, nor ``group:<group_id>`` to be among principals.

I believe it should be the other way around : it will be the role of our authorization policy to transform and compare native Pyramid principals (and result of group finder function) to the principals required for a particular operation on a particular resource.

PR #248, opened 2015-05-06
More realistic epoch milliseconds value in example

michielbdejong:
This parameter should be given in milliseconds since 1970, probably? So this way it's maybe clearer (timestamp 1234567890.123 was in February 2009)


leplatrem:
Thanks!

PR #250, opened 2015-05-06
412 responses for resources ignore rest of collection

michielbdejong:
Reading the rest of the text, and assuming REST-approach applies, these look like probably copy-paste errors.

### DIFF
--- a/docs/api/resource.rst
+++ b/docs/api/resource.rst
@@ -413,7 +413,7 @@ HTTP Status Code
 ----------------
 
 * ``200 OK``: The record was deleted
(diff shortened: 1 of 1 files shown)


leplatrem:
Thanks Michiel for taking the time to fix this !

An honour to see you come here to be honest ;)

michielbdejong:
The other way around! :) It's great to see you guys are working on this, I found out through http://www.servicedenuages.fr/eco-systeme-et-stockage-generique.html. Will open a few more discussion tickets!

Issue #137, opened 2015-03-23
Add a flag for all options cliquet exposes in initialize_cliquet


ametaireau:
A lot of things are hapenning in the initialize_cliquet, and one would benefit from having these enabled / disabled, so we should pref everything.

leplatrem:
A thought: 

What if, instead of having a lot of settings named ``<feature>_enabled``, we had this : 

```ini
cliquet.initialization_steps =  cliquet.monkey_patch_json
                                cliquet.load_default_settings
                                cliquet.setup_logging
                                cliquet.handle_api_redirection
                                cliquet.setup_end_of_life
                                cliquet.setup_auth
                                cliquet.setup_statsd

```

Each hook would just receive the ``config`` object as parameter.

This way any project could add/replace/remove initialization setup steps :)

Natim:
I think @leplatrem was merged :)

leplatrem:
We didn't resolve exactly resolved this.

As @ametaireau suggested, the current implementation has drawbacks:

> Also, this approach doesn't allow to easily manipulate the steps, like
> we would do in python for instance (add, remove). I'm not sure what we
> should do about that (add more settings with a remove or add prefix?)
> but if we consider explicit is better than implicit then that should work out.

PR #255, opened 2015-05-08
Move initialization functions to dedicated module (ref #137)

leplatrem:
r?

-----------
I chose not to expose the configuration of steps in the documentation yet. Prove me wrong :)

### DIFF
--- a/cliquet/__init__.py
+++ b/cliquet/__init__.py
@@ -1,42 +1,19 @@
 """Main entry point
 """
-import datetime
 import warnings
-
-from dateutil import parser as dateparser
 import pkg_resources
-import requests
+
 import structlog
-import webob
-
-try:
-    import newrelic.agent
-except ImportError:  # pragma: no cover
-    pass
-
-try:
-    from werkzeug.contrib.profiler import ProfilerMiddleware
-except ImportError:  # pragma: no cover
-    pass
-
-
 from cornice import Service
-from pyramid.events import NewRequest, NewResponse
-from pyramid.httpexceptions import HTTPTemporaryRedirect, HTTPGone
-from pyramid.renderers import JSON as JSONRenderer
-from pyramid.security import NO_PERMISSION_REQUIRED
-from pyramid.interfaces import IAuthenticationPolicy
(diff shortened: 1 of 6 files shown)


cliquet/__init__.py
```diff
@@ -122,272 +101,31 @@ def load_default_settings(config):
...
-
-    id_generator = config.maybe_dotted(settings['cliquet.id_generator'])
-    config.registry.id_generator = id_generator()
-
-    set_auth(config)
-    attach_http_objects(config)
+    # Monkey Patch Cornice Service to setup the global CORS configuration.
+    # XXX: Refactor @crud decorator and inherit Service instead.
+    cors_origins = settings['cliquet.cors_origins']
+    Service.cors_origins = (cors_origins,)
```
leplatrem: should probably be ``aslist(cors_origins)``

ametaireau: Well, having something imutable here is something I think we should keep. `tuple(aslist(cors_origins))` then :)

cliquet/__init__.py
```diff
@@ -122,272 +101,31 @@ def load_default_settings(config):
...
+    Service.default_cors_headers = ('Backoff', 'Retry-After', 'Alert')
 
-    # Handle StatsD
-    handle_statsd(config)
+    # Setup components.
+    for step in aslist(settings['cliquet.initialization_steps']):
+        setup_func = config.maybe_dotted(step)
```
ametaireau: setup doesn't seem right here, I would use `step_location` and `step`, or maybe `step_callable`.

ametaireau:
Not documenting is IMO an issue here. We obviously don't want to have this show up in the docs as something everyone needs to know, but that's a pretty powerful thing we should expose.

Also, this approach doesn't allow to easily manipulate the steps, like we would do in python for instance (add, remove). I'm not sure what we should do about that (add more settings with a `remove` or `add` prefix?) but if we consider explicit is better than implicit then that should work out.

Also, once we introduce this, we will probably have problems when we want to rename / move our initializers somewhere else, because users will be relying on their location.

PR #256, opened 2015-05-08
Mention CORS response headers in documentation (fixes #252)

### LINKED ISSUE #252: Consider adding CORS headers
CORS headers are fun! :)

GoogleDrive, Dropbox, remoteStorage, and Hoodie all expose CORS headers on their APIs. They allow users to choose their favorite server, under their control, and http requests will still work from client-side web apps without the need to give those apps elevated permissions.

What are your thoughts about this?

leplatrem:


ametaireau:
Which makes me think we should definitely add the "httpie" calls there in order to be able to regenerate these calls, otherwise we'll get out of sync soon :)

ametaireau:
LGTM.

PR #258, opened 2015-05-08
Remove Backoff header if none in settings (fixes #238)

### LINKED ISSUE #238: Undesired "Backoff: None" in headers
It should at least be empty, or even removed from headers if setting is None

leplatrem:


ametaireau:
Cool. Thanks.

PR #259, opened 2015-05-12
Fix PUT preconditions after DELETE (fixes #249)

### LINKED ISSUE #249: Clarify PUT-after-DELETE behavior
What happens if I create a resource, then delete it, and then send a PUT request to the id of the deleted record, with a `If-Unmodified-Since` header of before the deletion time?

Will the server keep a 'tombstone' for each resource that ever existed, documented when it was deleted for the last time? If the server runs for many years, will this history of tombstones ever be truncated?

It might be worth documenting this in http://cliquet.readthedocs.org/en/latest/api/resource.html#put-collection-id or elsewhere (couldn't find anything about this edge-case).

leplatrem:


cliquet/resource.py
```diff
@@ -257,9 +257,18 @@ def put(self):
...
         except HTTPNotFound:
             existing = None
+            # Look if this record used to exist (for preconditions check).
+            deleted = Filter(self.id_field, self.record_id, COMPARISON.EQ)
+            result, _ = self.db.get_all(filters=[deleted],
+                                        include_deleted=True,
+                                        **self.db_kwargs)
+            if len(result) > 0:
+                existing = result[0]
+
```
Natim: ``finally:`` ?

PR #261, opened 2015-05-13
Register heartbeats from initialization

leplatrem:
This will allow external packages like ``cliquet-fxa`` to register a ``ping()`` method


Natim:
r+

PR #262, opened 2015-05-13
Remove unicode content from CONTRIBUTORS file.

leplatrem:
Tests fail with python3 because of this : 
https://travis-ci.org/mozilla-services/cliquet/jobs/62362839

In setup.py it should be read like this in python3:

    f = open('CONTRIBUTORS.txt', encoding='utf-8')

and in python 2:

    f = codecs.open('CONTRIBUTORS.txt', encoding='utf-8')

...so instead of introducing compat code in setup, just
remove utf8 from CONTRIBUTORS file.

@tarekziade r?

### DIFF
--- a/CONTRIBUTORS.rst
+++ b/CONTRIBUTORS.rst
@@ -2,8 +2,9 @@ Contributors
 ============
 
 * Alexis Metaireau <alexis@mozilla.com>
+* Michiel de Jong <michiel@unhosted.org>
 * hiromipaw <silvia@nopressure.co.uk>
 * Mathieu Leplatre <mathieu@mozilla.com>
 * Nicolas Perriault <nperriault@mozilla.com>
-* Remy Hubscher <rhubscher@mozilla.com>
+* Rémy Hubscher <rhubscher@mozilla.com>
 * Tarek Ziade <tarek@mozilla.com>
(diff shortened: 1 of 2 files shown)


Natim:
r-

ametaireau:
This doc is missing @michielbdejong as well! (Michiel de Jong)

I would update the setup.py to use unicode rather than updating this doc to remove them.

leplatrem:
Please do it in this branch :)

I wonder why this didn't popup earlier !

Natim:
The problem will come back as soon as we have a Russian or Chinese contributor.

leplatrem:
We can also refuse to merge pull-requests from authors that have non ascii names !

CONTRIBUTORS.rst
```diff
@@ -2,6 +2,7 @@ Contributors
 ============
 
 * Alexis Metaireau <alexis@mozilla.com>
+* Michiel de Jong <michiel@unhosted.org>
```
ametaireau: I believe it was sorted. Gah, doesn't matter.

Natim: I was sorting them by the lastname didn't see that it changed.

Natim: Fixed with https://github.com/mozilla-services/cliquet/commit/54b34337e83806ecb8b4448b66642d6518dcb1e0

michielbdejong:
Thanks for the mention! :)

PR #264, opened 2015-05-13
Change load default settings to enable generic usage

leplatrem:
r?


Natim:
Can we add an example with an environment variable in the doc?

leplatrem:
There is this already : http://cliquet.readthedocs.org/en/latest/configuration.html#environment-variables

Should we link ?

Natim:
Yes maybe it worth it

PR #265, opened 2015-05-14
Manage authn/authz/groupfinder from settings (fixes #40)

### LINKED ISSUE #40: Manage authentication from settings

leplatrem:


leplatrem:
Tests fail because it requires last master of pyramid_multiauth with this https://github.com/mozilla-services/pyramid_multiauth/pull/6

Natim:
We should change "fxa_uid" to "fxa:uid" same for other policies.

cliquet/__init__.py
```diff
@@ -76,6 +75,13 @@
     'fxa-oauth.scope': 'profile',
     'fxa-oauth.state.ttl_seconds': 3600,  # 1 hour
     'fxa-oauth.webapp.authorized_domains': '',
+    'multiauth.policies': 'fxa',
+    'multiauth.policy.fxa.use': ('cliquet.authentication.'
```
ametaireau: I like the ".use".

cliquet/initialization.py
```diff
@@ -73,25 +73,22 @@ def _redirect_to_version_view(request):
...
+    # Track policy for logging.
+    def on_policy_selected(event):
+        value = event.policy.__class__.__name__
+        value = value.replace('Authentication', '').replace('Policy', '')
+        event.request.authn_type = value
 
-    config.set_authorization_policy(authz_policy)
-    config.set_authentication_policy(authn_policy)
-    config.set_default_permission('readwrite')
+    config.add_subscriber(on_policy_selected, MultiAuthPolicySelected)
```
ametaireau: Nice, I like this way of handling things :)

docs/quickstart.rst
```diff
@@ -98,6 +99,23 @@ But most likely, it will consist in **defining REST resources** using *Cliquet*
 python API !
 
 
+Authentication
+--------------
+
+Currently, if no authentication is set in settings, *Cliquet* relies on
+:term:`Firefox Accounts`, whose setup is tedious.
```
ametaireau: I thought this should be "which" but [apparently whose is the right word to use here](http://www.quickanddirtytips.com/education/grammar/whose-for-inanimate-objects?page=all)

docs/reference/configuration.rst
```diff
@@ -222,56 +222,60 @@ in configuration:
     # cliquet.userid_hmac_secret = b4c96a8692291d88fe5a97dd91846eb4
 
 
-Basic Auth
-::::::::::
+Authentication setup
+::::::::::::::::::::
+
+*Cliquet* relies on :github:`pyramid multiauth <mozilla-service/pyramid_multiauth>`_
```
ametaireau: didn't know about this github rule in sphinx!

PR #271, opened 2015-05-20
Fixed a few oversights in the Quickstart doc.

n1k0:
r=? @ametaireau @leplatrem @Natim

### DIFF
--- a/docs/quickstart.rst
+++ b/docs/quickstart.rst
@@ -57,13 +57,18 @@ deprecation indicators are now available.
     **Shortcut!**
 
     In order to bypass the installation and configuration of *Redis* required by the
-    default storage and cache, use the «in-memory» backend in :file:`myproject.ini`:
+    default storage and cache, use the «in-memory» backend in :file:`development.ini`:
 
     .. code-block:: ini
 
-        # myproject.ini
+        # development.ini
         cliquet.cache_backend = cliquet.cache.memory
         cliquet.storage_backend = cliquet.storage.memory
+
+
+Now is a good time to install the project::
+
+    $ python setup.py develop
 
 
 Run!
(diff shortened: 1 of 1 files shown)


docs/quickstart.rst
```diff
@@ -57,15 +57,20 @@ deprecation indicators are now available.
...
 
-        # myproject.ini
+        # development.ini
         cliquet.cache_backend = cliquet.cache.memory
         cliquet.storage_backend = cliquet.storage.memory
 
 
+Now is a good time to install the project::
+
+    $ python setup.py develop
```
leplatrem: Huh ? Strange you needed this ? Where is setup.py coming from ?

Natim: I would rather use ``pip install -e .`` instead.

Natim: @leplatrem The ``setup.py`` is coming from the ``pcreate`` call

leplatrem: Ah ok, pcreate scaffold the whole egg. Good catch.

leplatrem:
r+

docs/quickstart.rst
ametaireau: +1 for using pip everywhere rather than directly the setup.py call.

For completeness, [apparently](https://pip.pypa.io/en/stable/reference/pip_install.html#editable-installs) the difference between `pip install -e` and `python setup.py develop` is really tiny, and it will act exactly the same in this case: 

> The "SomeProject.egg-info" directory is created relative to the project path. This is one advantage over just using setup.py develop, which creates the "egg-info" directly relative the current working directory.

ametaireau: I've also had to install the project in order to make it work, with pcreate and cornice templates, it seems weird when you come from django for instance, but is required in pyramid.

This makes me think we could make a cliquet template for pcreate, [as we did for cornice](https://github.com/mozilla-services/cornice/tree/master/cornice/scaffolds) which doesn't require to copy/paste any code.

ametaireau:
Fixed and merged, thanks!

PR #275, opened 2015-05-21
Switch to cache/concurrency control to Etag (fixes #251)

### LINKED ISSUE #251: Switch from Last-Modified to ETag headers?
I think your `Last-Modified` headers are violating [RFC7232](https://tools.ietf.org/html/rfc7232#section-2.2). According to this RFC, you need to use that long human-readable date format, chosen in a time when 1000 milliseconds was still a reasonable versioning granularity. :)

This means compliant http clients would consider your `Last-Modified` headers to be malformed, possibly leading to unpredictable behavior.

It is also common for a browser or other http client (or any proxy or CDN along the way) to insert its own `If-Modified-Since` header whenever it has a version of the document in cache, and there is no telling how a server implementation would react to that.

Also, violating RFCs is naughty. ;)

GoogleDrive, Dropbox, ownCloud-flavoured WebDAV, CouchDB, and remoteStorage all use ETag instead of Last-Modified. We used to have Last-Modified millisecond-timestamps in the remoteStorage protocol as well, but switched to ETags, because they do effectively the same thing, but are treated as opaque strings by http clients.

To make requests fail if a pre-existing resource changed:
* `If-Unmodified-Since: 1234567890123` -> `If-Match: "1234567890123"`

To make requests fail if a pre-existing resource did not change:
* `If-Modified-Since: 1234567890123` -> `If-None-Match: "1234567890123"`

To make requests fail if a presumed-absent resource exists:
* `If-Modified-Since: 0` -> `If-None-Match: *`

In http responses:
* `Last-Modified: 1234567890123` -> `ETag: "1234567890123"`

The downsides:
* Even when retrieving a resource for the first time, `If-Modified-Since` headers can be used to only retrieve a resource if it is recent, and `If-Unmodified-Since` to only retrieve it if it has stabilized. When using ETags, a client would have to do a filtered query on the collection first, and then decide whether to retrieve the resource or not. So you may have an extra roundtrip there, but I think this is not a very common case.
*  If you know the `Last-Modified` timestamp of a collection, then you could in theory forget the modification times of individual documents, and save some representation space. With ETags, the client needs to remember the ETag of each resource as metadata. But I think most client implementations would already do that anyway.

leplatrem:
* [x] Replaced IMS/IUMS by If-Match/If-None-Match
* [x] Added ``ETag`` CORS exposed headers 
* [x] Updated resource endpoint docs
* [x] Completed timestamps section docs
* [x] Made sure the ETag returned with 412 is the one of record, not whole collection

@Natim @ametaireau  r? 

Questions:

* Should we still support IMS and IUMS headers using HTTP date even if their equivalent timestamp value is rounded to second ?
   * *I don't think so because there would be two ways to achieve the same goal*
* Did I manage ``If-None-Match: *`` properly ?

**note:**
Tests fails because of random CloudStorage backend ``BackendError: ConnectionError: ('Connection aborted.', error(104, 'Connection reset by peer'))``

### DIFF
--- a/cliquet/resource.py
+++ b/cliquet/resource.py
@@ -712,9 +712,7 @@ class BaseResource(object):
             assert if_none_match[0] == if_none_match[-1] == '"'
             modified_since = int(if_none_match[1:-1])
         except (IndexError, AssertionError, ValueError):
-            if if_none_match == '*':
-                modified_since = -1  # Always true.
-            else:
+            if if_none_match != '*':
                 error_details = {
                     'location': 'headers',
                     'description': "Invalid value for If-None-Match"
@@ -739,25 +737,28 @@ class BaseResource(object):
             :exc:`~pyramid:pyramid.httpexceptions.HTTPPreconditionFailed`
         """
         if_match = self.request.headers.get('If-Match')
+        if_none_match = self.request.headers.get('If-None-Match')
 
-        if not if_match:
+        if not if_match and not if_none_match:
             return
 
-        try:
-            assert if_match[0] == if_match[-1] == '"'
-            modified_since = int(if_match[1:-1])
-        except (IndexError, AssertionError, ValueError):
-            error_details = {
-                'location': 'headers',
-                'description': "Invalid value for If-Match"
-            }
-            raise_invalid(self.request, **error_details)
+        if if_none_match == '*':
+            modified_since = -1  # Always raise.
+        elif if_match:
+            try:
+                assert if_match[0] == if_match[-1] == '"'
+                modified_since = int(if_match[1:-1])
+            except (IndexError, AssertionError, ValueError):
+                error_details = {
+                    'location': 'headers',
+                    'description': "Invalid value for If-Match"
+                }
+                raise_invalid(self.request, **error_details)
 
         if record:
-            current_timestamp = record[self.modified_field]
+            current_timestamp = record[self.collection.modified_field]
         else:
-            current_timestamp = self.storage.collection_timestamp(
-                **self.storage_kw)
+            current_timestamp = self.collection.timestamp()
 
         if current_timestamp > modified_since:
             error_msg = 'Resource was modified meanwhile'
--- a/cliquet/resource.py
+++ b/cliquet/resource.py
@@ -681,8 +681,7 @@ class BaseResource(object):
         # Pyramid takes care of converting.
         response.last_modified = timestamp / 1000.0
         # Return timestamp as ETag.
-        timestamp = six.text_type(timestamp).encode('utf-8')
-        response.headers['ETag'] = '"%s"' % timestamp
+        response.headers['ETag'] = ('"%s"' % timestamp).encode('utf-8')
 
     def _raise_400_if_invalid_id(self, record_id):
         """Raise 400 if specified record id does not match the format excepted
@@ -707,6 +706,8 @@ class BaseResource(object):
 
         if not if_none_match:
             return
+
+        if_none_match = if_none_match.decode('utf-8')
 
         try:
             assert if_none_match[0] == if_none_match[-1] == '"'
@@ -742,7 +743,9 @@ class BaseResource(object):
         if not if_match and not if_none_match:
             return
 
-        if if_none_match == '*':
+        if_match = if_match.decode('utf-8') if if_match else None
+
+        if if_none_match and if_none_match.decode('utf-8') == '*':
             modified_since = -1  # Always raise.
         elif if_match:
             try:
--- a/cliquet/resource.py
+++ b/cliquet/resource.py
@@ -50,12 +50,12 @@ class ViewSet(object):
 
     default_collection_arguments = {}
     collection_get_arguments = {
-        'cors_headers': ('Next-Page', 'Total-Records', 'Last-Modified')
+        'cors_headers': ('Next-Page', 'Total-Records', 'Last-Modified', 'ETag')
     }
 
     default_record_arguments = {}
     record_get_arguments = {
-        'cors_headers': ('Last-Modified',)
+        'cors_headers': ('Last-Modified', 'ETag')
     }
 
     def __init__(self, **kwargs):
@@ -672,12 +672,17 @@ class BaseResource(object):
                                   errno=ERRORS.INVALID_RESOURCE_ID)
             raise response
 
-    def _add_timestamp_header(self, response):
+    def _add_timestamp_header(self, response, timestamp=None):
         """Add current timestamp in response headers, when request comes in.
 
         """
-        timestamp = six.text_type(self.timestamp).encode('utf-8')
-        response.headers['Last-Modified'] = timestamp
+        if timestamp is None:
+            timestamp = self.timestamp
+        # Pyramid takes care of converting.
+        response.last_modified = timestamp / 1000.0
+        # Return timestamp as ETag.
+        timestamp = six.text_type(timestamp).encode('utf-8')
+        response.headers['ETag'] = '"%s"' % timestamp
 
     def _raise_400_if_invalid_id(self, record_id):
         """Raise 400 if specified record id does not match the format excepted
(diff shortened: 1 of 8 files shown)


ametaireau:
I think it would be valuable to have @michielbdejong's feedback on this!

ametaireau:
> Should we still support IMS and IUMS headers using HTTP date even if their equivalent timestamp value is rounded to second ? 

I don't think so either. If we have a way to do one thing, let's keep it, more option means more code to maintain and more ways to get things wrong.

cliquet/resource.py
```diff
@@ -623,21 +628,34 @@ def _raise_304_if_not_modified(self, record=None):
...
-                current_timestamp = self.storage.collection_timestamp(
-                    **self.storage_kw)
+        if_none_match = if_none_match.decode('utf-8')
 
-            if current_timestamp <= modified_since:
-                response = HTTPNotModified()
-                self._add_timestamp_header(response)
-                raise response
+        try:
+            assert if_none_match[0] == if_none_match[-1] == '"'
```
ametaireau: Nice use of assert!

cliquet/resource.py
```diff
@@ -623,21 +628,34 @@ def _raise_304_if_not_modified(self, record=None):
...
+            modified_since = int(if_none_match[1:-1])
+        except (IndexError, AssertionError, ValueError):
+            if if_none_match != '*':
+                error_details = {
+                    'location': 'headers',
+                    'description': "Invalid value for If-None-Match"
+                }
+                raise_invalid(self.request, **error_details)
```
ametaireau: From the spec:
> if "*" is given and any current entity exists for that resource, then the server MUST NOT perform the requested method

I think we fail to do so here?

cliquet/resource.py
```diff
@@ -646,24 +664,40 @@ def _raise_412_if_modified(self, record=None):
...
-            if current_timestamp > unmodified_since:
-                error_msg = 'Resource was modified meanwhile'
-                response = http_error(HTTPPreconditionFailed(),
-                                      errno=ERRORS.MODIFIED_MEANWHILE,
-                                      message=error_msg)
-                self._add_timestamp_header(response)
-                raise response
+        if_match = if_match.decode('utf-8') if if_match else None
+
+        if if_none_match and if_none_match.decode('utf-8') == '*':
```
ametaireau: And, I withdraw y previous remark!

docs/api/timestamps.rst
```diff
@@ -4,20 +4,56 @@
...
-In order to avoid race conditions, *Cliquet* guarantees that each change will
-increment the timestamp of the related collection.
-If two changes happen at the same millisecond, they will still have two differents
-timestamps.
+::
+
+    Last-Modified: Wed May 20 17:22:38 2015 +0200
+
+
+.. versionchanged:: 2.0
```
ametaireau: I guess today is the day I learn about all these sphinx directives!

ametaireau:
Looks good to me.

michielbdejong:
Thanks for the invite to comment! :)

Quick answer re the choice of whether or not to support IMS/IUMS: afaik that is not up to you as spec authors, because your spec builds on http (which defines lots of things, including these headers).

So it should be left to the server implementer to either obey such headers, or always just execute the action / reply with the body anyway. This means clients can't rely on it, so in practice it will not be used a lot. But explicitly forbidding people to implement it would be subsetting http, which is bad practice afaik.

So you can choose not to talk about it, or mention it as a 'MAY'. (haven't looked yet if this is what you did).

I'll have a look!

docs/api/resource.rst
```diff
@@ -192,9 +194,9 @@ HTTP Status Codes
 -----------------
 
 * ``200 OK``: The request was processed
-* ``304 Not Modified``: Collection did not change since value in ``If-Modified-Since`` header
+* ``304 Not Modified``: Collection did not change since value in ``If-None-Match`` header
 * ``400 Bad Request``: The request querystring is invalid
-* ``412 Precondition Failed``: Collection changed since value in ``If-Unmodified-Since`` header
```
michielbdejong: Off the top of my head, a client could send a PUT with `If-None-Match: *` to make sure that only creates a new document if none exists, but never overwrites an existing one, would also fail with 412 and not 304. So I think 304s are for read actions and 412s are for write actions. Would have to look it up to be sure, though.

docs/api/resource.rst
```diff
@@ -351,10 +353,10 @@ GET /{collection}/<id>
 
 Returns a specific record by its id.
 
-For convenience and consistency, a header ``Last-Modified`` will also repeat the
+For convenience and consistency, a header ``ETag`` will also repeat the
 value of the ``last_modified`` field.
```
michielbdejong: Is there a reason to force server implementers to use milli-timestamp ETags? What would happen if someone implements a server that uses SHA hashes for ETags? This could easily happen if someone implements a server on top of some middleware library (like nodejs express or something), or configures a webserver like nginx to handle the whole ETag and 412/304 story.

docs/api/resource.rst
```diff
@@ -431,7 +433,7 @@ mapping validating the resource schema fields.
 
 Validation and conflicts behaviour is similar to creating records (``POST``).
 
-If the request header ``If-Unmodified-Since`` is provided, and if the record has
+If the request header ``If-Match`` is provided, and if the record has
 changed meanwhile, a ``412 Precondition failed`` error is returned.
```
michielbdejong: Maybe mention `If-None-Match: *` here, for 'PUT this here, but only if there is nothing there yet'

docs/api/resource.rst
```diff
@@ -476,7 +478,12 @@ HTTP Status Code
 * ``400 Bad Request``: The record is invalid
 * ``409 Conflict``: If replacing this record violates a field unicity constraint
 * ``412 Precondition Failed``: Record was changed or deleted since value
-  in ``If-Unmodified-Since`` header
+  in ``If-Match`` header.
+
+.. note::
+
+    A ``If-None-Match: *`` request header can be used to make sure the ``PUT``
+    won't overwrite any record.
```
michielbdejong: [re previous comment:] Ah ok, you mention that in the note here, sorry.

docs/api/timestamps.rst
```diff
@@ -4,20 +4,56 @@
...
+===================
+
+In order to prevent race conditions, like overwriting changes occured in the interim for example,
+a ``If-Match`` request header can be used. If the response is ``412 Precondition failed``
+then the resource has changed meanwhile.
+
+The client can then choose to:
+
+* overwrite by repeating the request without ``If-Match``;
+* reconcile the resource by fetching, merging and repeating the request.
```
michielbdejong: Nice!

docs/api/resource.rst
leplatrem: So far the philosophy of this toolkit has been to provide an implementation of a particular protocol. But you are right, ETags should remain «*opaque*» and shouldn't have too much semantics. 
I'll remove that sentence about «*repeating the last modified value*».

Thanks !

docs/api/resource.rst
leplatrem: Yes, that's how we did it. In brief:
* 304 on read operations when client version is «up to date»
* 412 on write operations when server version has changed meanwhile

leplatrem:
@ametaireau @Natim r?

cliquet/tests/resource/test_preconditions.py
```diff
@@ -63,6 +89,7 @@ def test_412_on_collection_has_last_modified_timestamp(self):
             self.resource.collection_get()
         except httpexceptions.HTTPPreconditionFailed as e:
             error = e
+        self.assertIsNotNone(error.headers.get('ETag'))
         self.assertIsNotNone(error.headers.get('Last-Modified'))
```
Natim: Do we keep the Last-Modified header?

Natim: Ok I've seen the doc. Never Mind.

PR #277, opened 2015-05-22
Refactor the imports - Refs #34

hiromipaw:
Imports resource directly in cliquet and change BaseResource to Resource. As per issue #34

### DIFF
--- a/cliquet/__init__.py
+++ b/cliquet/__init__.py
@@ -2,8 +2,8 @@
 """
 import warnings
 import pkg_resources
-
 import structlog
+
 from cornice import Service
 from pyramid.settings import asbool, aslist
 
@@ -13,7 +13,8 @@ logger = structlog.get_logger()
 from cliquet import utils
 from cliquet.initialization import (  # NOQA
     initialize, initialize_cliquet, install_middlewares)
-
+from cliquet import resource
+from cliquet import logger
 
 # Module version, as defined in PEP-0396.
 __version__ = pkg_resources.get_distribution(__package__).version
--- a/cliquet/errors.py
+++ b/cliquet/errors.py
@@ -1,6 +1,5 @@
 import six
 from pyramid import httpexceptions
-
 from cliquet import logger
 from cliquet.utils import Enum, json, reapply_cors
 
--- a/cliquet/resource.py
+++ b/cliquet/resource.py
@@ -12,7 +12,7 @@ from cliquet import logger
 from cliquet.storage import exceptions as storage_exceptions, Filter, Sort
 from cliquet.errors import (http_error, raise_invalid, ERRORS,
                             json_error_handler)
-from cliquet.schema import ResourceSchema
+from cliquet.schema import Schema
 from cliquet.utils import (
     COMPARISON, classname, native_value, decode64, encode64, json,
     current_service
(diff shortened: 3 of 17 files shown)


hiromipaw:
I have some dependency to fix. Sorry for being late with this.

docs/reference/resource.rst
```diff
@@ -15,12 +15,12 @@ Full example
...
     import colander
 
-    from cliquet import resource
+    import cliquet
     from cliquet import schema
     from cliquet import utils
 
 
-    class BookmarkSchema(resource.ResourceSchema):
+    class BookmarkSchema(cliquet.ResourceSchema):
```
leplatrem: I suppose we could even rename ``cliquet.ResourceSchema`` to ``cliquet.Schema`` ?

cliquet/__init__.py
```diff
@@ -2,6 +2,7 @@
 """
 import warnings
 import pkg_resources
+import resource
```
leplatrem: this bloc of imports groups the *stdlib* imports.

It should be ``from cliquet import resource`` along with others below like utils, etc.

cliquet/resource.py
```diff
@@ -8,7 +8,7 @@
                                     HTTPNotFound, HTTPConflict)
 import six
 
-from cliquet import logger
+from cliquet.logger import *
```
leplatrem: Importing ``*`` is not recommanded. Any good reason to have it?

leplatrem:
Please don't say you're late :)
We're still breaking things anyway :)

Thanks a lot for your contribs!!

docs/reference/resource.rst
ametaireau: +1

hiromipaw: :) allrighty

cliquet/resource.py
hiromipaw: Nope was doing a test to understand why I was getting a couldn't find logger kind of error... :)

hiromipaw:
Hey I have the same tests failing also on master. Is something with my local setup or am I forgetting something basic... ?

cliquet/tests/resource/test_views_cors.py
```diff
@@ -1,5 +1,5 @@
 import mock
-
+import cliquet
```
Natim: Is it necessary?

cliquet/__init__.py
```diff
@@ -2,8 +2,8 @@
 """
 import warnings
 import pkg_resources
-
```
Natim: Here we separate imports from the sdlib from imports from other libraries having ``import`` first and then ``from``

cliquet/errors.py
```diff
@@ -1,6 +1,5 @@
 import six
 from pyramid import httpexceptions
-
```
Natim: ditto

cliquet/__init__.py
```diff
@@ -13,7 +13,8 @@
 from cliquet import utils
 from cliquet.initialization import (  # NOQA
     initialize, initialize_cliquet, install_middlewares)
-
+from cliquet import resource
```
Natim: Here it is not what we want.

We should probably define __all__ to define the exaustive list of what we want to expose in cliquet itself.

See https://github.com/django/django/blob/master/django/core/files/__init__.py#L3

Natim:
refs https://github.com/nopressurelabs/cliquet/pull/1

hiromipaw:
Thanks for the comments. I'll keep working on this. :)

PR #278, opened 2015-05-25
Add decorator to cache function results

leplatrem:


leplatrem:
@Natim @ametaireau r?

Natim:
How do you invalidate a cached value?

leplatrem:
So far you can only rely on expiration. It may deserve a warning in docs you're right :)

ametaireau:
Looks good to me! Also, see [functools.lru_cache](https://docs.python.org/3/library/functools.html#functools.lru_cache) from std python.

cliquet/utils.py
```diff
@@ -116,3 +144,46 @@ def current_service(request):
...
+    from pyramid.threadlocal import get_current_registry
+
+    # On class method, skip ``self`` to compute cache key.
+    frames = inspect.stack()
+    defined_in_class = (len(frames) > 2 and
+                        frames[2][4][0].strip().startswith('class '))
+    skip_args = 1 if defined_in_class else 0
+
+    # Get cache from current thread.
+    cache = get_current_registry().cache
```
tarekziade: This MemoryBackend class seems to provide a cache that can grow indefinitely. That could be a problem if it's misused. 

We should have a maxsize option somewhere (or use Python's built in LRU cache) and prevent the memory can to grow indefinitely.

cliquet/utils.py
```diff
@@ -116,3 +144,46 @@ def current_service(request):
...
+
+    def decorator(func):
+        @functools.wraps(func)
+        def wrapped(*args, **kwargs):
+            key = prefix + func.__name__ + str(args[skip_args:]) + str(kwargs)
```
tarekziade: what happens when some arguments are not hashable ? like socket objects ? That line will raise an error

cliquet/utils.py
```diff
@@ -116,3 +144,46 @@ def current_service(request):
...
+                resp = requests.get('http://server/?allowed=' % user_id)
+                return resp.status_code == 200
+
+    """
+    from pyramid.threadlocal import get_current_registry
+
+    # On class method, skip ``self`` to compute cache key.
+    frames = inspect.stack()
+    defined_in_class = (len(frames) > 2 and
+                        frames[2][4][0].strip().startswith('class '))
```
tarekziade: this  `[2][4][0].strip().startswith('class ')` smeel too magic and not robust. You should define clearly what you're looking for here.

leplatrem: My cat went over the keyboard :D

cliquet/utils.py
leplatrem: I remember that @Natim did a great hashing function for the cache keys in some project but can't remember where...

cliquet/utils.py
leplatrem: See #406

leplatrem:
Dropping this in favor of future #558

cliquet/utils.py
leplatrem: here :)
https://github.com/mozilla/PyFxA/commit/d889a6e9dea4dd961a269ae6074cb4c14e621a42#diff-5b10591afd24b7722d6778775f450b5fR23

PR #279, opened 2015-05-25
Do limit cache value to strings

leplatrem:


leplatrem:
Tests fail because of ``BackendError: ConnectionError: ('Connection aborted.', error(104, 'Connection reset by peer'))``

@Natim r?

PR #282, opened 2015-05-28
Allow custom usage of resource

leplatrem:
* [x] Split current ``BaseResource`` into ``Collection`` and ``BaseResource``
* [x] Remove storage dependency to storage and mapping

**See before and after**:

This design has many advantages:
* Less indirections between storage and resource
* No *fake resource* to use and test the storage
* Allow custom usage of storage class

But has some drawbacks that I don't like:
* The signatures of storage methods are overloaded

Suggestions to get rid of drawbacks:
* Remove ability to customize ``id_field``, ``modified_field``, ``delete_field``

**TODO**

* [x] Make tests pass
* [x] Fix notion of Authorization in cloud storage backend
* [x] Fix wording in storage
    * ``resource_name`` ­→ ``category_id`` (?)
    * ``user_id`` → ``parent_id``
    *  ``record_id`` → ``object_id``


cliquet/resource.py
```diff
@@ -52,41 +52,253 @@ def wrapper(klass):
...
+class Collection(object):
+    """A collection stores and manipulate records in its attached storage.
 
-    validate_schema_for = ('POST', 'PUT')
-    """HTTP verbs for which the schema must be validated"""
+    It is not aware of HTTP environment nor protocol.
 
+    Records are isolated according to the provided `name` and `parent_id`.
+
+    Those notions have no particular semantic and can be represent anything.
```
ametaireau: can represent.

cliquet/resource.py
```diff
@@ -52,41 +52,253 @@ def wrapper(klass):
...
+    For example, the resource `name` can be the *type* of objects stored, and
+    `parent_id` can be the current *user id* or *a group* where the collections
+    belongs. If left empty, the resource records are global.
```
leplatrem: belong

leplatrem: the collections records are not isolated

Natim:
category_id -> collection_id

cliquet/resource.py
```diff
@@ -52,45 +52,260 @@ def wrapper(klass):
...
+        :param str parent_id: the default parent id
+        """
+        self.storage = storage
+        self.id_generator = id_generator
+        self.parent_id = parent_id
         try:
-            self.name = classname(self)
+            self.name = name
         except AttributeError:
             # Retrocompatibilty with former readonly @property.
```
Natim: We probably want to get rid of it.

cliquet/collection.py
```diff
@@ -0,0 +1,235 @@
...
+
+    deleted_field = 'deleted'
+    """Name of `deleted` field in deleted records"""
+
+    def __init__(self, storage, id_generator=None, collection_id='',
+                 parent_id='', auth=None):
+        """
+        :param storage: an instance of storage
+        :type storage: :class:`cliquet.storage.Storage`
+        :param id_generator: an instance of id generator, used by storage
```
ametaireau: if this is used by the storage, why passing it to the collection rather than to the storage?

ametaireau: okay, got it. All collections aren't using the same so it makes sense to have it here.

docs/reference/resource.rst
```diff
@@ -71,10 +71,46 @@ Resource class
...
+                                                                  unique_fields)
+            trackid = index.track(record)
+            record['trackid'] = trackid
+            return record
 
-.. automodule:: cliquet.resource
+
+    class Payment(resource.BaseResource):
+        def __init__(request):
+            super(Mushroom, self).__init__(request)
```
ametaireau: wrong classname.

PR #286, opened 2015-06-03
Permission backend alexis

ametaireau:


cliquet/permission/__init__.py
```diff
@@ -78,13 +77,13 @@ def object_permission_principals(self, object_id, permission):
         raise NotImplementedError
 
     def object_permission_authorized_principals(self, object_id, permission,
-                                                _get_perm_keys=None):
+                                                get_perm_keys=None):
```
leplatrem: Since you're tackling wording. Shouldn't this be a *bound permissions* ? *perm keys* is vague, but a wording for this is hard :)

leplatrem:
r+ if we leave get_perm_keys

PR #280, opened 2015-05-27
Add the permission backend.

Natim:
- [x] Write the base interface
- [x] Write the redis backend
- [x] Write the memory backend
- [x] Write permission backend generic unit tests
- [x] Write documentation

Issue #254, opened 2015-05-07
Replace Cliquet protocol by JSON-Api spec ?


leplatrem:
Yesterday I assisted to @steveklabnik talk about the [JSON-API spec](http://jsonapi.org/about/) at [APIdays BCN](http://mediterranea.apidays.io/).

I was amazed by the number of intentions we have in common! And, although we completely ignored its existence, [the protocol we designed for Cliquet](http://cliquet.readthedocs.org/en/latest/api/index.html) almost matches their spec! That's icing on the cake :)

Shared goals:

* save the developers the hassle of making choices when building a JSON-API;
* convention over configuration;
* simple thus loosely coupled; (*not like a couchdb protocol*)
* easy to debug;
* pagination/sort/filter;
* simple JavaScript clients (Ember data will support it natively in the next version);

Protocol differences:

* so far Cliquet doesn't manage relations between models, so we'll just ignore their work on that for a start;
* the Bulk API is different, and does not look like ours;
* we have a simple concurrency-control model, and that could become an extension (like Bulk);
* we have *conflict* responses regarding records field unicity;
* the use of a dedicated mimetype is kind of overkill, but we could also serve payloads as application/json;
* the pagination querystring param format differs a bit;
* the sort querystring param has ``+`` for ascending order;

I strongly believe that we should put some efforts in adjusting our custom protocol to conform with the JSON-API spec.

Thoughts ?

ametaireau:
Some notes while reading the document:

- I don't understand fully how this specification is different from some other ones (which tend to have a more sementic approach)
- Pagination seems handled with links, where we were using headers, and indicates when it's the "last" page (which we purposely don't do)
- It seems that the records are not returned "as is", but rather in an "attributes" object. Why not but it seems it also could add parsing overhead on the client side?
- I haven't found information about their batch endpoint.

That being said, I think it would be good to have a formalism on the APIs we're exposing, and if this is already being discussed elsewhere, then we definitely should consider it rather than doing or own.

leplatrem:
> I don't understand fully how this specification is different from some other ones (which tend to have a
> more sementic approach)

That was my question to @steveklabnik after the talk. He explained that most hypermedia initiatives tend to focus on semantics. ``JSON-LD`` tends to be like Semantic Web, ``collection+json`` tends to very good for search results, etc..

> I haven't found information about their batch endpoint.

http://jsonapi.org/extensions/bulk/

hiromipaw:
[...]

PR #287, opened 2015-06-03
Data field in request/response payload (ref #254)

leplatrem:
Based on top of #280.

Necessary to introduce ``permissions`` attribute in bodies and payloads.


leplatrem:
This should have been merged separately. But was merged via #288

Issue #102, opened 2015-03-18
Refactor structlog logger creation


Natim:
Currently we are creating the logger in cliquet/__init__.py, before importing thing that need to import the logger as well.

We should move that to another file in order to avoid to have to import it in the middle of the loading list.

See: https://github.com/mozilla-services/cliquet/blob/master/cliquet/__init__.py#L17-L18

leplatrem:
Related #173

leplatrem:
Is that really a problem?

Natim:
Maybe not :)

PR #291, opened 2015-06-05
Check permissions on incoming requests

ametaireau:
Adds permissions check to cliquet applications:
- [x] Create a context factory to provide information about the requested objects;
- [x] Update the authorization policy to look at what's contained in the context;
- [x] Change the default permission to be `dynamic`;
- [x] Add unit tests for context factory;
- [x] Add unit tests for authorization policy;
- [x] Update the functional tests;
- [x] Add functional tests without mocking authorization.

To review:
- Naming and conventions;
- Tests;
- I've updated the tests to use another fake authz backend, I think that would make everything easier to test, but that can be discussed.


cliquet/permission/__init__.py
```diff
@@ -0,0 +1,129 @@
...
+    def remove_principal_from_ace(self, object_id, permission, principal):
+        """Remove a principal to an Access Control Entry.
+
+        :param str object_id: The object to remove the permission principal to.
+        :param str permission: The permission that should be removed.
+        :param str principal: The principal to remove to the ACE.
+        """
+        raise NotImplementedError
+
+    def object_permission_principals(self, object_id, permission):
```
Natim: rename to ``get_ace_principals``

cliquet/permission/__init__.py
```diff
@@ -0,0 +1,129 @@
...
+
+        :param str object_id: The object_id the permission is set to.
+        :param str permission: The permission to query.
+        :returns: The list of user principals
+        :rtype: set
+
+        """
+        raise NotImplementedError
+
+    def object_permission_authorized_principals(self, object_id, permission,
```
Natim: Rename to ``get_ace_authorized_principals``

cliquet/__init__.py
```diff
@@ -7,13 +7,12 @@
...
 
-# Main Cliquet logger.
-logger = structlog.get_logger()
-
 from cliquet import utils
 from cliquet.initialization import (  # NOQA
     initialize, initialize_cliquet, install_middlewares)
 
+# Main Cliquet logger.
+logger = structlog.get_logger()
```
leplatrem: Is that #102 ?

cliquet/authorization.py
```diff
@@ -0,0 +1,74 @@
...
+
+
+@implementer(IAuthorizationPolicy)
+class AuthorizationPolicy(object):
+    def permits(self, context, principals, permission):
+        if permission == 'dynamic':
+            permission = context.required_permission
+        object_id = context.object_id
+        get_bound_permissions = context.get_bound_permissions
+        has_permission = context.has_permission
```
leplatrem: it's not very explicit that this is a callable.

cliquet/authorization.py
```diff
@@ -0,0 +1,74 @@
...
+
+        if permission == 'create':
+            permission = '%s:%s' % (context.object_type, permission)
+        return has_permission(object_id, permission, principals,
+                              get_bound_permissions)
```
leplatrem: Here, only ``permission`` and ``principals`` are coming from this function.
Is context supposed to be a stateless struct or could we curry ``has_permission()`` with ``object_id and get_bound_permission ?

cliquet/resource.py
```diff
@@ -198,7 +192,14 @@ def register_service(endpoint_type):
         name = viewset.get_service_name(endpoint_type, resource)
 
         service = Service(name, path, depth=depth,
-                          **viewset.service_arguments or {})
+                          **viewset.get_service_arguments())
+
+        # Attach viewset and resource to the service for later reference.
+        service.viewset = viewset
+        service.resource = resource
```
leplatrem: Oh, this is a great helper :)

Just wondering (issue for later if relevant) : shouldn't we add a reified property on request instead ? (using a closure for example)

cliquet/resource.py
```diff
@@ -239,7 +238,8 @@ class BaseResource(object):
     mapping = ResourceSchema()
     """Schema to validate records."""
 
-    def __init__(self, request):
+    def __init__(self, request, context=None):
```
leplatrem: When is context left None now that we have a root factory ? During tests only ?

cliquet/authorization.py
```diff
@@ -0,0 +1,74 @@
...
+        "get": "read",
+        "post": "create",
+        "delete": "write",
+        "patch": "write"
+    }
+
+    def __init__(self, request):
+        service = utils.current_service(request)
+        object_id = get_object_id(request)
+        self.object_type = service.viewset.get_name(service.resource)
```
leplatrem: Add note for future refactoring here: object_type vs. resource name

cliquet/authorization.py
```diff
@@ -0,0 +1,74 @@
...
+            # In the case of a "PUT", check if the associated record already
+            # exists, return "write" if it does, "create" otherwise.
+            # For now, consider whoever puts should have "create" access.
+
+            # If the view exists, call it with the request and catch an
+            # eventual NotFound.
+            if service is not None:
+                resource = service.resource(request)
+                try:
+                    resource.collection.get_record(resource.record_id)
```
leplatrem: Add a note for later: reify for later fetch in PUT view

cliquet/authorization.py
```diff
@@ -0,0 +1,74 @@
...
+                    permission = "write"
+        else:
+            permission = self.method_permissions[request.method.lower()]
+
+        self.required_permission = permission
+        self.object_id = object_id
+        self.has_permission = request.registry.permission.has_permission
+
+
+def get_object_id(request):
```
leplatrem: We could rename it to ``get_object_uid()`` to avoid confusion with ``record[id]`` ?

cliquet/authorization.py
```diff
@@ -0,0 +1,74 @@
...
+from zope.interface import implementer
+
+from cliquet import utils
+from cliquet.storage import exceptions as storage_exceptions
+
+
+@implementer(IAuthorizationPolicy)
+class AuthorizationPolicy(object):
+    def permits(self, context, principals, permission):
+        if permission == 'dynamic':
```
leplatrem: dynamic ? contextual ? relative ? deduced ? :)

cliquet/tests/resource/test_views.py
```diff
@@ -56,31 +66,117 @@ def test_all_views_require_authentication(self):
...
+        object_id = self.collection_url
+        self.app.app.registry.permission.add_principal_to_ace(
+            object_id, 'read', self.principal)
+
+        self.app.get(self.collection_url, headers=self.headers, status=200)
+
+    def test_collection_post_is_granted_when_authorized(self):
+        object_id = self.collection_url
+        self.app.app.registry.permission.add_principal_to_ace(
+            object_id, 'mushroom:create', self.principal)
```
leplatrem: Shouldn't it be ``/mushroom:create`` ?

cliquet/tests/resource/test_views.py
```diff
@@ -56,31 +66,117 @@ def test_all_views_require_authentication(self):
...
+        self.app.get(self.record_url, headers=self.headers, status=200)
+
+    def test_record_patch_is_granted_when_authorized(self):
+        object_id = self.record_url
+        self.app.app.registry.permission.add_principal_to_ace(
+            object_id, 'write', self.principal)
+
+        self.app.patch_json(self.record_url, MINIMALIST_RECORD,
+                            headers=self.headers, status=200)
+
```
leplatrem: Could be split into two tests cases : Denied / Granted

cliquet/tests/resource/test_views.py
```diff
@@ -56,31 +66,117 @@ def test_all_views_require_authentication(self):
...
+        self.app.put_json(self.record_url, MINIMALIST_RECORD,
+                          headers=self.headers, status=200)
 
-        url = self.item_url.format(id=resp.json['id'])
-        self.app.get(url)
-        self.assertEqual(permission_required(), 'readonly')
+    def test_record_put_on_unexisting_record_is_granted_when_authorized(self):
+        object_id = self.collection_url
+        self.app.app.registry.permission.add_principal_to_ace(
+                object_id, 'mushroom:create', self.principal)
```
leplatrem: this is repeated in a every method, could thus be in ``setUp()``

cliquet/tests/support.py
```diff
@@ -105,4 +110,30 @@ def _create_thread(self, *args, **kwargs):
         return thread
 
 
+@implementer(IAuthorizationPolicy)
+class AllowAuthorizationPolicy(object):
+    def permits(self, context, principals, permission):
+        if ('basicauth_9f2d363f98418b13253d6d7193fc88690302'
+                'ab0ae21295521f6029dffe9dc3b0') in principals:
```
leplatrem: Could be a constant :)

cliquet/initialization.py
```diff
@@ -76,7 +76,8 @@ def setup_authentication(config):
     from configuration.
     """
     config.include('pyramid_multiauth')
-    config.set_default_permission('readwrite')
+    # By default, permissions are handled dynamically.
+    config.set_default_permission('dynamic')
```
leplatrem: dynamic should be a constant like those http://docs.pylonsproject.org/projects/pyramid//en/latest/api/security.html#constants

cliquet/tests/resource/test_views.py
Natim: No because it is the resource name not the ``parent_id``

cliquet/tests/resource/test_views.py
Natim: Isn't it already the case? https://github.com/mozilla-services/cliquet/pull/291/files#diff-45fb1d7796e256e845674444b726cf81R122

cliquet/tests/resource/test_views.py
Natim: The permission name change each time and it is what makes the test understandable.

cliquet/__init__.py
ametaireau: That's just to make flake8 happy (otherwise imports are after an instruction)

cliquet/authorization.py
ametaireau: It's dynamic because it's dynamically computed at the request time.

cliquet/authorization.py
ametaireau: permits?

cliquet/authorization.py
ametaireau: That's a good remark indeed!

cliquet/authorization.py
ametaireau: Actually we should just attach it to the context! The context is now passed to the resource at initialisation time.

cliquet/authorization.py
ametaireau: Everywhere in the code we're talking about object_id. I'm not opposed to change the terms, but I don't think uid makes more sense here than object_id.

cliquet/resource.py
ametaireau: Here, it's not possible to add [a reified property](http://docs.pylonsproject.org/projects/pyramid/en/latest/api/decorator.html.) to the request just because … we don't have access to it just here! Do you have something else in mind?

cliquet/resource.py
ametaireau: I haven't explained it in more details here, but in case a context factory was used, the context will be populated. That means we can have a look at anything that's in the context.

In our case, to avoid double lookups for the data in the database (when we had to check the database for information, for instance in case of a PUT request) it might make sense to store the record in the context and then check the context rather than querying the database if possible.

I let this as an optimisation, for later.

ametaireau: Hmm, and I misread this it seems! You probably are right and we don't need to have this `context=None` here, I can just remove the default argument. /me tries.

ametaireau: Yep, tests fail without this. I could update them but I kinda like the fact that the context is optional.

cliquet/tests/resource/test_views.py
ametaireau: Done.

PR #294, opened 2015-06-09
Preserve record id on POST (fixes #293)

### LINKED ISSUE #293: Preserve provided id in record for POST on collection
As spotted with @n1k0 yesterday :)

leplatrem:


CHANGELOG.rst
```diff
@@ -6,6 +6,11 @@ This document describes changes between each past release.
 1.8.1 (unreleased)
 ------------------
 
+**Bug fixes**
```
Natim: It is not a bug it is a feature and @ametaireau was not ready to implement it that fast also it is a JSONAPI feature.

Natim:
Related to http://jsonapi.org/format/#crud-creating-client-ids

CHANGELOG.rst
Natim: See test here: https://github.com/mozilla-services/cliquet/pull/294/files#diff-45fb1d7796e256e845674444b726cf81L144

leplatrem: What do you mean ?

@ametaireau would be against this change ?

No problem, we can still discuss it, and reject it even if there's code attached...

Natim: No I just meant that In the changelog it should not be in the Bug category.

I think it is a feature we want (afaik I want it and @n1k0 want it too)

ametaireau:
I actually find it weird to have a POST specify the id of the record, where IMO it is the server's job. I would like to know more about this, and what are the reasons we want to have this.

ametaireau:
Hmm okay; @Natim bring some light here, neat. I haven't realized this is already possible with a PUT anyway, and that PUT is missing one interesting property: it erases the record if it already exists, while doing this with a POST allows the creation to fail in case the ID is already taken.

leplatrem:
It is possible to prevent PUT to replace records using ``If-Unmodified-Since: 0``

And what you describe about collision for POST is not tested at all. I guess that it will trigger a 50X :) storage backend integrity error.

I'll work something out, thanks !

leplatrem:
Before forced push, here is the diff between my rebase and this:

```diff
diff --git a/CHANGELOG.rst b/CHANGELOG.rst
index 3a888e3..810e7ce 100644
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -6,7 +6,7 @@ This document describes changes between each past release.
 1.8.1 (unreleased)
 ------------------
 
-**Bug fixes**
+**New features**
 
 - Preserve provided ``id`` field of records using POST on collection (fixes #293)
 
diff --git a/cliquet/resource.py b/cliquet/resource.py
index 274ade4..225cccc 100644
--- a/cliquet/resource.py
+++ b/cliquet/resource.py
@@ -363,9 +363,8 @@ class BaseResource(object):
         # Since ``id`` does not belong to schema, it can only be found in body.
         try:
             id_field = self.collection.id_field
-            provided_id = self.request.json['data'][id_field]
-            new_record[id_field] = provided_id
-            self._raise_400_if_invalid_id(provided_id)
+            new_record[id_field] = _id = self.request.json['data'][id_field]
+            self._raise_400_if_invalid_id(_id)
         except KeyError:
             pass
 
diff --git a/cliquet/tests/resource/test_views.py b/cliquet/tests/resource/test_views.py
index bb3fe56..bfbe7d9 100644
--- a/cliquet/tests/resource/test_views.py
+++ b/cliquet/tests/resource/test_views.py
@@ -272,9 +272,9 @@ class InvalidRecordTest(BaseWebTest):
         record = MINIMALIST_RECORD.copy()
         record_id = record['id'] = '472be9ec-26fe-461b-8282-9c4e4b207ab3'
         resp = self.app.post_json(self.collection_url,
-                                  {'data': record},
+                                  record,
                                   headers=self.headers)
-        self.assertEqual(resp.json['data']['id'], record_id)
+        self.assertEqual(resp.json['id'], record_id)
 
 

```

Natim:
It seems that something is missing for at least the last part of the diff

leplatrem:
@Natim r?

PR #300, opened 2015-06-10
Return 201 for PUT on creation (fixes #298)

### LINKED ISSUE #298: With PUT, response status should be 201 if object was created

leplatrem:


Natim:
Great thanks.

PR #301, opened 2015-06-11
Handle blank requests that doesn't return a service.

Natim:


cliquet/authorization.py
```diff
@@ -39,6 +39,11 @@ class RouteFactory(object):
 
     def __init__(self, request):
         service = utils.current_service(request)
+
+        # Handle blank requests
+        if service is None:
+            return
```
leplatrem: Is there any failing test for this fix ?

Natim: What makes it fails is when you create the configuration passing a root_factory in order use the RouteFactory everywhere.

Natim: We can probably create an app with a configurator like this to cover this if with tests.

PR #302, opened 2015-06-11
Permissions backend postgresql

leplatrem:
@Natim r?


cliquet/permission/memory.py
```diff
@@ -76,10 +76,10 @@ def object_permission_authorized_principals(self, object_id, permission,
             keys = [(object_id, permission)]
         else:
             keys = get_bound_permissions(object_id, permission)
-        permissions = set([])
+        principals = set([])
         for obj_id, perm in keys:
-            permissions |= self.object_permission_principals(obj_id, perm)
-        return permissions
+            principals |= self.object_permission_principals(obj_id, perm)
+        return principals
```
Natim: good catch

cliquet/permission/postgresql/__init__.py
```diff
@@ -0,0 +1,178 @@
...
+        here = os.path.abspath(os.path.dirname(__file__))
+        schema = open(os.path.join(here, 'schema.sql')).read()
+        with self.connect() as cursor:
+            cursor.execute(schema)
+        logger.info('Created PostgreSQL permission tables')
+
+    def flush(self):
+        query = """
+        DELETE FROM user_principals;
+        DELETE FROM aces;
```
Natim: Isn't ``access_control_entries`` more explicit?

cliquet/permission/postgresql/__init__.py
```diff
@@ -0,0 +1,178 @@
...
+        )
+        SELECT principal
+          FROM required_perms JOIN aces
+            ON (object_id = column1 AND permission = column2)
+        ;""" % perms
+        with self.connect() as cursor:
+            cursor.execute(query)
+            results = cursor.fetchall()
+        return set([r['principal'] for r in results])
+
```
Natim: We should had the check_permission method on postgresql ends as well here.

leplatrem: postgresql ends ?

cliquet/permission/postgresql/__init__.py
leplatrem: oh yes!

cliquet/permission/postgresql/__init__.py
Natim: Build a SQL query that return ``True`` or ``False`` for the ``check_permission``

leplatrem: Ah, ok. I just read the code of ``BasePermission.check_permission``.

But then why do we need ``object_permission_authorized_principals`` ?

Natim: The simple answer is "for debugging"

leplatrem: Then we can get rid of it :)

PR #303, opened 2015-06-11
Fix service.collection_path

Natim:


cliquet/authorization.py
```diff
@@ -56,7 +56,8 @@ def __init__(self, request):
                 try:
                     resource.collection.get_record(resource.record_id)
                 except storage_exceptions.RecordNotFoundError:
-                    object_id = service.collection_path
+                    object_id = service.collection_path.format(
+                        **request.matchdict)
```
leplatrem: Is there any failing test for this fix ?

Natim:
It is just because the collection_path as we put it was the template, now it is the real collection_path,

cliquet/tests/test_authorization.py
```diff
@@ -73,6 +73,24 @@ def test_attributes_are_none_with_non_resource_requests(self):
         self.assertIsNone(context.resource_name)
         self.assertIsNone(context.check_permission)
 
+    def test_service_collection_path_is_correct(self):
```
leplatrem: "is correct" is not explicit (https://plus.google.com/+JonathanLange/posts/YA3ThKWhSAj)

you could say : on creation, permission is bound to create on collection id

Natim: It is not the test, what you describe is: https://github.com/mozilla-services/cliquet/pull/303/files#diff-d6ddebfba0cf4aea5bdbcb72d77b3610R47

Here it is more, ``test_service_collection_path_resolve_to_current_collection_path``

PR #304, opened 2015-06-11
Use authorization.get_object_id() in ProtectedResource

leplatrem:


ametaireau:
I like it! r+

Natim:
Apparently the rebase is not that fast-forward.

PR #320, opened 2015-06-16
Remove custom code to disable collection delete (fixes #46)

### LINKED ISSUE #46: Refactor disabling of endpoint
I believe we should get rid of the code that disables the DELETE endpoint on collections via a specific setting.

IHMO, this should be : 
* out of cliquet scope (eg. do in readinglist endpoint override)
* applied to any endpoints through generic configuration (setting like ``{method}_{collection|record}_{resource}_enabled``)
* through options during registering of resource (see #12)

leplatrem:
@Natim r?

### DIFF
--- a/cliquet/resource.py
+++ b/cliquet/resource.py
@@ -144,6 +144,16 @@ class ViewSet(object):
         service_arguments.update(self.service_arguments)
         return service_arguments
 
+    def is_endpoint_enabled(self, endpoint_type, resource_name, method,
+                            config):
+        """Returns if the given endpoint is enabled or not.
+
+        Uses the settings to tell so.
+        """
+        setting_enabled = 'cliquet.%s_%s_%s_enabled' % (
+            endpoint_type, resource_name, method.lower())
+        return config.get(setting_enabled, True)
+
 
 def register(depth=1, **kwargs):
     """Ressource class decorator.
@@ -166,10 +176,6 @@ def register_resource(resource, settings=None, viewset=None, depth=1,
         The resource class to register.
         It should be a class or have a "name" attribute.
 
-    :param settings:
-        A dict of settings. It will be used to check if views aren't disabled
-        before registering them.
-
     :param viewset:
         A ViewSet object, which will be used to find out which arguments should
         be appended to the views, and where the views are.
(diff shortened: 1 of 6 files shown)


cliquet/resource.py
```diff
@@ -230,19 +229,19 @@ def register_service(endpoint_type):
...
 
-    services = [register_service('collection'), register_service('record')]
-
     def callback(context, name, ob):
         # get the callbacks registred by the inner services
         # and call them from here when the @resource classes are being
         # scanned by venusian.
+        config = context.config.with_package(info.module)
+        services = [register_service('collection', config.registry.settings),
+                    register_service('record', config.registry.settings)]
```
leplatrem: Caution: the parameter is called ``config`` and you pass ``settings``

cliquet/tests/resource/test_viewset.py
```diff
@@ -277,9 +310,15 @@ def test_record_views_are_registered_in_cornice(self, service_class):
...
-        register_resource(self.resource, viewset=self.viewset, settings={
+        venusian_callback = register_resource(
+            self.resource, viewset=self.viewset)
+
+        context = mock.MagicMock()
+        context.registry.settings = {
             'cliquet.record_fake_put_enabled': False
-        })
+        }
+        context.config.with_package.return_value = context
```
leplatrem: This is weird, but I believe there is nothing better :|

Natim: Are you sure that it works without mocks ? I mean that with_package actually returns you that?

cliquet/resource.py
ametaireau: updated.

cliquet/tests/resource/test_viewset.py
ametaireau: http://pyramid.readthedocs.org/en/latest/api/config.html#pyramid.config.Configurator.with_package

leplatrem:
We should remove/merge this : https://github.com/mozilla-services/cliquet/commit/82eae2a0253b94bfc663968a89ab1ee5be7ab539

ametaireau:
Updated.

PR #321, opened 2015-06-16
Raise 400 if specified permission name is invalid

Natim:
Currently the ``PermissionSchema`` is able to raise invalid errors if its ``known_perms`` tuple is set:

* https://github.com/mozilla-services/cliquet/blob/2.0.dev4/cliquet/tests/test_schema.py#L85-L90

However this tuple is not set when instanciated:

* https://github.com/mozilla-services/cliquet/blob/2.0.dev4/cliquet/resource.py#L103

The list of available permissions for a resource is only used to delete/replace existing permission from the permissio backend:

* https://github.com/mozilla-services/cliquet/blob/2.0.dev4/cliquet/resource.py#L1038

We should plug the two together !

### DIFF
--- a/cliquet/resource.py
+++ b/cliquet/resource.py
@@ -100,7 +100,9 @@ class ViewSet(object):
 
         class RecordPayload(colander.MappingSchema):
             data = resource.mapping
-            permissions = PermissionsSchema(missing=colander.drop)
+            permissions = PermissionsSchema(
+                missing=colander.drop,
+                permissions=getattr(resource, 'permissions', tuple()))
 
             def schema_type(self, **kw):
                 return colander.Mapping(unknown='raise')
--- a/cliquet/resource.py
+++ b/cliquet/resource.py
@@ -98,11 +98,14 @@ class ViewSet(object):
             # Simply validate that posted body is a mapping.
             return colander.MappingSchema(unknown='preserve')
 
+        # XXX: https://github.com/mozilla-services/cliquet/issues/322
+        resource_permissions = getattr(resource, 'permissions', tuple())
+
         class RecordPayload(colander.MappingSchema):
             data = resource.mapping
             permissions = PermissionsSchema(
                 missing=colander.drop,
-                permissions=getattr(resource, 'permissions', tuple()))
+                permissions=resource_permissions)
 
             def schema_type(self, **kw):
                 return colander.Mapping(unknown='raise')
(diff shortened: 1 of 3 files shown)


cliquet/tests/resource/test_views.py
```diff
@@ -88,6 +88,24 @@ def test_permissions_are_not_modified_if_not_specified(self):
...
+        self.assertDictEqual(resp.json, {
+            u'errno': ERRORS.INVALID_PARAMETERS,
+            u'message': u'permissions in body: "unknown" '
+                        u'is not one of read, write',
+            u'code': 400,
+            u'error': u'Invalid parameters',
+            u'details': [{u'description': u'"unknown" is not one '
+                                          u'of read, write',
+                          u'location': u'body',
+                          u'name': u'permissions'}]})
```
leplatrem: Are you sure that ``u`` is mandatory ?

cliquet/schema.py
```diff
@@ -82,7 +82,9 @@ class PermissionsSchema(colander.SchemaNode):
         }
 
     """
-    known_perms = tuple()
+    def __init__(self, *args, **kwargs):
+        self.known_perms = kwargs.pop('permissions', tuple())
```
leplatrem: +1

cliquet/tests/resource/test_views.py
Natim: Yes unfortunately. I wish I could get rid of them :'(

cliquet/resource.py
```diff
@@ -100,7 +100,9 @@ def get_record_schema(self, resource, method):
 
         class RecordPayload(colander.MappingSchema):
             data = resource.mapping
-            permissions = PermissionsSchema(missing=colander.drop)
+            permissions = PermissionsSchema(
+                missing=colander.drop,
+                permissions=getattr(resource, 'permissions', tuple()))
```
leplatrem: An intermediary variable would have been nice here ;)

Plus a small ``XXX`` comment towards issue #322 f

cliquet/tests/resource/test_views.py
leplatrem: IHMO you could only test this way:

``self.assertIn('is not one of read, write', resp.json['message'])``

leplatrem: you don't need to assert the whole response format, this is not the scope of this test

PR #325, opened 2015-06-16
Add support for PyPy

brouberol:
Cliquet now supports PyPy, by:

* using ``json`` instead of ``ujson`` (with PyPy only), because ``usjon`` is written in C and uses the CPython C API
* using ``psycopg2cffi`` instead of ``psycopg2`` (with PyPy only)

Warning: this is still a work in progress.
- [x] needs JSONB support in ``psycopg2cffi`` (see https://github.com/chtd/psycopg2cffi/issues/36), or a way to handle deserializing JSONB fields
- [x] a mention of pypy support in the documenation

### DIFF
--- a/cliquet/utils.py
+++ b/cliquet/utils.py
@@ -8,16 +8,16 @@ from base64 import b64decode, b64encode
 from binascii import hexlify
 
 # ujson is not installable with pypy
-try:
+try:  # pragma: no cover
     import ujson as json  # NOQA
-except ImportError:
+except ImportError:  # pragma: no cover
     import json  # NOQA
 
 # psycopg2cffi is installed under pypy, instead of psycopg2
 # pragma: no cover
-try:
+try:  # pragma: no cover
     import psycopg2  # NOQA
-except ImportError:
+except ImportError:  # pragma: no cover
     from psycopg2cffi import compat
     compat.register()
     import psycopg2  # NOQA
--- a/cliquet/utils.py
+++ b/cliquet/utils.py
@@ -8,14 +8,13 @@ from base64 import b64decode, b64encode
 from binascii import hexlify
 
 # ujson is not installable with pypy
-try:  # pragma: no cover
+try:
     import ujson as json  # NOQA
 except ImportError:  # pragma: no cover
     import json  # NOQA
 
 # psycopg2cffi is installed under pypy, instead of psycopg2
-# pragma: no cover
-try:  # pragma: no cover
+try:
     import psycopg2  # NOQA
 except ImportError:  # pragma: no cover
     from psycopg2cffi import compat
--- a/cliquet/utils.py
+++ b/cliquet/utils.py
@@ -7,7 +7,11 @@ import time
 from base64 import b64decode, b64encode
 from binascii import hexlify
 
-import ujson as json  # NOQA
+# ujson is not installable with pypy
+try:
+    import ujson as json  # NOQA
+except ImportError:
+    import json
 from cornice import cors
 from colander import null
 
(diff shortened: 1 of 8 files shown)


setup.py
```diff
@@ -24,9 +26,12 @@
     'requests',
     'six',
     'structlog',
-    'ujson',
 ]
 
+# ujson is not pypy compliant, as it uses the CPython C API
+if not installed_with_pypy:
```
Natim: If we do that for ``ujson``, we can probably do it as well for ``psycopgcffi`` in the ``cliquet[postgresql]`` rule

brouberol:
I have been thinking about the suggested implementation, and there's something that bugs me: the interpreter has an effect on the dependencies. If that was only about ``ujson`` not being installable through PyPy, that'd probably be ok. For me, the real issue is the ``psycopg2/psycopg2cffi`` dichotomy. 

We've seen that ``psycopg2cffi`` does not have the same development rate that ``psycopg2``, as some features introduced in PostgreSQL 9.2 (and ``psycopg2`` ~ a year ago -- see https://github.com/psycopg/psycopg2/issues/226) were not yet reflected in ``psycopg2cffi``. My main fear is that this would introduce some maintainability issues between CPython and PyPy.

So I guess what I'm really asking is, why do we want to support PyPy, what is the real gain, and is it worth the risk of introducing two different Postgres python drivers with different development rates?   

My 0.02$.

Natim:
This is interesting. Also we have tests that make sure that the backend does what we want it to do.
So if test are passing with PyPy and psycopg2cffi we can be confident that our storage works with PyPy and psycopg2cffi

brouberol:
So, now that JSONB support landed in psycopg2cffi, all tests are green for the PyPy tox target. I'll wait for a PyPI release and include a ``psycopg2cffi<=VERSION`` clause in the setup.py.

I'll also add some documentation about PyPy support, and the fact that it uses a different python driver for Postgres. 

Also, happy Canada 

Canada!

brouberol:
Side note: we need to wait for the PyPI release of the new psycopg2cffi (with JSONB support) before merging this one.

setup.py
```diff
@@ -24,14 +26,22 @@
...
 
-DEPENDENCY_LINKS = [
-]
+if not installed_with_pypy:
+    # ujson is not pypy compliant, as it uses the CPython C API
+    REQUIREMENTS.append('ujson')
+    # We install psycopg2cffi instead of psycopg2 when dealing with pypy
+    # Note: JSONB support landed after psycopg2cffi 2.7.0
+    POSTGRESQL_REQUIRES = [
+        'psycopg2cffi>2.7.0',
```
leplatrem: Isn't it the opposite?

cliquet/utils.py
```diff
@@ -7,7 +7,21 @@
...
 
-import ujson as json  # NOQA
+# ujson is not installable with pypy
+try:  # pragma: no cover
+    import ujson as json  # NOQA
+except ImportError:  # pragma: no cover
+    import json  # NOQA
+
+# psycopg2cffi is installed under pypy, instead of psycopg2
+# pragma: no cover
```
leplatrem: Some pragmas can be removed IMO

brouberol: You mean removing the "pragma: no cover" in the "except" clause?

leplatrem:
Very good job! Thanks!

I'm sorry that I had put it as easy-pick... Was not so trivial to tackle :/

cliquet/utils.py
leplatrem: Yes, the ones in front of try (always executed unlike except)

brouberol:
Heh, that was a fun one anyway.

Natim:
Good Job @brouberol and thanks you for the fish :)

PR #326, opened 2015-06-17
Prefix userid outside authn policies (fixes #299)

### LINKED ISSUE #299: Prefix authenticated user id outside authentication policies
Instead of doing this:
https://github.com/mozilla-services/cliquet/blob/acd7f1700ed8a32baf5d895816671675d8de330a/cliquet/authentication.py#L34

And this:
https://github.com/mozilla-services/cliquet-fxa/blob/c00e15028078c742be478c94e14b9bfa7a1b50ba/cliquet_fxa/authentication.py#L87

I suggest that we rely on this : 
https://github.com/mozilla-services/cliquet/blob/acd7f1700ed8a32baf5d895816671675d8de330a/cliquet/initialization.py#L85
 
And do something like this:

```python
request.prefixed_authenticated_userid = request.authn_type + ':' + request.authenticated_userid
```

This way, we will be able to have a prefixed userid for any authentication policy available in the Pyramid ecosystem.

leplatrem:
The main idea of this is to be able to use any kind of authn policy available for Pyramid.

Prefixing user_id with authentication type should thus be done by Cliquet.

See also https://github.com/mozilla-services/cliquet-fxa/pull/5

### DIFF
--- a/cliquet/authorization.py
+++ b/cliquet/authorization.py
@@ -19,8 +19,17 @@ class AuthorizationPolicy(object):
     get_bound_permissions = None
 
     def permits(self, context, principals, permission):
-        # User id prefixed with authentication policy type.
-        principals.insert(0, context.prefixed_userid)
+        # Add prefixed user id to principals.
+        if context.prefixed_userid:
+            principals = principals + [context.prefixed_userid]
+            prefix, user_id = context.prefixed_userid.split(':', 1)
+            # Remove unprefixed user id to avoid conflicts.
+            # (it is added via Pyramid Authn policy effective principals)
+            if user_id in principals:
+                principals.remove(user_id)
+            # Retro-compatibility with cliquet 2.0 '_' user id prefixes.
+            # Just in case it was used in permissions definitions.
+            principals.append('%s_%s' % (prefix, user_id))
 
         if permission == DYNAMIC:
             permission = context.required_permission
@@ -52,6 +61,15 @@ class RouteFactory(object):
     }
 
     def __init__(self, request):
+        # Prefix the user id with the authn policy type name.
+        user_id = request.authenticated_userid
+        # This comes from ``cliquet.initialization.setup_authentication()``.
+        authn_type = getattr(request, 'authn_type', None)
+        self.prefixed_userid = None
+        if user_id and authn_type:
+            self.prefixed_userid = '%s:%s' % (authn_type.lower(), user_id)
+
+        # Store service, resource, record and required permission.
         service = utils.current_service(request)
 
         is_on_resource = (service is not None and
@@ -75,7 +93,7 @@ class RouteFactory(object):
 
                 # If the view exists, call it with the request and catch an
                 # eventual NotFound.
-                resource = service.resource(request)
+                resource = service.resource(request=request, context=self)
                 try:
                     resource.collection.get_record(resource.record_id)
                 except storage_exceptions.RecordNotFoundError:
@@ -99,13 +117,6 @@ class RouteFactory(object):
         self.required_permission = permission
         self.resource_name = resource_name
         self.check_permission = check_permission
-
-        prefixed_userid = None
-        if request.authenticated_userid:
-            authn_type = request.authn_type.lower()
-            user_id = request.authenticated_userid
-            prefixed_userid = '%s_%s' % (authn_type, user_id)
-        self.prefixed_userid = prefixed_userid
 
 
 def get_object_id(object_uri):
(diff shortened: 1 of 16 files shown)


cliquet/authorization.py
```diff
@@ -19,6 +19,9 @@ class AuthorizationPolicy(object):
     get_bound_permissions = None
 
     def permits(self, context, principals, permission):
+        # User id prefixed with authentication policy type.
+        principals.insert(0, context.prefixed_userid)
```
ametaireau: Why do you need to put it in the first position?

Natim: We don't need to add it in first position, also we should remove the ``authenticated_userid`` from the principals

cliquet/resource.py
```diff
@@ -258,7 +258,8 @@ class BaseResource(object):
     def __init__(self, request, context=None):
 
         # Collections are isolated by user.
-        parent_id = request.authenticated_userid
+        parent_id = (context.prefixed_userid if context else
+                     request.authenticated_userid)
```
Natim: We should not fallback on ``request.authenticated_userid``

ametaireau: If we have authentication, then context is set. Why do we need the fallback?

cliquet/tests/resource/test_object_permissions.py
```diff
@@ -71,8 +71,8 @@ def setUp(self):
         record_id = record['id']
         record_uri = '/articles/%s' % record_id
         self.permission.add_principal_to_ace(record_uri, 'read', 'fxa:user')
+        self.resource.context.prefixed_userid = 'basic:userid'
```
Natim: should probably be an _ instead of :

cliquet/resource.py
leplatrem: What do I do with parent_id if context is None then ?

Natim: I would set parent_id to None in that case but it should not happens.

cliquet/authorization.py
```diff
@@ -19,6 +19,18 @@ class AuthorizationPolicy(object):
...
+        if context.prefixed_userid:
+            principals = principals + [context.prefixed_userid]
+            prefix, user_id = context.prefixed_userid.split(':', 1)
+            # Remove unprefixed user id to avoid conflicts.
+            # (it is added via Pyramid Authn policy effective principals)
+            if user_id in principals:
+                principals.remove(user_id)
+            # Retro-compatibility with cliquet 2.0 '_' user id prefixes.
+            # Just in case it was used in permissions definitions.
+            principals.append('%s_%s' % (prefix, user_id))
```
Natim: Do we really want to handle retrocompatibility here since we don't have anything in production yet?

cliquet/tests/support.py
```diff
@@ -118,9 +120,8 @@ def _create_thread(self, *args, **kwargs):
 @implementer(IAuthorizationPolicy)
 class AllowAuthorizationPolicy(object):
     def permits(self, context, principals, permission):
-        if USER_PRINCIPAL in principals:
-            return True
-        return False
+        # Cliquet default authz policy uses prefixed_userid.
+        return USER_PRINCIPAL in (principals + [context.prefixed_userid])
```
Natim: Should we remove user_id from principals as well here?

cliquet/authorization.py
leplatrem: Well, as you want, I thought it was cheap and safe.

cliquet/authorization.py
```diff
@@ -49,6 +61,15 @@ class RouteFactory(object):
     }
 
     def __init__(self, request):
+        # Prefix the user id with the authn policy type name.
+        user_id = request.authenticated_userid
+        # This comes from ``cliquet.initialization.setup_authentication()``.
+        authn_type = getattr(request, 'authn_type', None)
+        self.prefixed_userid = None
+        if user_id and authn_type:
+            self.prefixed_userid = '%s:%s' % (authn_type.lower(), user_id)
```
leplatrem: OH ! This won't work with previous release of cliquet-fxa.
I should do test if already prefixed ! ``if not user_id.startswith(authn_type + '_')``

cliquet/tests/support.py
leplatrem: I don't think so since this policy is used to allow ``USER_PRINCIPAL`` only. But I may be wrong.

Natim: No you are right :)

cliquet/authorization.py
leplatrem: Nevermind, I just released cliquet-fxa 1.1.0 and we'll consider that cliquet-fxa was never used by anybody (only mozilla apps can use it)

PR #329, opened 2015-06-18
Remove references to IUS / IMS.

ametaireau:
This was already removed from the code but not from the docstrings.

### DIFF
--- a/cliquet/resource.py
+++ b/cliquet/resource.py
@@ -299,12 +299,12 @@ class BaseResource(object):
         """Collection ``GET`` endpoint: retrieve multiple records.
 
         :raises: :exc:`~pyramid:pyramid.httpexceptions.HTTPNotModified` if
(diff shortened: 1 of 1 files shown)


cliquet/resource.py
```diff
@@ -299,12 +299,12 @@ def collection_get(self):
         """Collection ``GET`` endpoint: retrieve multiple records.
 
         :raises: :exc:`~pyramid:pyramid.httpexceptions.HTTPNotModified` if
-            ``If-Modified-Since`` header is provided and collection not
+            ``If-Match`` header is provided and collection not
```
Natim: It is the reverse modification, ``If-Modified-Since`` is ``If-None-Match`` and ``If-Unmodified-Since`` is ``If-Match``

leplatrem:
I'll update this.

Natim:
:dancer:

PR #331, opened 2015-06-19
Using a wrong HTTP verb raise a 500

Natim:
I sent:

```echo '{"data": {}}' | http PUTT https://kinto.dev.mozaws.net/v1/buckets/default --auth 'user:password'```

```  File "/home/ubuntu/venvs/kinto/local/lib/python2.7/site-packages/cliquet/authorization.py", line 85, in __init__
    permission = self.method_permissions[request.method.lower()]
KeyError: 'putt'```

### DIFF
--- a/cliquet/authorization.py
+++ b/cliquet/authorization.py
@@ -103,7 +103,11 @@ class RouteFactory(object):
                 else:
                     permission = "write"
             else:
-                permission = self.method_permissions[request.method.lower()]
+                method = request.method.lower()
+                try:
+                    permission = self.method_permissions[method]
+                except KeyError:
+                    permission = None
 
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: cliquet/authorization.py [hunk 1]
```diff
@@ -103,7 +103,11 @@ def __init__(self, request):
                 else:
                     permission = "write"
             else:
-                permission = self.method_permissions[request.method.lower()]
+                method = request.method.lower()
+                try:
+                    permission = self.method_permissions[method]
+                except KeyError:
+                    permission = None
```
leplatrem: isn't it ``self.method_permissions.get()`` then ?

Natim: Good point, thanks :+1:

PR #332, opened 2015-06-19
Kinto is broken with cliquet master

Natim:
Running:

```
echo '{"data": {}}' | http PUT http://localhost:8888/v1/buckets/default --auth 'user:password'
```
Will bring you this stacktrace:

```
  File "~/mozilla/kinto/kinto/views/buckets.py", line 14, in __init__
    super(Bucket, self).__init__(*args, **kwargs)
TypeError: __init__() takes exactly 3 arguments (2 given)
```

This is because of this commit: https://github.com/mozilla-services/cliquet/commit/d870a0110586157b4e3972f90fbc0aa533503fc7

That remove the ``context=None`` fallback.

### DIFF
--- a/cliquet/authorization.py
+++ b/cliquet/authorization.py
@@ -61,13 +61,8 @@ class RouteFactory(object):
     }
 
     def __init__(self, request):
-        # Prefix the user id with the authn policy type name.
-        user_id = request.authenticated_userid
-        # This comes from ``cliquet.initialization.setup_authentication()``.
-        authn_type = getattr(request, 'authn_type', None)
-        self.prefixed_userid = None
-        if user_id and authn_type:
-            self.prefixed_userid = '%s:%s' % (authn_type.lower(), user_id)
+        # We need it in permits as well
+        self.prefixed_userid = getattr(request, "prefixed_userid", None)
 
         # Store service, resource, record and required permission.
         service = utils.current_service(request)
--- a/cliquet/authorization.py
+++ b/cliquet/authorization.py
@@ -61,7 +61,7 @@ class RouteFactory(object):
     }
 
     def __init__(self, request):
-        # We need it in permits as well
+        # Make it available for the authorization policy.
         self.prefixed_userid = getattr(request, "prefixed_userid", None)
 
         # Store service, resource, record and required permission.
--- a/cliquet/initialization.py
+++ b/cliquet/initialization.py
@@ -88,6 +88,21 @@ def setup_authentication(config):
 
     config.add_subscriber(on_policy_selected, MultiAuthPolicySelected)
 
+    # Build the prefixed_userid
+    def on_new_request(event):
+        # This comes from ``cliquet.initialization.setup_authentication()``.
+        authn_type = getattr(event.request, 'authn_type', None)
+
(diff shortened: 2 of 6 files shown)


leplatrem:
r-, no test :)

And I believe that instantiation a route factory is not the scope of resource

Natim:
I shift one's ground r? @leplatrem

cliquet/initialization.py
```diff
@@ -88,6 +88,21 @@ def on_policy_selected(event):
 
     config.add_subscriber(on_policy_selected, MultiAuthPolicySelected)
 
+    # Build the prefixed_userid
+    def on_new_request(event):
+        # This comes from ``cliquet.initialization.setup_authentication()``.
```
leplatrem: url

cliquet/authorization.py
```diff
@@ -61,13 +61,8 @@ class RouteFactory(object):
...
 
     def __init__(self, request):
-        # Prefix the user id with the authn policy type name.
-        user_id = request.authenticated_userid
-        # This comes from ``cliquet.initialization.setup_authentication()``.
-        authn_type = getattr(request, 'authn_type', None)
-        self.prefixed_userid = None
-        if user_id and authn_type:
-            self.prefixed_userid = '%s:%s' % (authn_type.lower(), user_id)
+        # We need it in permits as well
```
leplatrem: ``# Make it available for the authorization policy.``

cliquet/initialization.py
```diff
@@ -88,6 +88,21 @@ def on_policy_selected(event):
...
+
+        # Prefix the user id with the authn policy type name.
+        user_id = event.request.authenticated_userid
+
+        event.request.prefixed_userid = None
+        if user_id and authn_type:
+            event.request.prefixed_userid = '%s:%s' % (authn_type.lower(),
+                                                       user_id)
+
+    config.add_subscriber(on_new_request, NewRequest)
```
leplatrem: Are we sure that this is triggered **after** the policiy selected event ?

cliquet/resource.py
```diff
@@ -255,9 +255,9 @@ class BaseResource(object):
     mapping = ResourceSchema()
     """Schema to validate records."""
 
-    def __init__(self, request, context):
+    def __init__(self, request, context=None):
```
leplatrem: I see no test that makes sure resource constructor context is optional, due a bug in Cornice here https://github.com/mozilla-services/cornice/blob/master/cornice/service.py#L535-L536

cliquet/initialization.py
Natim: :+1:

cliquet/initialization.py
Natim: As far as I know yes.

Natim: Be I am more confident than sure.

cliquet/resource.py
Natim: I will add one.

leplatrem:
capture d ecran de 2015-06-19 17 08 30

Which one is to consider ?

Natim:
Merged in #330

PR #330, opened 2015-06-19
Sometimes the collection_timestamp < record.last_modified

Natim:
```http
HTTP/1.1 200 OK
Access-Control-Expose-Headers: Backoff, Retry-After, Alert, Next-Page, Total-Records, Last-Modified, ETag
Backoff: 10
Connection: keep-alive
Content-Length: 104
Content-Type: application/json; charset=UTF-8
Date: Fri, 19 Jun 2015 08:31:46 GMT
ETag: "1434702683855"
Last-Modified: Fri, 19 Jun 2015 08:31:23 GMT
Server: nginx/1.4.6 (Ubuntu)
Total-Records: 1

{
    "data": [
        {
            "id": "64a8f2a2-2f32-4b72-9bc1-0b1a244f998a", 
            "last_modified": 1434702683856, 
            "title": "amanite"
        }
    ]
}
```

As you can see:

``ETag: "1434702683855"`` and ``"last_modified": 1434702683856, ``

So basically ``1434702683855 < 1434702683856`` which should not be possible ever.

Here I was using the redis backend.

### DIFF
--- a/cliquet/resource.py
+++ b/cliquet/resource.py
@@ -257,7 +257,7 @@ class BaseResource(object):
 
     def __init__(self, request, context=None):
         # Collections are isolated by user.
-        parent_id = request.prefixed_userid
+        parent_id = self.get_parent_id(request)
 
         # Authentication to storage is transmitted as is (cf. cloud_storage).
         auth = request.headers.get('Authorization')
@@ -277,6 +277,9 @@ class BaseResource(object):
         # Log resource context.
         logger.bind(collection_id=self.collection.collection_id,
                     collection_timestamp=self.timestamp)
+
+    def get_parent_id(self, request):
+        return getattr(request, 'prefixed_userid', None)
 
     def is_known_field(self, field):
         """Return ``True`` if `field` is defined in the resource mapping.
--- a/cliquet/resource.py
+++ b/cliquet/resource.py
@@ -279,6 +279,15 @@ class BaseResource(object):
                     collection_timestamp=self.timestamp)
 
     def get_parent_id(self, request):
+        """Return the parent_id of the resource with regards to the current
+        request.
+
+        :param request:
+            The request used to create the resource.
+
+        :rtype: str
+
+        """
         return getattr(request, 'prefixed_userid', None)
 
     def is_known_field(self, field):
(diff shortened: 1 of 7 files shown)


cliquet/tests/support.py
```diff
@@ -34,8 +34,8 @@ def __init__(self, *args, **kwargs):
         self.headers = {}
         self.errors = cornice_errors.Errors(request=self)
         self.authenticated_userid = 'bob'
-        self.authn_type = 'basicauth'
-        self.prefixed_userid = 'basicauth:bob'
+        self.authn_type = 'basic'
+        self.prefixed_userid = 'basic:bob'
```
leplatrem: I would have left ``basicauth`` since this is the actual value for ``BasicAuthAuthenticationPolicy``

cliquet/tests/resource/test_object_permissions.py
```diff
@@ -16,6 +16,16 @@ def get_request(self):
         return request
 
 
+class ProtectedResourceTest(BaseTest):
```
leplatrem: I would have moved this somewhere else, maybe in ``test.resource.__init__`` ?

cliquet/resource.py
```diff
@@ -278,6 +278,9 @@ def __init__(self, request, context):
         logger.bind(collection_id=self.collection.collection_id,
                     collection_timestamp=self.timestamp)
 
+    def get_parent_id(self, request):
```
leplatrem: Why passing the request? 

Since this is becoming public API, it deserves a docstring and some tests :)

cliquet/resource.py
```diff
@@ -278,6 +278,9 @@ def __init__(self, request, context):
...
+        return getattr(request, 'prefixed_userid', None)
```
leplatrem: What happens if parent_id is None ? AFAIK it fails in Collection.

leplatrem:
It looks like diff does not reflect the issue label anymore (or at least, there is no new test asserting that a bug was fixed)

cliquet/resource.py
Natim: Because you need the request in order to get the request and get the parent_id :)

Natim:
The problem is that the ``collection_timestamp`` was got with the ``prefixed_userid`` and set on the right ``parent_id`` reason why it wasn't updated.

cliquet/resource.py
Natim: It might be good behavior to fail in that case.

Natim:
I want to add a test in kinto for the test you are talking about.

PR #339, opened 2015-06-27
Remove UserWarning exception when running tests

elemoine:
This removes `UserWarning` exceptions that occur when running the initialization tests:

```
./home/elemoine/src/cliquet/cliquet/initialization.py:357: UserWarning: No value specified for `project_name` warnings.warn('No value specified for `project_name`')
```


Natim:
Cool thanks.

PR #341, opened 2015-06-30
Return a 404 and not a 307 in case we are trying to access an unsupported version.

Natim:
r? @n1k0

### DIFF
--- a/cliquet/initialization.py
+++ b/cliquet/initialization.py
@@ -49,11 +49,7 @@ def setup_version_redirection(config):
     version_prefix_redirection_enabled = asbool(redirect_enabled)
 
     route_prefix = config.route_prefix
-
-    def on_new_request(event):
-            event.request.registry.route_prefix = route_prefix
-
-    config.add_subscriber(on_new_request, NewRequest)
+    config.registry.route_prefix = route_prefix
 
     # Redirect to the current version of the API if the prefix isn't used.
     # Do not redirect if cliquet.version_prefix_redirect_enabled is set to
--- a/cliquet/initialization.py
+++ b/cliquet/initialization.py
@@ -65,7 +65,7 @@ def setup_version_redirection(config):
     config.route_prefix = None
 
     config.add_route(name='redirect_to_version',
-                     pattern='{path:(?!v[0-9]+).*}')
+                     pattern='/{path:(?!v[0-9]+).*}')
 
     config.add_view(view=_redirect_to_version_view,
                     route_name='redirect_to_version',
--- a/cliquet/initialization.py
+++ b/cliquet/initialization.py
@@ -49,7 +49,11 @@ def setup_version_redirection(config):
     version_prefix_redirection_enabled = asbool(redirect_enabled)
 
     route_prefix = config.route_prefix
-    config.add_settings(route_prefix=route_prefix)
+
+    def on_new_request(event):
+            event.request.registry.route_prefix = route_prefix
+
+    config.add_subscriber(on_new_request, NewRequest)
 
     # Redirect to the current version of the API if the prefix isn't used.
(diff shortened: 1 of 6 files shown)


Natim:
As per discussion in https://github.com/mozilla-services/cliquetis/pull/39

The client is expecting feature that the server is not supporting (yet or anymore)

In that case it looks more as a server error for me.

What do you think of using **501 Unimplemented** in that case?

It was first created as an error when switching from HTTP/1.0 to HTTP/1.1 but it is now used by other web service for this purpose. https://en.wikipedia.org/wiki/List_of_HTTP_status_codes#5xx_Server_Error

Another solution would be to return a 403 in case we consider that the client should not use this server like that.

I also like the 404 because it means the expected version was not found.

elemoine:
501 often implies future availability, which isn't the case here I think. So 403 and 404 are better IMO. Both work I think, although using 403 (Forbidden) sounds more confusing to me, as it's usually used as an authorization error.

cliquet/initialization.py
```diff
@@ -48,6 +48,9 @@ def setup_version_redirection(config):
     redirect_enabled = settings['cliquet.version_prefix_redirect_enabled']
     version_prefix_redirection_enabled = asbool(redirect_enabled)
 
+    route_prefix = config.route_prefix
+    config.add_settings(route_prefix=route_prefix)
```
elemoine: It sounds to me that if `route_prefix` is set in the `settings` then it should not be set directly on the `config` object as done currently. This was a hack anyway.

And if you set `route_prefix` in the `settings` you may want to use the name `cliquet.route_prefix` to avoid clashes with application-level names.

n1k0:
FWIW, 404 with an error code in the json response body is what the titter API does on unsupported version.

elemoine:
> FWIW, 404 with an error code in the json response body is what the twitter API does on unsupported version.

Yeah, makes sense to me.

cliquet/initialization.py
Natim: I think you are right, we should add the route_prefix to the registry rather than adding it in the settings.

cliquet/initialization.py
```diff
@@ -48,6 +48,13 @@ def setup_version_redirection(config):
     redirect_enabled = settings['cliquet.version_prefix_redirect_enabled']
     version_prefix_redirection_enabled = asbool(redirect_enabled)
 
+    route_prefix = config.route_prefix
+
+    def on_new_request(event):
+            event.request.registry.route_prefix = route_prefix
+
+    config.add_subscriber(on_new_request, NewRequest)
```
leplatrem: This piece of code is unnecessary, since you can do : 
```
config.registry.route_prefix = config.route_prefix
```

cliquet/initialization.py
```diff
@@ -59,11 +66,10 @@ def _redirect_to_version_view(request):
             '/%s/%s' % (route_prefix, request.matchdict['path']))
 
     # Disable the route prefix passed by the app.
-    route_prefix = config.route_prefix
     config.route_prefix = None
 
     config.add_route(name='redirect_to_version',
-                     pattern='/{path:(?!%s).*}' % route_prefix)
+                     pattern='{path:(?!v[0-9]+).*}')
```
leplatrem: Just curious: why no trailing slash anymore ?

elemoine: You probably mean "leading" slash :)

Natim: The route doesn't starts with a /
They were a bug in the previous regexp. 

Actually it was a catch all. Now it only catches if it doesn't starts with a v.

this makes me thing that we should probably add a ``^`` at the beginning to make sure we matches the right ``v[0-9]+``

PR #342, opened 2015-06-30
Use default settings only if the value was not defined yet.

Natim:
r? @leplatrem


cliquet/initialization.py
```diff
@@ -338,11 +338,14 @@ def initialize(config, version=None, project_name=None, default_settings=None):
...
-    if default_settings:
-        config.add_settings(default_settings)
-
     settings = config.get_settings()
 
+    if default_settings:
+        for key, value in default_settings.items():
+            if not settings.get(key):
+                settings[key] = value
+        config.add_settings(settings)
```
elemoine: Pretty minor, but isn't this call to `config.set_settings` unnecessary?

cliquet/__init__.py
```diff
@@ -26,7 +26,7 @@
 DEFAULT_SETTINGS = {
     'cliquet.backoff': None,
     'cliquet.batch_max_requests': 25,
-    'cliquet.cache_backend': 'cliquet.cache.redis',
+    'cliquet.cache_backend': 'cliquet.cache.memory',
```
leplatrem: IMO this is not the scope of this PR :)

cliquet/initialization.py
leplatrem: I think so too, unless ``config.get_settings()`` returns a copy

cliquet/initialization.py
```diff
@@ -338,11 +338,14 @@ def initialize(config, version=None, project_name=None, default_settings=None):
...
     :param dict default_settings: Override cliquet default settings values.
     """
-    if default_settings:
-        config.add_settings(default_settings)
-
     settings = config.get_settings()
 
+    if default_settings:
+        for key, value in default_settings.items():
+            if not settings.get(key):
```
leplatrem: We may want to use/merge/refactor ``cliquet.load_default_settings()`` since it does something pretty similar

cliquet/tests/test_initialization.py
```diff
@@ -60,13 +60,30 @@ def test_set_the_project_name_from_settings_even_if_specified(self):
...
+        config = Configurator()
+        defaults = {'cliquet.project_name':  'cliquet'}
+        cliquet.initialize(config, '0.0.1', 'project_name',
+                           default_settings=defaults)
+        self.assertEqual(config.registry.settings['cliquet.project_name'],
+                         'cliquet')
+
+    def test_specified_not_overridded_by_default_settings(self):
+        config = Configurator()
+        config.add_settings({"cliquet.project_name": "readinglist"})
```
leplatrem: nit: unconsistent double quotes

cliquet/tests/test_initialization.py
```diff
@@ -60,13 +60,30 @@ def test_set_the_project_name_from_settings_even_if_specified(self):
...
                            default_settings=defaults)
         self.assertEqual(config.registry.settings['cliquet.paginate_by'], 102)
 
+    def test_overriden_empty_default_settings(self):
+        config = Configurator()
+        defaults = {'cliquet.project_name':  'cliquet'}
+        cliquet.initialize(config, '0.0.1', 'project_name',
+                           default_settings=defaults)
+        self.assertEqual(config.registry.settings['cliquet.project_name'],
+                         'cliquet')
```
leplatrem: what is the difference with the previous test ?

cliquet/initialization.py
elemoine: It does not.

cliquet/__init__.py
Natim: It is why I created the PR in the first place and it is the only thing left that this PR doesn't handle (having different default for kinto and cliquet on some settings.)

cliquet/tests/test_initialization.py
Natim: The previous test is adding a settings that doesn't exists.

cliquet/__init__.py
leplatrem: I'm sorry but I don't follow you here.

If this PR is supposed to fix the handling of default settings, why do we have to put the same default backend settings in Cliquet and Kinto ?

Natim: This PR prevents overriding the user settings (in her .ini file) with Kinto default settings.

What this PR doesn't do (and maybe should do) is to let kinto override cliquet default settings.

cliquet/initialization.py
Natim: If we could manage to read settings in that order that would be much better:

 - cliquet.DEFAULT_SETTINGS
 - kinto.DEFAULT_SETTINGS (updates keys from cliquet.DEFAULT_SETTINGS)
 - read the user config.ini

cliquet/__init__.py
leplatrem: Ok I get you.

IMO switching from Redis to Memory is still out of scope :) :hear_no_evil:

cliquet/initialization.py
leplatrem: Back to the original spec of this ``default_settings``: «*Override cliquet default settings values*»

It could become something like this:
```python
for key, value in default_settings.items():
    if settings[key] == cliquet.DEFAULT_SETTINGS[key]:
        # Override only if not set by user.
        settings[key]  = value
```

cliquet/tests/test_initialization.py
leplatrem: But ``'cliquet.paginate_by'`` is set to None and ``'cliquet.project_name'`` to ``''`` in DEFAULT_SETTINGS, so both tests cover the current ``if not settings.get(key)`` in the code

cliquet/initialization.py
Natim: It is what @ametaireau was suggesting this morning, but since we are doing:

 - cliquet.DEFAULT_SETTINGS
 - read the user config.ini
 - kinto.DEFAULT_SETTINGS (updates keys from cliquet.DEFAULT_SETTINGS)

If the user is defining something that is inside the cliquet.DEFAULT_SETTINGS then kinto.DEFAULT_SETTINGS might force an unwanted settings.

Let say, cliquet defines: ``'cliquet.cache_backend': 'cliquet.cache.redis',`` and kinto propose ``'cliquet.cache_backend': 'cliquet.cache.memory',`` then it will be impossible to use redis with kinto because if the user configure redis it will be overrided to memory by kinto.

leplatrem: Yes indeed.

But we could still use ``cliquet.load_default_settings()`` in Kinto before loading Cliquet then.
Like suggested here http://cliquet.readthedocs.org/en/latest/ecosystem.html#configuration

Sorry, my brain looks like that currently:
xp_defrag 3

Natim: I don't follow you :sandal:

cliquet/tests/test_initialization.py
Natim: Yes you are right my bad.

Natim:
The path taken by https://github.com/mozilla-services/cliquet/pull/343 is much better.

PR #343, opened 2015-07-01
Override default settings instead of final settings

leplatrem:
My proposition instead of #342

### DIFF
--- a/cliquet/initialization.py
+++ b/cliquet/initialization.py
@@ -16,6 +16,7 @@ try:
 except ImportError:  # pragma: no cover
     pass
 
+import cliquet
 from cliquet import errors
 from cliquet import logger
 from cliquet import utils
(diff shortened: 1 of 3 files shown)


Natim:
Cool :+1:

leplatrem:
As usual, I was not able to think properly until I put a hand on some code :((

@Natim I haven't tested with Kinto though, but it looks consistent

Natim:
I don't know if we can add a test to make sure default_settings (both from cliquet and the app) are overridden by the config.ini file.

leplatrem:
That's what I was telling you here: https://github.com/mozilla-services/kinto/pull/90#issuecomment-117616892

Settings of ``[app:main]`` section of ``.ini`` arrive in ``**settings`` arg of ``main()``. https://github.com/mozilla-services/kinto/blob/master/kinto/__init__.py#L28-L29

Natim:
Ok perfect :)

cliquet/initialization.py
```diff
@@ -340,21 +341,25 @@ def initialize(config, version=None, project_name=None, default_settings=None):
         in application settings.
     :param dict default_settings: Override cliquet default settings values.
     """
+    cliquet_defaults = cliquet.DEFAULT_SETTINGS.copy()
+
     if default_settings:
-        config.add_settings(default_settings)
+        cliquet_defaults.update(default_settings)
```
ametaireau: Thanks :)

PR #345, opened 2015-07-02
Add public settings in hello view (fixes #318)

### LINKED ISSUE #318: Expose batch endpoint configuration in the API.
Clients need to know what's the maximum batch operations they can do, in order to split their calls accordingly.

We should expose that in either an extra endpoint or the root one. I would advocate to create another endpoint exposing some public configuration values.

leplatrem:
r? @Natim @n1k0

### DIFF
--- a/cliquet/__init__.py
+++ b/cliquet/__init__.py
@@ -127,6 +127,9 @@ def includeme(config):
     # Heartbeat registry.
     config.registry.heartbeats = {}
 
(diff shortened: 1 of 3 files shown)


Natim:
r+

ametaireau:
This actually needs an update of the documentation to be landed. https://github.com/mozilla-services/cliquet/blob/master/docs/api/utilities.rst

PR #347, opened 2015-07-02
Specify database encoding and timezone in doc

elemoine:
Cliquet requires that `UTC` be used as the Postgres database timezone, and `UTF-8` as the encoding. This PR mentions this in the documentation.

I think the documentation still needs some love, I'll see if I can find time for more changes.

### DIFF
--- a/docs/reference/installation.rst
+++ b/docs/reference/installation.rst
@@ -152,6 +152,14 @@ if ran by the ``postgres`` system user. The following command will assign it:
 
     sudo -u postgres psql -c "ALTER USER postgres PASSWORD 'postgres';"
 
+Cliquet requires that ``UTC`` be used as the database timezone, and that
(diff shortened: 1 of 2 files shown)


ametaireau:
Oh yes, the documentation definitely needs some more love. Any contributions to this would be instrumental. We kinda fail to see what's missing because we're getting used to it.

We've started doing doc changes on kinto, but nothing done for cliquet yet. That's one of the things we really want to get better at.

docs/reference/installation.rst
```diff
@@ -152,6 +152,14 @@ if ran by the ``postgres`` system user. The following command will assign it:
 
     sudo -u postgres psql -c "ALTER USER postgres PASSWORD 'postgres';"
 
+Cliquet requires that ``UTC`` be used as the database timezone, and that
```
ametaireau: requires ``UTC`` to be used as*

PR #349, opened 2015-07-02
Update installation.rst

ametaireau:
@elemoine r?


elemoine:
r+ (learning the language here)

Issue #354, opened 2015-07-04
Obtain list of shared records on collection


leplatrem:
On protected resources, where permissions are enabled, it should be possible for user to list the records on which they have a permission.

# Use cases:

* Bob wants to share some records publicly on his collection. He adds ``system.Everyone`` to the ``read`` permission on some records. Anonymous users can now obtain a list of Bob public records.
* Alice creates records on a collection, where she has no other permission than ``records:create``. She becomes the owner of those records. She must able to list them and delete them.
* Bob creates some records in his collection, on behalf of Alice. He sets the ``read`` permission to Alice on those records. Alice can now see the records that Bob created for her.

# Challenges:

* Couple storage (filtering) and permission backends in an elegant way
* Do not reject guests in authorization policy for list operation on collections

# Several strategies:

**1 Query permission backend to obtain list of ids**
In the resource code, if the user has no perm on the collection itself, add an implicit filter when querying records on storage backend. The filter is a (id, IN, <list of allowed record ids>), and the list of allowed record ids is obtained from the permission backend.

Pros:
* easiest to implement
* good separation of concern

Cons:
* does not scale (``WHERE id IN <millionrecords>``)

**2 Add a permission field in the record data**
Mix data and permissions in the JSON field of storage.
When retrieving the record data from storage, remove the permissions.

Pros:
* seemly easy to implement

Cons:
* no separation of concerns, both backends manage permission notions
* must create DB indices on JSON attributes that contain list values (e.g. ``data->permission->read``)
* transactions management between storage and permission backends ?

**3 Add permissions notion in storage**
Add new table in storage backend, index and query those columns.
Pros:
* efficient (i.e. scales)

Cons:
* no separation of concerns, both backends manage permission notions
* transactions management between storage and permission backends ?

**4 Move permissions to storage**
Leave in permissions the user principals operations.
Manage permissions from storage. Each record has a set of permissions, and the authorization queries the storage backend to know who is allowed.

Pros:
* Clean separation of concern
* Permissions are linked to record using foreign keys, and we get rid of 'object URIs' to refer to objects
* Efficient (scales)

Cons:
* a lot of code to rewrite

leplatrem:
Merged !

PR #355, opened 2015-07-04
Filter collection records for guests (ref #354)

leplatrem:
I decided to follow the first strategy described in #354. Even though the one I prefer is the number 5. But for me hacking some code is a good way to start thinking about it.

* [x] Add a IN filter in storage
* [x] Query the permissions backend to obtain the list of object ids
* [x] Add some hack in authorization to let guest list the collections
* [ ] Make sure normal resources are not affected (**currently open door**), should be only ProtectedResources
* [x] Filter by permissions (by collection name)
* [ ] Think about parent_id etc.
* [ ] Add some functional tests to make sure use cases described in #354 are covered

I take reviews for:
* vetos against this strategy (*i.e. give the feature but does not scale*)
* alternative solutions in the authorization code :)

### DIFF
--- a/cliquet/authorization.py
+++ b/cliquet/authorization.py
@@ -51,6 +51,13 @@ class AuthorizationPolicy(object):
                 principals,
                 get_bound_permissions=self.get_bound_permissions)
 
+        is_list_operation = (context.on_collection and
+                             'create' not in permission)
+        if not allowed and is_list_operation:
+            # Guest trying to reach a collection.
+            setattr(context, 'guest_principals', principals)
+            return True  # OMG, resource code must be rock-solid.
+
         return allowed
 
     def principals_allowed_by_permission(self, context, permission):
@@ -72,6 +79,10 @@ class RouteFactory(object):
 
         # Store service, resource, record and required permission.
         service = utils.current_service(request)
+
+        # Determine if current request hits a collection.
+        # XXX: this will fail if ViewSet.service_name is customized.
+        self.on_collection = service.name.endswith('-collection')
 
         is_on_resource = (service is not None and
                           hasattr(service, 'viewset') and
@@ -124,3 +135,8 @@ class RouteFactory(object):
 def get_object_id(object_uri):
     # Remove potential version prefix in URI.
     return re.sub(r'^(/v\d+)?', '', six.text_type(object_uri))
+
+
+def extract_object_id(object_id):
+    # XXX: Help needed: use something like route.matchdict.get('id').
+    return object_id.split('/')[-1]
--- a/cliquet/authorization.py
+++ b/cliquet/authorization.py
@@ -138,10 +138,17 @@ class RouteFactory(object):
             object_id_match='*%s*' % object_id)
 
     def fetch_shared_records(self, principals):
-        self.shared_ids = self.get_shared_ids(principals=principals)
+        shared_ids = self.get_shared_ids(principals=principals)
+        self.shared_ids = [extract_object_id(shared_id)
+                           for shared_id in shared_ids]
         return self.shared_ids
 
 
 def get_object_id(object_uri):
     # Remove potential version prefix in URI.
     return re.sub(r'^(/v\d+)?', '', six.text_type(object_uri))
+
+
+def extract_object_id(object_id):
+    # XXX: Help needed: use something like route.matchdict.get('id').
+    return object_id.split('/')[-1]
--- a/cliquet/authorization.py
+++ b/cliquet/authorization.py
@@ -51,12 +51,14 @@ class AuthorizationPolicy(object):
                 principals,
                 get_bound_permissions=self.get_bound_permissions)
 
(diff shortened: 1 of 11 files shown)


cliquet/authorization.py
```diff
@@ -67,6 +74,10 @@ def __init__(self, request):
         # Store service, resource, record and required permission.
         service = utils.current_service(request)
 
+        # Determine if current request hits a collection.
+        # XXX: this will fail if ViewSet.service_name is customized.
+        self.on_collection = service.name.endswith('-collection')
```
ametaireau: We could attach metadata to the service then.

cliquet/permission/redis.py
```diff
@@ -79,6 +79,23 @@ def object_permission_principals(self, object_id, permission):
...
     @wrap_redis_error
+    def principals_accessible_objects(self, principals, permission,
+                                      object_id_match=None):
+        if object_id_match is None:
+            object_id_match = '*'
+        key_pattern = 'permission:%s:%s' % (object_id_match, permission)
+
+        # XXX: not optimal at all.
+        objects = set([])
+        _, keys = self._client.scan(match=key_pattern)
```
Natim: We should scan until the end.

cliquet/permission/redis.py
```diff
@@ -79,6 +79,23 @@ def object_permission_principals(self, object_id, permission):
         return self._decode_set(members)
 
     @wrap_redis_error
+    def principals_accessible_objects(self, principals, permission,
```
Natim: We should also pass the ``get_bound_permissions=None`` callback to get the permission inheritance.

cliquet/authorization.py
```diff
@@ -45,6 +45,13 @@ def permits(self, context, principals, permission):
                 principals,
                 get_bound_permissions=self.get_bound_permissions)
 
+        is_list_operation = (context.on_collection and
+                             'create' not in permission)
+        if not allowed and is_list_operation:
+            # Guest trying to reach a collection.
+            setattr(context, 'guest_principals', principals)
```
tarekziade: why not initializing `context.guest_principals` with an empty list instead of adding an attribute on the fly ? doing a `setattr` is making your life harder to read that list for no obvious reason I can see

cliquet/permission/memory.py
```diff
@@ -70,6 +72,21 @@ def object_permission_principals(self, object_id, permission):
...
+    def principals_accessible_objects(self, principals, permission,
+                                      object_id_match=None):
+        if object_id_match is None:
+            object_id_match = re.compile('.*')
+        else:
+            object_id_match = re.compile(object_id_match.replace('*', '.*'))
+        objects = set([])
+        for key, value in self._store.items():
+            if key.endswith(permission):
+                if len(set(principals) & value) > 0:
```
tarekziade: the `set(principals)` call can be done before the `for` loop

cliquet/permission/memory.py
```diff
@@ -70,6 +72,21 @@ def object_permission_principals(self, object_id, permission):
         members = self._store.get(permission_key, set([]))
         return members
 
+    def principals_accessible_objects(self, principals, permission,
+                                      object_id_match=None):
+        if object_id_match is None:
+            object_id_match = re.compile('.*')
+        else:
+            object_id_match = re.compile(object_id_match.replace('*', '.*'))
+        objects = set([])
```
tarekziade: if you want to work with an empty set, just call `set()`

cliquet/permission/redis.py
```diff
@@ -79,6 +79,23 @@ def object_permission_principals(self, object_id, permission):
...
+                                      object_id_match=None):
+        if object_id_match is None:
+            object_id_match = '*'
+        key_pattern = 'permission:%s:%s' % (object_id_match, permission)
+
+        # XXX: not optimal at all.
+        objects = set([])
```
tarekziade: reading through the code it seems that you are always working with sets and convert lists as sets. So, can't principals be a set in the first place ?

Natim: Unfortunately principals is returned by pyramid as a list :(

cliquet/permission/redis.py
Natim: Fixed in #423

cliquet/permission/redis.py
Natim: Actually we don't need to because we already checked the other permissions before.

Natim:
The diff is unreadable, please use https://github.com/mozilla-services/cliquet/compare/master...354-filter-collection to see it correctly. I am afraid github doesn't handle psykorebase correctly just yet.

Natim:
We need to add functional tests.

leplatrem:
Thanks for bringing this further ! 

I quickly overlooked, but it seems ok to be merged, isn't it ?

ametaireau:
YES!

Natim:
Oh yeah :8ball:

PR #361, opened 2015-07-06
Refactor registration of protected resources (fixes #348, fixes #322)

### LINKED ISSUE #322: Add a ProtectedViewSet class
Currently, the default `ViewSet` class has some behaviour tied to the `ProtectedResource` class. We shouldn't have anything related to protected resources in the default viewset since it's a concept which isn't known to the viewset.

Instead, we should define a new `ProtectedViewSet` which handles this and modify the  `register_resource` method to accept a `protected` argument.

Also, I belive this will be the right timing to split viewsets out of the `resource.py` file.

### LINKED ISSUE #348: Error in authorization if no root factory is defined
```
  File "/home/mathieu/Code/Mozilla/readinglist/.venv/local/lib/python2.7/site-packages/pyramid/config/views.py", line 241, in _permitted
    permission)
  File "/home/mathieu/Code/Mozilla/cliquet/cliquet/authorization.py", line 27, in permits
    if context.prefixed_userid:
AttributeError: DefaultRootFactory instance has no attribute 'prefixed_userid'
```

If the permission system is not used (e.g. readinglist), then it's a pity to query the permission backend. 

Suggestions:
* Add default value for root factory kwarg to ``cliquet.authorization.RouteFactory`` to ``@register`` decorator
* Document clearly that ``cliquet.articles_read_principals = system.Authenticated`` can be used to bypass the permissions backend
* Or otherwise, add a simple fallback like this one in authorization policy:

```python
    def permits(self, context, principals, permission):
        # If no root factory, just required authentication.
        if not isinstance(context, RouteFactory):
            return Authenticated in principals
```

leplatrem:
r? @Natim && @ametaireau

### DIFF
--- a/cliquet/resource.py
+++ b/cliquet/resource.py
@@ -40,7 +40,6 @@ class ViewSet(object):
 
     readonly_methods = ('GET',)
 
-    factory = authorization.RouteFactory
     service_arguments = {
         'description': 'Collection of {resource_name}',
         'cors_origins': ('*',),
@@ -100,14 +99,8 @@ class ViewSet(object):
             # Simply validate that posted body is a mapping.
             return colander.MappingSchema(unknown='preserve')
 
-        # XXX: https://github.com/mozilla-services/cliquet/issues/322
-        resource_permissions = getattr(resource, 'permissions', tuple())
-
         class RecordPayload(colander.MappingSchema):
             data = resource.mapping
-            permissions = PermissionsSchema(
-                missing=colander.drop,
-                permissions=resource_permissions)
 
             def schema_type(self, **kw):
                 return colander.Mapping(unknown='raise')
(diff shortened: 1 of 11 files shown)


Natim:
Looks good to me. r+

ametaireau:
I'm unclear about what happens if we define a resource without any factory attached.

The Authorization policy [seems to require a context](https://github.com/mozilla-services/cliquet/blob/master/cliquet/authorization.py#L23) and as such it doesn't really make sense to be able to set the factory to `None`.

Another approach would be to:
- Only check the context if there a permission required;
- Have a separate `ViewSet` subclass for protected resources, which sets the factory to what we're waiting for.

leplatrem:
> doesn't really make sense to be able to set the factory to None.

It only makes sense if another AuthzPolicy is used indeed.

> Have a separate ViewSet subclass for protected resources

This makes a lot of sense, and would like to see it as well. But it requires a lot more efforts on several parts of the code. I wonder if we should do it in the scope of this bug fix...

I'll look at your suggestion and see if we can get rid of those configuration lines to specify principals...

ametaireau:
Great, I prefer it much more like that, thanks!

ametaireau:
(r+)

cliquet/resource.py
```diff
@@ -252,6 +271,9 @@ def callback(context, name, ob):
 class BaseResource(object):
     """Base resource class providing every endpoint."""
 
+    default_viewset = ViewSet
```
ametaireau: neat.

PR #363, opened 2015-07-07
Include Git SHA in the /v1/ endpoint

Natim:
Imported from https://github.com/mozilla-services/readinglist/issues/188

### DIFF
--- a/cliquet/views/hello.py
+++ b/cliquet/views/hello.py
@@ -21,8 +21,11 @@ def get_hello(request):
     if eos:
         data['eos'] = eos
 
-    if git and git.revision:
+    try:
         data['commit'] = git.revision
+    except TypeError:
+        # In case we are not running from a git repository.
+        pass
 
     public_settings = request.registry.public_settings
     data['settings'] = {k: settings[k] for k in public_settings}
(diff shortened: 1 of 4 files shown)


ametaireau:
From [their documentation](https://github.com/klen/dealer#id24), if you actually include their contrib then the git revision is available at the request level (`request.registry.dealer.revision`):

```
config.include('dealer.contrib.pyramid')
```

ametaireau:
Actually, nevermind. What you propose is clearer / cleaner in my opinion.

Natim:
> Actually, nevermind. What you propose is clearer / cleaner in my opinion.

Yes I have seen their helper and I went to the same conclusion.

cliquet/views/hello.py
```diff
@@ -20,6 +21,9 @@ def get_hello(request):
     if eos:
         data['eos'] = eos
 
+    if git and git.revision:
+        data['commit'] = git.revision
```
leplatrem: Could you please add a note the endpoints docs ?

Especially, what happens if installing from pip ?

Natim: ``TypeError: fatal: Not a git repository (or any parent up to mount point /home)``

PR #368, opened 2015-07-08
The permissions backend doesn't handles empty inheritance permission set

Natim:
See https://github.com/mozilla-services/kinto/issues/128

### DIFF
--- a/cliquet/permission/postgresql/__init__.py
+++ b/cliquet/permission/postgresql/__init__.py
@@ -146,19 +146,21 @@ class PostgreSQL(PostgreSQLClient, PermissionBase):
         else:
             perms = get_bound_permissions(object_id, permission)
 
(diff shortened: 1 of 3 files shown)


cliquet/permission/postgresql/__init__.py
```diff
@@ -146,19 +146,21 @@ def object_permission_authorized_principals(self, object_id, permission,
...
-        )
-        SELECT principal
-          FROM required_perms JOIN access_control_entries
-            ON (object_id = column1 AND permission = column2);
-        """ % perms_values
-        with self.connect() as cursor:
-            cursor.execute(query)
-            results = cursor.fetchall()
-        return set([r['principal'] for r in results])
+        if perms:
```
leplatrem: nit: instead of indenting the whole block, prefer ``if not perms: return set()``

PR #372, opened 2015-07-10
Fix missing last_modified field on PATCH when no field was changed  (fixes #371)

leplatrem:
r? @Natim

PR #376, opened 2015-07-10
Fix lost querystring during version redirection (fixes #364)

### LINKED ISSUE #364: On version redirect, the 307 doesn't honour the querystring.
In the ``Location`` header of our 307, we don't keep the querystring that was given.

leplatrem:
A small fix to check if TravisCI fails like for #372


Natim:
r+

PR #377, opened 2015-07-10
Add StatsD counters for the permission backend

leplatrem:


Natim:
Refs: https://github.com/mozilla-services/kinto/pull/134

I am wondering if this should not be part of Cliquet as well.

Natim:
r+

PR #378, opened 2015-07-10
Fix mysterious breaking tests

leplatrem:


leplatrem:
r? @Natim

I could reproduce locally using a fresh venv. Did not compare deps versions though.

PR #388, opened 2015-07-15
Force JSON error handler on cliquet.Service

leplatrem:
Instead of having to specify it everywhere, set it on
the Service subclass in cliquet module.

@Natim r?


Natim:
Yes why not :) r+

ametaireau:
:-)

PR #389, opened 2015-07-15
Remove generic storage from Cliquet docs

leplatrem:
There seems to be a general confusion between Cliquet and Kinto.

Start by removing Kinto from Cliquet rationale. We'll move the
diagrams to Kinto docs instead.

@Natim @n1k0 @tarekziade r?


Natim:
Works for me.

ametaireau:
Yes, that's far better that way. We should never talk about kinto in cliquet, and probably the same applies for kinto: we shouldn't talk about cliquet in our explanations, nor point to the cliquet API docs for explanations, as it will get people lost.

ametaireau:
Where the graphs added back in kinto? I don't see them there!

PR #319, opened 2015-07-15
Expose the authenticated_userid in an endpoint

Natim:
Clients need to know what's their userid is, in order to set and manage permission.

We should expose that in either an extra endpoint or the root one.
We may use the same endpoint as #318 in case we decide to create one.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -6,10 +6,15 @@ This document describes changes between each past release.
 2.4 (unreleased)
 ----------------
 
+**New features**
+
+- Provide the userid when calling the hello page with an Authorization
+  header. (#319)
+
 **Bug fixes**
 
-- Fix crash on hello view when application is not deployed from Git repository
-  (fixes #382)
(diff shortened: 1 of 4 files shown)


leplatrem:
I remember that @ametaireau did not agree with that, since the hello endpoint does not require authentication...
On my side, I don't have any strong opinion on this, and would r+ (with documentation and changelog updated)

Natim:
Note that with this PR the hello endpoint doesn't require authentication.

Natim:
I will merge for now and ask @ametaireau when he comes back and we can change the behaviour before the next release if necessary.

The reason why I think this is a good and acceptable solution:

 - We can retrieve the userid on the / endpoint as well as the settings (could be done in one request)
 - We don't force people to access the / endpoint with an Authorization header

n1k0:
Lagguy r+

leplatrem:
I don't understand why we were in a hurry for this...

Natim:
We are not.

PR #390, opened 2015-07-15
Expose Content-Length when using CORS.

Natim:
r? @n1k0


n1k0:
r=me conditional to green builds :)

PR #328, opened 2015-07-15
With Python 3 Header value should be string

Natim:
AssertionError: Header value b'30' is not a string in ('Retry-After', b'30')

### DIFF
--- a/cliquet/tests/test_utils.py
+++ b/cliquet/tests/test_utils.py
@@ -109,28 +109,28 @@ class CurrentServiceTest(unittest.TestCase):
 
 class EncodeHeaderTest(unittest.TestCase):
 
-    def test_giving_string_return_a_string(self):
+    def test_returns_a_string_if_passed_a_string(self):
         entry = str('Toto')
         value = encode_header(entry)
         self.assertEqual(entry, value)
         self.assertEqual(type(value), str)
 
-    def test_giving_bytes_return_a_string(self):
+    def test_returns_a_string_if_passed_bytes(self):
         entry = 'Toto'.encode('utf-8')
         value = encode_header(entry)
         self.assertEqual(type(value), str)
 
-    def test_giving_bytes_and_encoding_return_a_string(self):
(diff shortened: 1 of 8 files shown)


cliquet/utils.py
```diff
@@ -208,3 +209,23 @@ def build_response(response, request):
...
     return dict_obj
+
+
+def encode_header(value, encoding='utf-8'):
+    """Make sure the value is of type ``str`` in both PY2 and PY3."""
+    value_type = type(value)
+    if value_type != str:
+        # Test for Python3
+        if value_type == six.binary_type:  # pragma: no cover
+            value = value.decode(encoding)  # pragma: no cover
```
leplatrem: both pragmas are required ? (I would have thought that only the one for ``if`` is necessary)

cliquet/utils.py
```diff
@@ -208,3 +209,23 @@ def build_response(response, request):
...
+        # Test for Python2
+        elif value_type == six.text_type:  # pragma: no cover
+            value = value.encode(encoding)  # pragma: no cover
+    return value
+
+
+def decode_header(value, encoding='utf-8'):
+    """Makes sure the header is an unicode string."""
```
leplatrem: nit: Makes vs Make ?

cliquet/utils.py
Natim: I will try :)

cliquet/tests/test_utils.py
```diff
@@ -101,3 +105,50 @@ def test_current_service_returns_none_for_unexisting_patterns(self):
         request.registry.cornice_services = {}
 
         self.assertEqual(current_service(request), None)
+
+
+class EncodeHeaderTest(unittest.TestCase):
+
+    def test_giving_string_return_a_string(self):
```
leplatrem: nit: in other test cases, the formulation is something like ``test_returns_blah_if_foo``

leplatrem:
r+

PR #392, opened 2015-07-16
Added an example in the polling sections

tarekziade:
Plus a few typo fixes

### DIFF
--- a/docs/api/resource.rst
+++ b/docs/api/resource.rst
@@ -139,7 +139,7 @@ Polling for changes
 
 The ``_since`` parameter is provided as an alias for ``gt_last_modified``.
 
-* ``/collection?_since=1437034064``
+* ``/collection?_since=1437035923844``
 
 When filtering on ``last_modified`` every deleted records will appear in the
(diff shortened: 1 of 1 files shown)


docs/api/resource.rst
```diff
@@ -139,14 +139,15 @@ Polling for changes
 
 The ``_since`` parameter is provided as an alias for ``gt_last_modified``.
 
-* ``/collection?_since=1234567890123``
+* ``/collection?_since=1437034064``
```
Natim: This timestamp is not long enough. Should looks like: ``1437035923844``

docs/api/resource.rst
```diff
@@ -154,6 +155,52 @@ response is returned.
    ``lt_last_modified`` (*strictly inferior*).
 
 
+**Request**:
+
+.. code-block:: http
+
+    GET /articles?_since=1437034064 HTTP/1.1
```
Natim: 1437035923844

leplatrem:
Thanks !

PR #393, opened 2015-07-17
Sources of slides for talk at Python Meetup Barcelona

leplatrem:


Natim:
I think we should not merge this into master but into gh-pages.

PR #394, opened 2015-07-17
Change the reference title.

Natim:
I'd like to change the Reference title that doesn't help too much.

Also I'm not sure how to name it.

Since we have API endpoint, I was thinking about
 - Advanced documentation
 - Python API
 - Cliquet guts

### DIFF
--- a/docs/api/index.rst
+++ b/docs/api/index.rst
@@ -1,6 +1,6 @@
 .. _api-endpoints:
 
-API Endpoints
+HTTP Protocol
 #############
 
 .. toctree::
--- a/docs/reference/index.rst
+++ b/docs/reference/index.rst
@@ -1,6 +1,6 @@
 .. _reference:
 
-Reference
+Internals
 #########
 
 .. toctree::


leplatrem:
+1

I propose we also rename *API* to *Protocol*

Natim:
Updated.

leplatrem:
Oh! You changed to guts ? Do you have an example of a library using «guts» instead of *reference* ? 

Also, we mentioned the ``/api/`` in many places (like in Kinto docs). Should we provide a redirection using some kind of sphinx plugin ?

Natim:
> Do you have an example of a library using «guts» instead of reference ? 

Nor guts nor reference actually, I can revert to reference if you'd like.

> Also, we mentioned the /api/ in many places (like in Kinto docs). Should we provide a redirection using some kind of sphinx plugin ?

We can also keep the filename and just change the TOC what do you think?

tarekziade:
Guts sounds like frenglish. Internals ?

Also if protocol refers to the HTTP protocol, I'd be explicit

Natim:
I like ``Internals``

leplatrem:
+1 for *Internals* and *HTTP Protocol*

ametaireau:
> +1 for Internals and HTTP Protocol

For me as well :)

Natim:
Updated.

PR #391, opened 2015-07-17
Rename _to parameter to _before

Natim:
Deprecate ``_to`` and add an alias ``_before``

### DIFF
--- a/cliquet/errors.py
+++ b/cliquet/errors.py
@@ -168,7 +168,7 @@ def raise_invalid(request, location='body', name=None, description=None,
     raise response
 
 
-def send_alert(request, message, url=None, code='soft-eol'):
+def send_alert(request, message=None, url=None, code='soft-eol'):
     """Helper to add an Alert header to the response.
 
     :param code: The type of error 'soft-eol', 'hard-eol'
--- a/cliquet/errors.py
+++ b/cliquet/errors.py
@@ -2,7 +2,7 @@ import six
 from pyramid import httpexceptions
 
 from cliquet import logger
-from cliquet.utils import Enum, json, reapply_cors
+from cliquet.utils import Enum, json, reapply_cors, encode_header
 
 
 ERRORS = Enum(
@@ -166,3 +166,20 @@ def raise_invalid(request, location='body', name=None, description=None,
     request.errors.add(location, name, description, **kwargs)
     response = json_error_handler(request.errors)
     raise response
+
+
+def send_alert(request, message, url=None, code='soft-eol'):
+    """Helper to add an Alert header to the response.
+
+    :param code: The type of error 'soft-eol', 'hard-eol'
+    :param message: The description message.
+    :param url: The URL for more information, default to the documentation url.
+    """
+    if url is None:
+        url = request.registry.settings['cliquet.project_docs']
+
+    request.response.headers['Alert'] = encode_header(json.dumps({
+        'code': code,
+        'message': message,
+        'url': url
+    }))
--- a/cliquet/resource.py
+++ b/cliquet/resource.py
@@ -11,7 +11,7 @@ from cliquet import authorization
 from cliquet import logger
 from cliquet import Service
 from cliquet.collection import Collection
-from cliquet.errors import http_error, raise_invalid, ERRORS
+from cliquet.errors import http_error, raise_invalid, send_alert, ERRORS
 from cliquet.schema import ResourceSchema, PermissionsSchema
 from cliquet.storage import exceptions as storage_exceptions, Filter, Sort
 from cliquet.utils import (
@@ -920,14 +920,9 @@ class BaseResource(object):
                     if param == '_to':
                         message = ('_to is now deprecated, '
                                    'you should use _before instead')
-                        url = ('http://cliquet.rtfd.org/en/latest/api/resource'
+                        url = ('http://cliquet.rtfd.org/en/2.4.0/api/resource'
                                '.html#list-of-available-url-parameters')
-                        self.request.response.headers['Alert'] = encode_header(
-                            json.dumps({
-                                'code': 'soft-eol',
-                                'message': message,
-                                'url': url
-                            }))
+                        send_alert(self.request, message, url)
                     operator = COMPARISON.LT
                 filters.append(
                     Filter(self.collection.modified_field, value, operator)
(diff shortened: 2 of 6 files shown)


Natim:
I was planning to do it like that, but as @tarekziade said, maybe it is easier to just rename it ``_before`` and stop handling ``_to`` what do you guys think?

Natim:
Oh yeah:

```
$ http GET http://localhost:8888/v1/buckets/default/collections/tasks/records?_to=1234 --auth natim:
HTTP/1.1 200 OK
Access-Control-Expose-Headers: Backoff, Retry-After, Alert, Content-Length, Next-Page, Total-Records, Last-Modified, ETag
Alert: {"url":"http:\/\/cliquet.rtfd.org\/en\/latest\/api\/resource.html#list-of-available-url-parameters","message":"_to is now deprecated, you should use _before instead","code":"soft-eol"}
Content-Length: 11
Content-Type: application/json; charset=UTF-8
Date: Fri, 17 Jul 2015 09:30:31 GMT
Etag: "1437125431741"
Last-Modified: Fri, 17 Jul 2015 09:30:31 GMT
Server: waitress
Total-Records: 0

{
    "data": []
}
```

tarekziade:
I think there's a decision to make on whether you want to be backward compatible *now* or once the project has reached a stable milestone. I would not bother at this point to complexify the code base for this kind of thing and make it clear in the doc that the API is still a bit unstable.

Natim:
> make it clear in the doc that the API is still a bit unstable.

I would prefer people to be able to trust our API.

tarekziade:
it's not a problem of trust but clarity. If it's clear in the doc that the http spec has not reached 1.0 and is in a draft state, but changes should be minimal - you are not breaking trust.

You are breaking trust when you are not clear about the API stability and/or its deprecation process and stability.

For instance here, you are changing _to but nowhere in the doc or code you are telling the user when the old version is going away. in 2 versions ? 1 version ? never ?

Natim:
> it's not a problem of trust but clarity. If it's clear in the doc that the http spec has not reached 1.0 and is in a draft state, but changes should be minimal - you are not breaking trust.

Our plan is to use semver for this. So that if we are changing the API, we should upgrade to the next major version.

> here, you are changing _to but nowhere in the doc or code you are telling the user when the old version is going away. in 2 versions ? 1 version ? never ?

I think it is until we upgrade to the next major version.

tarekziade:
> Our plan is to use semver for this. 
Nice!

> So that if we are changing the API, we should upgrade to the next major version.

Sounds perfect.  btw (forgive me if it's already teh case) We should have a separated version for the HTTP API and the actual code implementation.  I guess this will be clearer once we have a separated HTTP spec doc .

> I think it is until we upgrade to the next major version.
it would be nice to have a doc explaining what to do on API changes (e.g. doc impact, deprecation process etc)

leplatrem:
Before we agree on what to do about changing the protocol, I would like to see:
* resource documentation updated
* a ``.. versionchanged :: `` sphinx directive

More generally, I agree with both of you... and I wouldn't have done so much code indeed :)

What I suggest : 
* support transparent backward compatibilty (as long as the hack remains succint)
* remove every deprecation code/hacks when major is increased
* do not use Alert for that (we documented it only for end of life features), just keep it simple

We do not have production code yet, but we have an ecosystem to maintain, and it is energy+time consuming to fix bugs about integration being broken....

cliquet/resource.py
```diff
@@ -899,22 +899,35 @@ def _extract_filters(self, queryparams=None):
...
+                        message = ('_to is now deprecated, '
+                                   'you should use _before instead')
+                        url = ('http://cliquet.rtfd.org/en/latest/api/resource'
+                               '.html#list-of-available-url-parameters')
+                        self.request.response.headers['Alert'] = encode_header(
+                            json.dumps({
+                                'code': 'soft-eol',
+                                'message': message,
+                                'url': url
+                            }))
```
leplatrem: I personally find this a bit overkill.

Natim: It is what I had in mind when we designed this Alert header. Also we could add a helper for this.

``self.soft-eol-alert(message, url)``

My understanding is that we should notify the developer when they are using deprecated features, if we don't they may not noticed that it has been deprecated.

Natim:
> it would be nice to have a doc explaining what to do on API changes (e.g. doc impact, deprecation process etc)

:+1: I have created #396 about that.

Natim:
> do not use Alert for that (we documented it only for end of life features), just keep it simple

For me ``_to`` is an end-of-life feature, did I miss something?

cliquet/resource.py
leplatrem: Ok I agree with you.

Yes, please add a helper (more like ``errors.send_alert(message)`` along with raise_invalid).

I propose that we don't put too much details in the url, and leave something generic like ``settings[documentation_url]``. Because for example, there is no guarantee that the version pointed on readthedocs is the one running on the server.

leplatrem:
> For me _to is an end-of-life feature, did I miss something?

True. My bad.

cliquet/resource.py
Natim: > there is no guarantee that the version pointed on readthedocs is the one running on the server.

True, maybe we should tag the version as we do for Kinto?

>  leave something generic like settings[documentation_url]

The documentation_url is the one from the project, not the one from cliquet, we can be confident that the project doesn't explain the change in details.

docs/api/resource.rst
```diff
@@ -151,10 +151,17 @@ collection was not changed, a ``304 Not Modified`` response is returned.
...
-   The ``_to`` parameter is also available, and is an alias for
+   The ``_before`` parameter is also available, and is an alias for
    ``lt_last_modified`` (*strictly inferior*).
 
 
+.. versionchanged:: 2.4::
+
+   ``_to`` was renamed ``_before``. ``_to`` will still be supported,
+   until the next major version of cliquet, also you should update
+   your code to use ``_before`` from now on.
```
leplatrem: nit:

```
``_to`` was renamed ``_before`` is now deprecated.
It will be supported until the next major version of Cliquet.
```

cliquet/errors.py
```diff
@@ -166,3 +166,20 @@ def raise_invalid(request, location='body', name=None, description=None,
     request.errors.add(location, name, description, **kwargs)
     response = json_error_handler(request.errors)
     raise response
+
+
+def send_alert(request, message, url=None, code='soft-eol'):
```
leplatrem: we may want to use it in the end-of-life in initialization.py

PR #397, opened 2015-07-17
If schema is empty, accept empty body (fixes mozilla-services/kinto#63)

leplatrem:
While I was writing this, I was strongly thinking in changing the default behaviour (schemaless by default).

Since this would be out of scope for this PR, I didn't change anything apart from fixing mozilla-services/kinto#63

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -14,7 +14,7 @@ This document describes changes between each past release.
 **Bug fixes**
 
 - ``data`` is not mandatory in request body if the resource does not define
-  any schema (fixes mozilla-services#63)
+  any schema or if no field is mandatory (fixes mozilla-services/kinto#63)
 
 2.3.1 (2015-07-15)
 ------------------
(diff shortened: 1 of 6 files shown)


CHANGELOG.rst
```diff
@@ -11,6 +11,10 @@ This document describes changes between each past release.
 - Provide the userid when calling the hello page with an Authorization
   header. (#319)
 
+**Bug fixes**
+
+- ``data`` is not mandatory in request body if the resource does not define
+  any schema (fixes mozilla-services#63)
```
leplatrem: typo: missing kinto

cliquet/schema.py
```diff
@@ -61,6 +61,12 @@ def is_readonly(self, field):
         """
         return field in self.get_option("readonly_fields")
 
+    @classproperty
+    def missing(cls):
+        if cls != ResourceSchema:
```
Natim: Actually I would remove this sanity check.
If there is a schema, the missing properties in the schema will take care of raising an error (see /buckets/id/groups/id for instance) We cannot assume that only ResourceSchema accepts empty data.

Natim: I tried and it doesn't check for the missing properties. Is it possible to set {} as the default value for data but test that {} meets the requirements of the defined schema?

leplatrem:
Updated :)

Natim:
Thanks, works for me.

PR #399, opened 2015-07-17
Move the FormattedErrorMixin to support to use it in Kinto.

Natim:
r? @leplatrem


Natim:
Actually this method is not good enough for the test I want to do in kinto.

PR #400, opened 2015-07-17
Cliquet storage should be able to remove tombstones

Natim:
Refs mozilla-services/kinto#136

At the moment, delete_all() remove all objects and create tombstones, we should be able to pass a parameter ``delete_tombstones=True`` to also remove tombstones.

### DIFF
--- a/cliquet/storage/__init__.py
+++ b/cliquet/storage/__init__.py
@@ -154,7 +154,7 @@ class StorageBase(object):
         raise NotImplementedError
 
     def delete(self, collection_id, parent_id, object_id,
-               id_field=DEFAULT_ID_FIELD, with_deleted=True,
+               with_deleted=True, id_field=DEFAULT_ID_FIELD,
                modified_field=DEFAULT_MODIFIED_FIELD,
                deleted_field=DEFAULT_DELETED_FIELD,
                auth=None):
@@ -175,6 +175,7 @@ class StorageBase(object):
         :param str parent_id: the collection parent.
 
         :param str object_id: unique identifier of the object
+        :param bool with_deleted: track deleted record with a tombstone
 
         :returns: the deleted object, with minimal set of attributes.
         :rtype: dict
@@ -182,7 +183,7 @@ class StorageBase(object):
         raise NotImplementedError
 
     def delete_all(self, collection_id, parent_id, filters=None,
-                   id_field=DEFAULT_ID_FIELD, with_deleted=True,
+                   with_deleted=True, id_field=DEFAULT_ID_FIELD,
                    modified_field=DEFAULT_MODIFIED_FIELD,
                    deleted_field=DEFAULT_DELETED_FIELD,
                    auth=None):
@@ -193,6 +194,7 @@ class StorageBase(object):
 
         :param filters: Optionnally filter the objects to delete.
         :type filters: list of :class:`cliquet.storage.Filter`
+        :param bool with_deleted: track deleted records with a tombstone
 
         :returns: the list of deleted objects, with minimal set of attributes.
         :rtype: list of dict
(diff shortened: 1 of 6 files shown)


cliquet/storage/__init__.py
```diff
@@ -199,6 +199,24 @@ def delete_all(self, collection_id, parent_id, filters=None,
         """
         raise NotImplementedError
 
+    def delete_tombstones(self, collection_id, parent_id, before=None,
```
leplatrem: nit: purge ? purge_deleted ? purge_tombstones ?

leplatrem: *(distinguish from delete records)*

Natim: I like ``purge_deleted``

cliquet/storage/__init__.py
```diff
@@ -199,6 +199,24 @@ def delete_all(self, collection_id, parent_id, filters=None,
...
+                          id_field=DEFAULT_ID_FIELD,
+                          modified_field=DEFAULT_MODIFIED_FIELD,
+                          auth=None):
+        """Delete all tombstones for deleted object in this `collection_id`
+        for this `parent_id`.
+
+        :param str collection_id: the collection id.
+        :param str parent_id: the collection parent.
+
+        :param int before: Optionnally elete tombstones older than it.
```
leplatrem: «Optionnal timestamp to limit deletion (?)»

inclusive ? exclusive ?

Natim: exclusive, reason why I wanted to use older than.

cliquet/storage/postgresql/__init__.py
```diff
@@ -448,8 +448,48 @@ def delete_all(self, collection_id, parent_id, filters=None,
...
+            DELETE
+            FROM deleted
+            WHERE parent_id = %%(parent_id)s
+              AND collection_id = %%(collection_id)s
+              %(conditions_filter)s
+            RETURNING id
+        )
+        SELECT COUNT(*) as deleted FROM deleted_records
+        RETURNING deleted;
+        """
```
leplatrem: I think you can just DELETE FROM and use cursor.rowcount to get the number of affected rows

Natim: :+1:

leplatrem:
I have one doubt : I am not sure we should mix the delete of records and purge of tombstones.

I don't have a strong opinion about this, but maybe we could have an option that prevents tombstones to be created on deletion, and keep the purge separated.
```python
storage.delete_all(..., tombstones=False)
storage.purge_tombstones(...)
```
Thoughts ?

Natim:
Actually I am not sure we should let people delete things without tombstone. What is the usecase?

Natim:
Also I agree to keep them separated. I did it because when I asked, you told me a param would be nice, apparently we were not talking about the same feature.

leplatrem:
> told me a param would be nice
...handling two channels of communication at the same time

> delete things without tombstone. What is the usecase?
... in Kinto: delete a collection with 12000 records. You don't want the 12000 tombstones to be created if you purge them just after, no ?

Natim:
> You don't want the 12000 tombstones to be created if you purge them just afte

Good point.

tarekziade:
what is a "tombstone" ? I could not find its definition anywhere.

leplatrem:
> what is a "tombstone" ? I could not find its definition anywhere.

You are right. If we introduce this notion in the code, we should add it to the glossary.

So that clients can obtain the list of deleted records when polling for changes, a tombstone is created when to keep track of the deletion.

cliquet/storage/__init__.py
ametaireau: +1

Natim:
Ok so I have updated the code and tests for the purge_deleted part.
I will now start adding the remove without tombstone part as well.

PR #403, opened 2015-07-20
[WIP] Starting again with issue 34

hiromipaw:
- [x] Make resource classes available in ``cliquet/__init__.py`` to ease usage
- [x] Remove Base* prefixes from almost every cliquet classes


dump.rdb
```diff
@@ -0,0 +1 @@
+REDIS0006�ܳC�Z��V
```
Natim: We should not add this file.

hiromipaw: Yep I will exclude it. :+1:

cliquet/__init__.py
```diff
@@ -16,6 +16,7 @@
 from cliquet.initialization import (  # NOQA
     initialize, initialize_cliquet, install_middlewares)
 
+__all__ = ['utils', 'errors', 'resource']
```
Natim: I am not sure of what you are trying to do here.

cliquet/resource.py
```diff
@@ -12,7 +12,7 @@
 from cliquet import Service
 from cliquet.collection import Collection
 from cliquet.errors import http_error, raise_invalid, ERRORS
-from cliquet.schema import ResourceSchema, PermissionsSchema
+from cliquet.schema import Schema, PermissionsSchema
```
Natim: I think we wanted to remove Base but not Resource in #34

cliquet/__init__.py
Natim: If the aim is to do ``from cliquet import ResourceSchema`` then we should have a list of all the think we want to expose in ``cliquet``

Something like: ``__all__ = ['ResourceSchema', 'PermissionsSchema', 'register', 'Service', 'DEFAULT_SETTINGS', ...]``

Natim: The exhaustive list of everything we want to expose is the following:

- ProtectedResource
- BaseResource (Renamed Resource)
- register
- ProtectedViewSet
- ViewSet

Natim:
Just push new commits to your ``imports_refactoring`` branch to update your PR, no need to create a new PR.

cliquet/resource.py
leplatrem: I was wondering that too... But maybe it's ok to just have ``cliquet.Schema`` and ``cliquet.Resource``.

leplatrem:
Thanks Silvia !

cliquet/resource.py
ametaireau: Having a `Schema` like that doesn't tell what it is.

hiromipaw:
Hey guys, sorry I felt behind with this again...

Natim:
Superseeded by #34

Natim:
Thanks @hiromipaw I am merging this after a rebase.

PR #409, opened 2015-07-30
Remove postgresql-contrib is not mandatory anymore.

Natim:


Natim:
R? @elemoine @leplatrem

elemoine:
If the [uuid-ossp extension](http://www.postgresql.org/docs/current/static/uuid-ossp.html) is not needed anymore then, yes, the postgresql-contrib package is not needed either.

Thanks!

PR #410, opened 2015-07-30
 Remove cliquet protocol docs.

Natim:
Refs https://github.com/mozilla-services/cliquet-protocol/pull/5

r? @ametaireau @leplatrem


ametaireau:
I will build upon this pull request in addition to #413 while waiting on an agreement.

I still personally believe that the best way forward is to avoid adding yet another repository to the mix (to be able to issue pull requests with docs + tests + code), but I'll take the path of least resistance to get the documentation in a better shape.

ametaireau:
We decided to let `cliquet` be the canonical reference for now, and have a `cliquet-protocol` repository which will import from cliquet. We can revert this decision later if we detect this is a problem for implementers.

Closing.

PR #411, opened 2015-07-30
Update default generator regexp.

Natim:
r? @ametaireau

### DIFF
--- a/cliquet/storage/generators.py
+++ b/cliquet/storage/generators.py
@@ -11,7 +11,7 @@ class Generator(object):
     resource level to validate record id in requests paths.
     """
 
-    regexp = r'^[a-zA-Z0-9\-]+$'
+    regexp = r'^[a-zA-Z0-9][a-zA-Z0-9_-]*$'
     """Default record id pattern. Can be changed to comply with custom ids."""
 
     def __init__(self, config=None):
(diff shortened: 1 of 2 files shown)


ametaireau:
so, this isn't tested anywhere?

cliquet/storage/generators.py
```diff
@@ -11,7 +11,7 @@ class Generator(object):
     resource level to validate record id in requests paths.
     """
 
-    regexp = r'^[a-zA-Z0-9\-]+$'
+    regexp = r'^[a-zA-Z0-9][a-zA-Z0-9_-]*$'
```
leplatrem: Are you sure there is no need for blackslash ?

...would definitely require a small test :) Especially to assert the difference between first and other chracacters

Natim: Yes I am sure we don't need backslash because the dash is at the end of the list.
I will add a test.

Natim:
I have added a test.

cliquet/tests/test_storage.py
```diff
@@ -29,6 +29,17 @@ def __call__(self):
 
         self.assertRaises(AssertionError, Dumb)
 
+    def test_default_generator_allow_underscores_dash_alphabet(self):
+        class Dumb(generators.Generator):
+            def __call__(self):
+                return '1234'
```
ametaireau: Not sure to understand why you actually need this. you probably can just test `generators.Generator().match()`

Natim: Actually I tried this and I had a NotImplementedError.

leplatrem:
r+

Issue #373, opened 2015-07-10
Invalid JSON raises 500 on record PATCH


leplatrem:
Following code (I know it's bad):
```python
data = json.dumps(data)
resp = self.session.patch(url, {'data': data}, auth=self.auth)
```

Makes the server fail:
```
12:20:05,742 ERROR [venusian] "PATCH /v1/articles/543212c6-f9bd-4379-af3f-d1ec5ebc408e" ? (? ms) Expected object or value lang=None; exception=Traceback (most recent call last):
  File "/home/mathieu/Code/Mozilla/readinglist/.venv/local/lib/python2.7/site-packages/pyramid/tweens.py", line 21, in excview_tween
    response = handler(request)
  File "/home/mathieu/Code/Mozilla/readinglist/.venv/local/lib/python2.7/site-packages/pyramid/router.py", line 163, in handle_request
    response = view_callable(context, request)
  File "/home/mathieu/Code/Mozilla/readinglist/.venv/local/lib/python2.7/site-packages/pyramid/config/views.py", line 596, in __call__
    return view(context, request)
  File "/home/mathieu/Code/Mozilla/readinglist/.venv/local/lib/python2.7/site-packages/pyramid/config/views.py", line 329, in attr_view
    return view(context, request)
  File "/home/mathieu/Code/Mozilla/readinglist/.venv/local/lib/python2.7/site-packages/pyramid/config/views.py", line 305, in predicate_wrapper
    return view(context, request)
  File "/home/mathieu/Code/Mozilla/readinglist/.venv/local/lib/python2.7/site-packages/pyramid/config/views.py", line 245, in _secured_view
    return view(context, request)
  File "/home/mathieu/Code/Mozilla/readinglist/.venv/local/lib/python2.7/site-packages/pyramid/config/views.py", line 355, in rendered_view
    result = view(context, request)
  File "/home/mathieu/Code/Mozilla/readinglist/.venv/local/lib/python2.7/site-packages/pyramid/config/views.py", line 501, in _requestonly_view
    response = view(request)
  File "/home/mathieu/Code/Mozilla/readinglist/.venv/local/lib/python2.7/site-packages/cornice/service.py", line 567, in wrapper
    response = view_()
  File "/home/mathieu/Code/Mozilla/cliquet/cliquet/resource.py", line 567, in patch
    changes = self.request.json.get('data', {})  # May patch only perms.
[...]

PR #412, opened 2015-08-03
Fix issue #373: 500 on invalid json

msoedov:


leplatrem:
Hi @msoedov! 
Sorry for the delay!

Would you mind adding a small test to assert that the bug is (and remain) fixed please ?

Natim:
The current version of Cliquet is already handling that. See: https://github.com/mozilla-services/cliquet/blob/master/cliquet/tests/resource/test_views.py#L433-L452

Feel free to add a test and reopen if needed.

PR #413, opened 2015-08-12
Make the cliquet documentation importable.

ametaireau:
It is now possible to import the docs from another project, for instance in the
Sphinx configuration file, using `cliquet_docs.copy_docs(section,
destination)`.


cliquet_docs
```diff
@@ -0,0 +1 @@
+docs
```
Natim: ?

ametaireau: that's a symbolic link.

leplatrem:
This conflicts with #410.

We have to take a decision :)

ametaireau:
We decided to let `cliquet` be the canonical reference for now, and have a `cliquet-protocol` repository which will import from cliquet. We can revert this decision later if we detect this is a problem for implementers.

PR #415, opened 2015-08-13
Fix CORS header missing in 404 responses for unknown URLs (fixes #414)

### LINKED ISSUE #414: CORS is missing on unknown URL (404)
CORS are present when a 404 occurs from a resource missing record.

However when the URL is unknown, the ``reapply_cors`` method has no effect (since ``utils.current_service()`` returns ``None``)

Note: This was reported here https://github.com/mozilla-services/syncto/issues/5

leplatrem:
@ametaireau r?

### DIFF
--- a/cliquet/utils.py
+++ b/cliquet/utils.py
@@ -27,6 +27,7 @@ except ImportError:  # pragma: no cover
         import psycopg2  # NOQA
 
 from pyramid.request import Request
+from pyramid.settings import aslist
 from cornice import cors
 from colander import null
 
@@ -160,9 +161,10 @@ def reapply_cors(request, response):
         # No existing service is concerned, and Cornice is not implied.
         origin = request.headers.get('Origin')
         if origin:
-            allowed_origins = request.registry.settings['cliquet.cors_origins']
-            origins_match = decode_header(origin) == allowed_origins
-            if allowed_origins == '*' or origins_match:
+            settings = request.registry.settings
+            allowed_origins = set(aslist(settings['cliquet.cors_origins']))
+            required_origins = {'*', decode_header(origin)}
+            if allowed_origins.intersection(required_origins):
                 origin = encode_header(origin)
                 response.headers['Access-Control-Allow-Origin'] = origin
     return response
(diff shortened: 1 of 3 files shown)


cliquet/utils.py
```diff
@@ -156,6 +156,15 @@ def reapply_cors(request, response):
     if service:
         request.info['cors_checked'] = False
         response = cors.ensure_origin(service, request, response)
+    else:
+        # No existing service is concerned, and Cornice is not implied.
+        origin = request.headers.get('Origin')
+        if origin:
+            allowed_origins = request.registry.settings['cliquet.cors_origins']
+            origins_match = decode_header(origin) == allowed_origins
```
Natim: I would add parenthesis to be more readable.

cliquet/utils.py
```diff
@@ -156,6 +156,15 @@ def reapply_cors(request, response):
     if service:
         request.info['cors_checked'] = False
         response = cors.ensure_origin(service, request, response)
+    else:
+        # No existing service is concerned, and Cornice is not implied.
+        origin = request.headers.get('Origin')
+        if origin:
+            allowed_origins = request.registry.settings['cliquet.cors_origins']
```
Natim: I think cors_origins should be a list rather than a value.

cliquet/utils.py
```diff
@@ -156,6 +156,15 @@ def reapply_cors(request, response):
...
+            origins_match = (decode_header(origin) in allowed_origins)
+            if allowed_origins == '*' or origins_match:
```
Natim: foobar = Service(name='foobar', path='/foobar', cors_origins=('*',))

So we should probably have ``origins_match = (decode_header(origin) in allowed_origins or '*' in allowed_origins)``

ametaireau: See https://github.com/mozilla-services/cornice/blob/master/cornice/cors.py#L102 to reproduce the same behaviour.

ametaireau:
r+ wc

Natim:
r+

PR #422, opened 2015-08-21
Collection can now be filtered with multiple values (fixes #39)

### LINKED ISSUE #39: Add syntax to filter with multiple values for a field
For example, something like ``field_in=1,2,a``

    This should be mentionned in protocol repo too

leplatrem:
(took that from #355)

@Natim r?

### DIFF
--- a/cliquet/resource.py
+++ b/cliquet/resource.py
@@ -906,9 +906,9 @@ class BaseResource(object):
 
         filters = []
 
-        for param, value in queryparams.items():
+        for param, paramvalue in queryparams.items():
             param = param.strip()
-            value = native_value(value)
+            value = native_value(paramvalue)
 
             # Ignore specific fields
             if param.startswith('_') and param not in ('_since',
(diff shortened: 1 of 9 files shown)


Natim:
Shall we add the ``not_in`` in this pull-request as well?

cliquet/resource.py
```diff
@@ -955,6 +955,9 @@ def _extract_filters(self, queryparams=None):
                 }
                 raise_invalid(self.request, **error_details)
 
+            if operator is COMPARISON.IN:
+                value = tuple([native_value(v) for v in paramvalue.split(',')])
```
Natim: ``tuple`` or ``set`` here?

cliquet/storage/memory.py
```diff
@@ -94,12 +94,19 @@ def apply_filters(self, records, filters):
...
 
         for record in records:
-            matches = [operators[f.operator](record.get(f.field), f.value)
-                       for f in filters]
-            if all(matches):
+            matches = True
+            for f in filters:
+                left = record.get(f.field)
+                right = f.value
+                if f.operator == COMPARISON.IN:
```
Natim: I don't like this too much I would rather create the ``COMPARISON.IN: lambda x,y: operator.contains(y, x)``

Natim:
LGTM.

Natim:
> Shall we add the not_in in this pull-request as well?

I will create another one to be able to fix: https://github.com/mozilla-services/readinglist/issues/68

PR #423, opened 2015-08-21
Obtain object ids from permission backend

leplatrem:
Took that from #355 

This deserves some attention regarding Redis key scanning.

I thought it would be easier if contributed as a separate pull-request. @ametaireau @Natim please take up the torch at least for this part of the big project started in #355 :)

### DIFF
--- a/cliquet/permission/postgresql/__init__.py
+++ b/cliquet/permission/postgresql/__init__.py
@@ -163,6 +163,30 @@ class PostgreSQL(PostgreSQLClient, PermissionBase):
             results = cursor.fetchall()
         return set([r['principal'] for r in results])
 
+    def principals_accessible_objects(self, principals, permission,
+                                      object_id_match=None):
+        if object_id_match is None:
+            object_id_condition = 'true'
+        else:
+            object_id_match = object_id_match.replace('*', '%%')
+            object_id_condition = "object_id LIKE '%s'" % object_id_match
+
+        principals_values = ','.join(["('%s')" % p for p in principals])
+        query = """
+        WITH user_principals AS (
+          VALUES %s
+        )
+        SELECT object_id
+          FROM user_principals JOIN access_control_entries
+            ON (principal = column1)
+         WHERE permission = %%(permission)s
+           AND %s;
+        """ % (principals_values, object_id_condition)
+        with self.connect() as cursor:
+            cursor.execute(query, dict(permission=permission))
+            results = cursor.fetchall()
+        return set([r['object_id'] for r in results])
+
     def check_permission(self, object_id, permission, principals,
                          get_bound_permissions=None):
         if get_bound_permissions is None:
--- a/cliquet/permission/postgresql/__init__.py
+++ b/cliquet/permission/postgresql/__init__.py
@@ -165,11 +165,13 @@ class PostgreSQL(PostgreSQLClient, PermissionBase):
 
     def principals_accessible_objects(self, principals, permission,
                                       object_id_match=None):
+        placeholders = {'permission': permission}
         if object_id_match is None:
             object_id_condition = 'true'
         else:
             object_id_match = object_id_match.replace('*', '%%')
-            object_id_condition = "object_id LIKE '%s'" % object_id_match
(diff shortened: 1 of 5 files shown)


cliquet/permission/redis.py
```diff
@@ -79,6 +79,23 @@ def object_permission_principals(self, object_id, permission):
...
+                                      object_id_match=None):
+        if object_id_match is None:
+            object_id_match = '*'
+        key_pattern = 'permission:%s:%s' % (object_id_match, permission)
+
+        # XXX: not optimal at all.
+        objects = set([])
+        _, keys = self._client.scan(match=key_pattern)
+        for key in keys:
+            authorized = self._decode_set(self._client.smembers(key))
```
Natim: We should use SUNION here http://redis.io/commands/sunionstore

cliquet/permission/redis.py
```diff
@@ -79,6 +79,23 @@ def object_permission_principals(self, object_id, permission):
...
     @wrap_redis_error
+    def principals_accessible_objects(self, principals, permission,
+                                      object_id_match=None):
+        if object_id_match is None:
+            object_id_match = '*'
+        key_pattern = 'permission:%s:%s' % (object_id_match, permission)
+
+        # XXX: not optimal at all.
+        objects = set([])
+        _, keys = self._client.scan(match=key_pattern)
```
Natim: This doesn't work because scan uses pagination and you don't navigate through pages here.

cliquet/permission/redis.py
Natim: Well SUNION will not let us findout the list of available records...

Natim:
It looks good to me. @ametaireau would you like to have a second look?

Natim:
Oh actually it is missing the get_bound_permissions callback.

cliquet/permission/postgresql/__init__.py
```diff
@@ -163,6 +163,30 @@ def object_permission_authorized_principals(self, object_id, permission,
             results = cursor.fetchall()
         return set([r['principal'] for r in results])
 
+    def principals_accessible_objects(self, principals, permission,
+                                      object_id_match=None):
+        if object_id_match is None:
+            object_id_condition = 'true'
+        else:
+            object_id_match = object_id_match.replace('*', '%%')
+            object_id_condition = "object_id LIKE '%s'" % object_id_match
```
ametaireau: In order to avoid SQL injections, we usually prefer to build the SQL apart from the values we will pass during the query. and then execute the request with `cursor.execute(query, placeholders)`.

I believe it would be saner to actually append text to the query and to the placeholders rather than building everything like this.

cliquet/permission/redis.py
```diff
@@ -79,6 +79,23 @@ def object_permission_principals(self, object_id, permission):
...
 
     @wrap_redis_error
+    def principals_accessible_objects(self, principals, permission,
+                                      object_id_match=None):
+        if object_id_match is None:
+            object_id_match = '*'
+        key_pattern = 'permission:%s:%s' % (object_id_match, permission)
+
+        objects = set([])
+        keys = self._client.scan_iter(match=key_pattern)
```
ametaireau: Neat. I like how this is exposed by the redis client.

Natim: Yes much better than: https://github.com/mozilla-services/cliquet/commit/3a430453dbc44f7293078e9bd651c834024050f2

cliquet/permission/postgresql/__init__.py
Natim: Oh good catch.

Natim: Oh actually the %% are escaped if we put them in a placeholder. Also the risk for SQL injection here is inexistant because this parameter is not coming from a user input.

Natim: Well let me try one more thing.

Natim: Fixed :)

ametaireau: Well, we never now from where the input comes from. And in any case, trusting inputs is a bad idea ™

PR #426, opened 2015-08-26
Rename the docs folder to cliquet_docs.

ametaireau:
This is useful for setuptools, otherwise the `python setup.py egg_info` command
is broken, because it expects cliquet_docs to be a folder, where it was
a symbolic link.

r? @Natim


Natim:
r+

PR #428, opened 2015-08-26
Document the release process in the cliquet documentation

ametaireau:
@Natim r?

### DIFF
--- a/cliquet_docs/contributing.rst
+++ b/cliquet_docs/contributing.rst
@@ -147,17 +147,13 @@ Step 2
 ------
 
 Once the pull request is validated, you can merge it and do a release. We are
-using `twine <https://pypi.python.org/pypi/twine>`_ to upload to PyPI and
-`wheel <https://pypi.python.org/pypi/wheel>`_ as a packaging format.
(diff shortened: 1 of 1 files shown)


cliquet_docs/contributing.rst
```diff
@@ -112,3 +112,35 @@ on ``#storage`` on ``irc.mozilla.org``.
...
   channel on the mozilla IRC network.
+
+
+How to release
+==============
+
+In order to prepare a new release, we are following the following steps:
+
+1. Update the CHANGELOG with the changes since the last release;
+2. Update the version in setup.py;
```
Natim: We are using pre-release from zest.releaser for that

Natim: Here are the steps we are following for Kinto. https://github.com/Kinto/kinto/pull/137

ametaireau: I personally don't because it doesn't fit well with our release process in my opinion. I will add a note about it.

Natim: I like it to be clear and it prevent us for doing mistakes (forgetting the Back to development commit for instance)

PR #432, opened 2015-08-28
Let people exclude values with filters. (refs mozilla-services/#68).

Natim:
@leplatrem r?

### DIFF
--- a/cliquet/resource.py
+++ b/cliquet/resource.py
@@ -941,7 +941,7 @@ class BaseResource(object):
                 )
                 continue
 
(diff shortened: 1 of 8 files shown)


Natim:
@ametaireau ?

cliquet/resource.py
```diff
@@ -958,6 +958,9 @@ def _extract_filters(self, queryparams=None):
             if operator is COMPARISON.IN:
```
ametaireau: `if operator in (COMPARISON.IN, COMPARISON.EXCLUDE)`

ametaireau:
I'm updating the PR.

PR #34, opened 2015-09-04
Refactor the imports

Natim:
Currently, we import everything from `cliquet.resource`, and it might make more sense to actually import from `cliquet` directly.

e.g.
```python
from cliquet import BaseResource, ResourceSchema, register
```


Natim:
@hiromipaw it seems to be a little more complicated than I though, actually we have circular imports all over the place if we try to import ``Resource`` in ``__init__.py`` which is mandatory if we want to define it in ``__all__``

I feel like we should probably just keep it the way it is at the moment.

@leplatrem @hiromipaw @ametaireau any thoughts ?

hiromipaw:
I am ok with leaving it like this for now.

leplatrem:
This is a very high priority.

I feel embarrassed about the efforts Sylvia provided on this though...

hiromipaw:
@leplatrem it's ok I learned a lot and I can always take another issue from the stack :)

leplatrem:
Well, I am truely sorry, I sincerely hope your next experience will be even more positive (ie. «merged») :)

Thanks again! 

Closing :(

PR #437, opened 2015-09-08
PATCH doesn't replace permissions 

Natim:
Refs Kinto/kinto#155


Natim:
r? @leplatrem 

I think we can go for this for now and we should refactor the ``_delete_permissions`` in another PR.

leplatrem:
r+, even though, as you said, this part requires a huge cleaning/refactor.

ametaireau:
ccing @phrawzty since he looked at this in the past and was interested in the outcome.

PR #443, opened 2015-09-09
Fix for value list should be tuple.

Natim:
This fix the issue but we should add a test.

### DIFF
--- a/cliquet/tests/test_storage.py
+++ b/cliquet/tests/test_storage.py
@@ -303,6 +303,15 @@ class BaseTestStorage(object):
                                           **self.storage_kw)
         self.assertEqual(len(records), 2)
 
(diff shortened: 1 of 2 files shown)


Natim:
Here we've got the root of the problem.

leplatrem:
r+

cliquet/tests/test_storage.py
```diff
@@ -303,6 +303,15 @@ def test_get_all_can_filter_with_list_of_values(self):
                                           **self.storage_kw)
         self.assertEqual(len(records), 2)
 
+    def test_get_all_can_filter_with_list_of_values_on_id(self):
```
ametaireau: Thanks!

PR #435, opened 2015-09-09
Allow-Control-Expose-Headers headers is missing on 400 errors.

Natim:
Refs: https://github.com/Kinto/kinto.js/issues/110


ametaireau:
That's an implementation limitation in Cornice. We might want to document it.

This PR looks okay to me. I'm not sure about the location of the import, but it looks good to me otherwise.

Natim:
>  I'm not sure about the location of the import

This is because cliquet import utils which import Service :(

Natim:
Maybe should I put it on line 174?

ametaireau:
Ah okay. Might be a solution then yes.

PR #444, opened 2015-09-09
Plug get_bound_permissions for partial collections

leplatrem:
Currently, partial collections only work with records on which the user has the same specific permission.

With this pull-request we plug the ``get_bound_permission`` in order to obtain the list of accessible object ids from the permission backend.

For Kinto, this mean that if a user has the ``write`` permission on a record, it will be appear in the corresponding collection with a ``GET`` (because in Kinto, write implies read).

* [x] Refactor code in resource and authorization
* [x] Handle get_bound_permission response in Memory permission backend
* [x] Handle in Postgresql
* [x] Handle in Redis

### DIFF
--- a/cliquet/permission/memory.py
+++ b/cliquet/permission/memory.py
@@ -76,16 +76,16 @@ class Memory(PermissionBase):
                                       object_id_match=None,
                                       get_bound_permissions=None):
         principals = set(principals)
+        if object_id_match is None:
+            object_id_match = '*'
 
         if get_bound_permissions is None:
-            if object_id_match is None:
-                object_id_match = '*'
             object_id_match = object_id_match.replace('*', '.*')
             keys = [(re.compile(object_id_match), permission)]
         else:
             keys = get_bound_permissions(object_id_match, permission)
-            keys = [(re.compile(obj_id.replace('*', '.*')), p)
-                    for (obj_id, p) in keys]
+            keys = [(re.compile(obj_id.replace('*', '.*')), p) for (obj_id, p)
+                    in keys if obj_id.endswith(object_id_match)]
 
         objects = set()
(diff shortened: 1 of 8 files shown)


cliquet/permission/memory.py
```diff
@@ -73,19 +73,28 @@ def object_permission_principals(self, object_id, permission):
...
+                                      object_id_match=None,
+                                      get_bound_permissions=None):
         principals = set(principals)
-        if object_id_match is None:
-            object_id_match = re.compile('.*')
+
+        if get_bound_permissions is None:
+            if object_id_match is None:
+                object_id_match = '*'
+            object_id_match = object_id_match.replace('*', '.*')
```
leplatrem: nit: do it once after if/else

cliquet/tests/test_permission.py
```diff
@@ -256,6 +256,29 @@ def test_obtain_object_ids_can_be_filtered_with_pattern(self):
...
+        self.assertEquals(object_ids, set(['/url/a/id/1', '/url/a/id/2']))
+
+    def test_obtain_object_ids_with_get_bound_permissions_without_match(self):
+        self.permission.add_principal_to_ace('/url/a/id/1', 'write', 'user1')
+        self.permission.add_principal_to_ace('/url/a/id/2', 'read', 'user1')
+        self.permission.add_principal_to_ace('/url/b/id/2', 'read', 'user1')
+        object_ids = self.permission.principals_accessible_objects(
+            ['user1'],
+            'read',
+            get_bound_permissions=lambda o, p: [('/url/a/id/1', 'write'),
```
Natim: Should it be ``id/*`` instead of ``id/1`` ?

PR #446, opened 2015-09-10
Add resource cache control (fixes #401)

### LINKED ISSUE #401: Add cache headers
We will use Kinto behind a proxy cache for Go Faster on all GETs.

In order to do this we need to implement a few headers:  Expires & Cache-Control.

We're probably going to use Nginx as a cache server - since it's already our proxy. 

So this link http://wiki.nginx.org/NginxHttpProxyModule#proxy_cache is a good reading to implement the cache headers. 

This page is also cool to understand how to set up nginx as a proxy cache : https://serversforhackers.com/nginx-caching

leplatrem:
@Natim @ametaireau r?

Further work will happen in Kinto.

### DIFF
--- a/cliquet/resource.py
+++ b/cliquet/resource.py
@@ -50,12 +50,14 @@ class ViewSet(object):
 
     default_collection_arguments = {}
     collection_get_arguments = {
-        'cors_headers': ('Next-Page', 'Total-Records', 'Last-Modified', 'ETag')
+        'cors_headers': ('Next-Page', 'Total-Records', 'Last-Modified', 'ETag',
+                         'Cache-Control', 'Expires', 'Pragma')
     }
 
     default_record_arguments = {}
     record_get_arguments = {
-        'cors_headers': ('Last-Modified', 'ETag')
+        'cors_headers': ('Last-Modified', 'ETag',
+                         'Cache-Control', 'Expires', 'Pragma')
     }
 
     def __init__(self, **kwargs):
@@ -362,6 +364,7 @@ class BaseResource(object):
             if filters or sorting are invalid.
         """
         self._add_timestamp_header(self.request.response)
+        self._add_cache_header(self.request.response)
         self._raise_304_if_not_modified()
         self._raise_412_if_modified()
 
@@ -480,6 +483,7 @@ class BaseResource(object):
         record = self._get_record_or_404(self.record_id)
         timestamp = record[self.collection.modified_field]
         self._add_timestamp_header(self.request.response, timestamp=timestamp)
+        self._add_cache_header(self.request.response)
         self._raise_304_if_not_modified(record)
         self._raise_412_if_modified(record)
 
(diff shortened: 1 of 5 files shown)


cliquet/resource.py
```diff
@@ -753,6 +755,15 @@ def _add_timestamp_header(self, response, timestamp=None):
         # Return timestamp as ETag.
         response.headers['ETag'] = encode_header('"%s"' % timestamp)
 
+    def _add_cache_header(self, response):
+        """Add Cache-Control and Expire headers, based a setting for this
```
Natim: based on settings for the current resource.

cliquet/resource.py
```diff
@@ -753,6 +755,15 @@ def _add_timestamp_header(self, response, timestamp=None):
...
+        resource.
+        """
+        key = 'cliquet.%s_cache_expires_seconds' % self.context.resource_name
```
ametaireau: So we always have the same policy for all GETs on the same resource (collection get and resource get). We might want to add the `resource_type` to the configuration otherwise.

Natim:
We should probably add a functional test for that even if we are testing pyramid code.

ametaireau:
unit tests!

leplatrem:
I only wrote unit tests indeed.

I will add an integration test (pyramid via webtest).

Our only hope to get a functional test for this is the kinto load test :)

cliquet/resource.py
leplatrem: I thought of this but gave up with the idea. Do you have a strong argument in favor of distinguishing collection and records ?

Natim: I think we can leave it like that for now, since it matches our use case and then add more precises settings if needed.

For now, I think we should have the same for collection_get and get, also we may want to differenciate between authenticated and unauthenticated (cache for unauthenticated only for instance)

leplatrem:
> differenciate between authenticated and unauthenticated

Would you make that a blocker for this one?

cliquet/resource.py
```diff
@@ -50,12 +50,14 @@ class ViewSet(object):
 
     default_collection_arguments = {}
     collection_get_arguments = {
-        'cors_headers': ('Next-Page', 'Total-Records', 'Last-Modified', 'ETag')
+        'cors_headers': ('Next-Page', 'Total-Records', 'Last-Modified', 'ETag',
+                         'Cache-Control', 'Expires', 'Pragma')
```
Natim: Are you sure that Kinto.js needs to read these headers?

Natim:
> Would you make that a blocker for this one?

No we could open another issue, but if we don't add it we will have problems editing the collection.

cliquet/resource.py
Natim: Actually it is a good think to let them if they do.

Natim:
r+

PR #408, opened 2015-09-11
Contributors list in documentation?

Natim:
There is a hardcoded [list of contributors](https://cliquet.readthedocs.org/en/latest/contributors.html) in the documentation. This list may or may not line up with the [repo contributors graph](https://github.com/mozilla-services/cliquet/graphs/contributors), and in any case, poses problems in terms of maintenance.

I am curious who manages this list, and what the criteria are for inclusion (or not).  I'd also like to open bikeshed on whether this list should even be included at all - perhaps just a like to the contributors graph would be more honest (and sustainable in terms of maint)?  Either that or some sort of automated generation?

### DIFF
--- a/cliquet_docs/contributing.rst
+++ b/cliquet_docs/contributing.rst
@@ -136,6 +136,7 @@ Step 1
 - Update ``CHANGELOG.rst``
 - Update version in ``docs/conf.py``
 - Known good versions of dependencies in ``requirements.txt``
+- Update ``CONTRIBUTORS.rst`` using: ``git shortlog -sne | awk '{$1=""; sub(" ", ""); print}'``
 
 .. code-block:: bash
 
--- a/cliquet_docs/contributing.rst
+++ b/cliquet_docs/contributing.rst
@@ -136,7 +136,7 @@ Step 1
 - Update ``CHANGELOG.rst``
 - Update version in ``docs/conf.py``
 - Known good versions of dependencies in ``requirements.txt``
-- Update ``CONTRIBUTORS.rst`` using: ``git shortlog -sne | awk '{$1=""; sub(" ", ""); print}'``
+- Update ``CONTRIBUTORS.rst`` using: ``git shortlog -sne | awk '{$1=""; sub(" ", ""); print}' | awk -F'<' '!x[$1]++' | awk -F'<' '!x[$2]++'``
 
 .. code-block:: bash
 


phrawzty:
In principle this is what we want; however, consider the following example from Kinto:

```
$ git shortlog -sne | awk '{$1=""; sub(" ", ""); print}'
Rémy HUBSCHER <rhubscher@mozilla.com>
Mathieu Leplatre <mathieu@mozilla.com>
Alexis Métaireau <alexis@notmyidea.org>
Rémy HUBSCHER <hubscher.remy@gmail.com>
Alexis Metaireau <alexis@notmyidea.org>
Mathieu Leplatre <leplatrem@gmail.com>
Mathieu Agopian <mathieu@agopian.info>
Tarek Ziade <tarek@ziade.org>
dan phrawzty <phrawzty+github@gmail.com>
phrawzty <phrawzty+github@gmail.com>
Andy McKay <amckay@mozilla.com>
Fil <fil@rezo.net>
dan <phrawzty+github@gmail.com>
```

As you can see, some names are listed more than one (notably, I have terrible `.git/config` hygeine), which makes for a messy list. Git has a built-in mechanism for dealing with this ([mailmap](https://www.kernel.org/pub/software/scm/git/docs/git-shortlog.html)) but it requires knowledge of the contents of the shortlog ahead of time.

I propose that we process the list and remove duplicates on the fly by harnessing the **awesome power of `awk` and pipes**. :neckbeard: In the following example I pass the shortlog to `awk` (as above), then pipe it through two additional field sorts: one each for *name* and *email address*.

```
$ git shortlog -sne | awk '{$1=""; sub(" ", ""); print}' | awk -F'<' '!x[$1]++' | awk -F'<' '!x[$2]++'
Rémy HUBSCHER <rhubscher@mozilla.com>
Mathieu Leplatre <mathieu@mozilla.com>
Alexis Métaireau <alexis@notmyidea.org>
Mathieu Agopian <mathieu@agopian.info>
Tarek Ziade <tarek@ziade.org>
dan phrawzty <phrawzty+github@gmail.com>
Andy McKay <amckay@mozilla.com>
Fil <fil@rezo.net>
```

Thoughts?

Natim:
I would rather make this manually because your script doesn't always select the right email address.
Or we could read the existing CONTRIBUTORS file and just display the one that needs to be added?

phrawzty:
> I would rather make this manually because your script doesn't always select the right email address.

It's a question of sort order - and, to be fair, the fault lies with the contributors (not the sorting mechanism). :smirk: Anyway, I have no appetite for defending this automation any further, so do as you feel is best.

Natim:
We can also use your script and select the right email address yes, I am going to send a PR to git-extras :+1:

ametaireau:
As we don't have any way to know what's the "correct" email address, I would put this in the documentation, and let the releaser take care of updating accordingly if needed.

leplatrem:
Well, I believe we can merge this.

PR #452, opened 2015-09-15
Fetch record only once from storage

leplatrem:
This allowed me to go from 29 RPS to 35 RPS (+18%)

@ametaireau r?

### DIFF
--- a/cliquet/authorization.py
+++ b/cliquet/authorization.py
@@ -73,6 +73,7 @@ class RouteFactory(object):
     required_permission = None
     allowed_principals = None
     permission_object_id = None
+    current_record = None
     get_shared_ids = None
 
     method_permissions = {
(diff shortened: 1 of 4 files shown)


ametaireau:
We could add a unittest on the authorization policy which check that `current_record` is set correctly and another one on resource which checks that the record used is the one from the context.

Otherwise, looks good to me.

cliquet/authorization.py
```diff
@@ -105,11 +106,12 @@ def __init__(self, request):
...
                 # exists, return "write" if it does, "create" otherwise.
 
-                # If the view exists, call it with the request and catch an
+                # If the view exists, use its collection to catch an
                 # eventual NotFound.
                 resource = service.resource(request=request, context=self)
                 try:
-                    resource.collection.get_record(resource.record_id)
+                    record = resource.collection.get_record(resource.record_id)
+                    current_record = record
```
Natim: can't we just write into current_record directly ?

leplatrem: just a matter of taste+linewrap :(

leplatrem:
Thanks for your feedback. 2 tests added.

PR #454, opened 2015-09-18
Refactoring the Permissions backend

Natim:


Natim:
r? @leplatrem 

I think we might want to remove other methods now that we have those.

cliquet/permission/postgresql/__init__.py
```diff
@@ -236,6 +237,60 @@ def check_permission(self, object_id, permission, principals,
...
+    def replace_object_permissions(self, object_id, permissions):
+        query = """
+        DELETE FROM access_control_entries
+         WHERE object_id = %(object_id)s
+           AND permission in %(permissions)s;"""
+        with self.connect() as cursor:
+            cursor.execute(query, dict(object_id=object_id,
+                                       permissions=tuple(permissions)))
+
+        # XXX: Not good enough
```
Natim: We should try to write this in one postgresql query.

Natim:
LGTM thanks for your help.

cliquet/permission/memory.py
```diff
@@ -108,6 +108,37 @@ def object_permission_authorized_principals(self, object_id, permission,
...
+    def object_permissions(self, object_id, permissions=None):
+        if permissions is None:
+            aces = [k for k in self._store
+                    if k.startswith('permission:%s' % object_id)]
+        else:
+            aces = ['permission:%s:%s' % (object_id, permission)
+                    for permission in permissions]
+        permissions = {}
+        for ace in aces:
+            _, _, permission = ace.split(':')
```
leplatrem: XXX this fails with ``permission:/url/id:record:create``

Natim: You can use ``split(':', 2)``

cliquet/permission/memory.py
```diff
@@ -108,6 +108,37 @@ def object_permission_authorized_principals(self, object_id, permission,
...
+            if permission_key in self._store and len(principals) == 0:
+                del self._store[permission_key]
+            else:
+                self._store[permission_key] = set(principals)
+        return permissions
+
+    def delete_object_permissions(self, *object_id_list):
+        to_delete = []
+        for key in self._store:
+            _, object_id, _ = key.split(':')
```
Natim: ditto

leplatrem:
@Natim

With pypy: ```TypeError: set([u'basicauth:9f2d363f98418b13253d6d7193fc88690302ab0ae21295521f6029dffe9dc3b0']) is not JSON serializable
```

Where do we cast ?

**Comment by Natim** [comment]
We should do what you said and return list in response instead of sets.

File: cliquet/collection.py [hunk 4]
```diff
@@ -233,3 +233,95 @@ def delete_record(self, record):
...
+        # Object permission id.
+        self.get_permission_object_id = None
+        # Current user main principal.
+        self.current_principal = None
+
+    def _allow_write(self, perm_object_id):
+        """Helper to give the ``write`` permission to the current user.
+        """
+        self.permission.add_principal_to_ace(perm_object_id,
+                                             'write',
```
Natim: This looks really Kinto specific to me.

Natim: Also it was done in ProtectedResource before so let's keep it here for now.

cliquet/collection.py
```diff
@@ -233,3 +233,95 @@ def delete_record(self, record):
...
+        The current principal is added to the owner (``write`` permission).
+        """
+        permissions = record.pop(self.permissions_field, {})
+        record = super(ProtectedCollection, self).create_record(
+            record, parent_id, unique_fields)
+
+        record_id = record[self.id_field]
+        perm_object_id = self.get_permission_object_id(record_id)
+        self.permission.replace_object_permissions(perm_object_id, permissions)
+        self._allow_write(perm_object_id)
```
Natim: We should add this with the replace_object_permissions call instead. adding the principal to the permissions dict.

cliquet/collection.py
```diff
@@ -233,3 +233,95 @@ def delete_record(self, record):
...
+        The current principal is added to the owner (``write`` permission).
+        """
+        permissions = record.pop(self.permissions_field, {})
+        record = super(ProtectedCollection, self).update_record(
+            record, parent_id, unique_fields)
+
+        record_id = record[self.id_field]
+        perm_object_id = self.get_permission_object_id(record_id)
+        self.permission.replace_object_permissions(perm_object_id, permissions)
+        self._allow_write(perm_object_id)
```
Natim: ditto, if we use replace, we should not use add_ace after it.

cliquet/collection.py
```diff
@@ -233,3 +233,95 @@ def delete_record(self, record):
...
+    permissions_field = '__permissions__'
+
+    def __init__(self, *args, **kwargs):
+        super(ProtectedCollection, self).__init__(*args, **kwargs)
+        # Permission backend.
+        self.permission = None
+        # Object permission id.
+        self.get_permission_object_id = None
+        # Current user main principal.
+        self.current_principal = None
```
Natim: I don't see how this is supposed to be configured?

cliquet/collection.py
Natim: How does this pass this test: test_permissions_can_be_removed_with_patch ?

Natim:
It looks good to me.

leplatrem:
\o/ thanks !

PR #457, opened 2015-09-22
Fix permission backend replace with empty input

leplatrem:
@Natim r?


Natim:
It makes a lot of sense. Thanks.

PR #458, opened 2015-09-22
Index permissions (ref #354)

leplatrem:
In PR #354, some JOIN were introduced, and no index was hit.
In Kinto tutorial load test, went from ~90 RPS to ~110 RPS.


Natim:
I like that :)

PR #459, opened 2015-09-23
Add request bound data, shared with subrequests.

leplatrem:
This allows to store any arbitrary data in a ``dict`` on the current request,
instead of setting custom attributes and testing their existance.

It is shared with subrequests during batch.

In the short term, this will allow us to reduce storage hits in batch
subrequests on the default bucket in Kinto.

In the long term, this will allow us to share storage connections accross
batch subrequests.

@Natim r?


Natim:
So it is just for subrequest?

PR #460, opened 2015-09-23
Explicitly call to collection create_record on PUT

leplatrem:
While discussing with @phrawzty I was convinced that the collection method called during a «creating» PUT was ``create_record()``. I was very surprised it wasn't.

This fixes it. The change is obvious, but there is a test asserting it of course :)

@phrawzty @Natim r?


Natim:
Yes I realized that when debugging loadtests.

PR #465, opened 2015-09-24
Expose statsd client in the registry.

Natim:
This is needed in order to add specific metrics in Syncto for instance. (https://github.com/mozilla-services/syncto/issues/18)

### DIFF
--- a/cliquet/tests/test_initialization.py
+++ b/cliquet/tests/test_initialization.py
@@ -177,9 +177,23 @@ class StatsDConfigurationTest(unittest.TestCase):
         mocked.assert_not_called()
 
     @mock.patch('cliquet.statsd.Client')
+    def test_statsd_is_set_to_none_if_statsd_url_not_set(self, mocked):
+        self.config.add_settings({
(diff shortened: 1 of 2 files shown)


cliquet/tests/test_initialization.py
```diff
@@ -177,9 +177,23 @@ def test_statsd_isnt_called_if_statsd_url_is_not_set(self, mocked):
...
+            'cliquet.statsd_url': None
+        })
+        initialization.setup_statsd(self.config)
+        self.assertEqual(self.config.registry.statsd, None)
+
+    @mock.patch('cliquet.statsd.Client')
     def test_statsd_is_called_if_statsd_url_is_set(self, mocked):
         initialization.setup_statsd(self.config)
         mocked.assert_called_with('host', 8080, 'cliquet')
+        self.assertEqual(self.config.registry.statsd, mocked.return_value)
```
leplatrem: This assert is not necessary, since there is the next one

leplatrem:
r+wc

PR #466, opened 2015-09-25
Add header to cache CORS preflight requests

leplatrem:
ref https://github.com/Kinto/kinto/issues/185

**Requires** https://github.com/mozilla-services/cornice/pull/338

* [x] Take cors max age from settings
* [x] Test app initialization with custom setting

### DIFF
--- a/cliquet/__init__.py
+++ b/cliquet/__init__.py
@@ -31,6 +31,7 @@ DEFAULT_SETTINGS = {
     'cliquet.cache_pool_size': 10,
     'cliquet.cache_url': '',
     'cliquet.cors_origins': '*',
+    'cliquet.cors_max_age': 3600,
     'cliquet.eos': None,
     'cliquet.eos_message': None,
     'cliquet.eos_url': None,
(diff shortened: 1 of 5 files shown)


leplatrem:
This is ready to be reviewed, but waiting for cornice release.

Should require a cornice git tag rev instead, at least during dev ?

leplatrem:
@ametaireau r?

cliquet/__init__.py
```diff
@@ -130,6 +131,9 @@ def includeme(config):
     Service.cors_origins = tuple(aslist(cors_origins))
     Service.default_cors_headers = ('Backoff', 'Retry-After', 'Alert',
                                     'Content-Length')
+    cors_max_age = settings['cliquet.cors_max_age_seconds']
+    Service.cors_max_age = int(cors_max_age) if cors_max_age else None
```
ametaireau: nit: I prefer to have the "if" at the beginning of the line to ease readability.

cliquet_docs/reference/configuration.rst
```diff
@@ -254,6 +254,20 @@ Basically, this will add both ``Cache-Control: max-age=3600`` and
 ``Expire: <server datetime + 1H>`` response headers to the ``GET`` responses.
 
 
+CORS
+----
+
+By default, CORS headers are cached by clients during 1H (``Access-Control-Max-Age``).
+
+The duration can be set from settings. If set to empty or to 0, the header
```
ametaireau: I'm not actually sure it's true if it's set to `0`, from the code I read in cornice.

leplatrem: You can be confident, I wrote a test :)

cliquet/__init__.py
leplatrem: well this can be discussed in this particular case :) It is a safety check and it allows the main flow to remain concise... Matter of taste?

ametaireau:
r+!

PR #467, opened 2015-09-25
Remove hard coded CORS origins

leplatrem:
Viewset should not override settings value with hard coded value

* [x] Write test that origins can be controlled from settings


leplatrem:
@Natim r?

Natim:
Seems good to me.

PR #468, opened 2015-09-28
Remove duplicate base test (fixes #297)

### LINKED ISSUE #297: Refactorize the use of BaseWebTest
We have three BaseWebTest classes for our tests: `cliquet.tests.support.BaseWebTest`, `cliquet.tests.test_views.BaseWebTest` and `cliquet.tests.resource.BaseWebTest`.

- the first two share some scope and probably should be refactorized together;
- the last one shouldn't have the same name than the other two.

leplatrem:
@ametaireau r?

### DIFF
--- a/cliquet/tests/support.py
+++ b/cliquet/tests/support.py
@@ -74,14 +74,14 @@ class BaseWebTest(object):
 
     def __init__(self, *args, **kwargs):
         super(BaseWebTest, self).__init__(*args, **kwargs)
-        self.app = self._get_test_app()
+        self.app = self.get_test_app()
         self.storage = self.app.app.registry.storage
         self.headers = {
             'Content-Type': 'application/json',
             'Authorization': 'Basic bWF0OjE='
         }
 
-    def _get_test_app(self, settings=None):
+    def get_test_app(self, settings=None):
         app = webtest.TestApp(testapp(self.get_app_settings(settings)))
         app.RequestClass = get_request_class(self.api_prefix)
         return app
--- a/cliquet/tests/support.py
+++ b/cliquet/tests/support.py
@@ -71,11 +71,16 @@ class BaseWebTest(object):
     """
 
     api_prefix = "v0"
+    authorization_policy = 'cliquet.tests.support.AllowAuthorizationPolicy'
+    collection_url = '/mushrooms'
+    principal = USER_PRINCIPAL
 
     def __init__(self, *args, **kwargs):
         super(BaseWebTest, self).__init__(*args, **kwargs)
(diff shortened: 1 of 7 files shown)


cliquet/tests/support.py
```diff
@@ -71,17 +71,22 @@ class BaseWebTest(object):
     """
 
     api_prefix = "v0"
+    authorization_policy = 'cliquet.tests.support.AllowAuthorizationPolicy'
+    collection_url = '/mushrooms'
```
ametaireau: While it's easier to use, I'm not  sure we should put a collection url by default. (explicit vs implicit)

ametaireau:
Flake8 fails because of
```
cliquet/tests/resource/test_views.py:2:1: F401 'webtest' imported but unused
cliquet/tests/resource/test_views.py:4:1: F401 'testing' imported but unused
cliquet/tests/resource/test_views.py:6:1: F401 'cliquet' imported but unused
cliquet/tests/resource/test_views.py:9:1: F401 'USER_PRINCIPAL' imported but unused
cliquet/tests/resource/test_views.py:9:1: F401 'get_request_class' imported but unused
```

ametaireau:
Coverage is missing:

```
cliquet/utils.py                     120      6    95%   168-173
cliquet/views/errors.py               52      1    98%   95
```

ametaireau:
This is very nice to have only one base test to look at, thanks :-) r+wc

cliquet/tests/support.py
leplatrem: I am not shocked by that, since this test case uses the `cliquet.tests.testapp`.

I would need your help because I can't think of a better way. Using constants ? What about `get_item_url()`. Ideas?

ametaireau: Ah yes, testapp exposes mushrooms. Disregard this comment then!

leplatrem:
@ametaireau I had to add this https://github.com/mozilla-services/cliquet/pull/468/files#diff-c6183ea6846eeab46df201a6634d0e2aR92  to get coverage of this : 
https://github.com/mozilla-services/cliquet/blob/master/cliquet/views/errors.py#L94-L95

So this was probably not well tested before :D

ametaireau:
Wow interesting! Coverage rocks ;)

Natim:
Strangely enough, this will probably breaks syncto, see https://github.com/mozilla-services/syncto/blob/master/syncto/views/errors.py#L12

leplatrem:
> Strangely enough, this will probably breaks syncto

Don't you have a way to narrow the context to errors from syncclient ?

PR #469, opened 2015-09-29
Provide current request to backend methods

leplatrem:
This small refactor would pave the way for:
* persisting backend connections on current request (instead of pull/push from queue for each operation performed with one request, e.g batched requests)
* having one transaction per request (with a tween to connect / commit). Particularly interesting for batched requests. (see https://github.com/Kinto/kinto/issues/194) 
* **edit**: have one event per request, with every created/updated/deleted records of a batch

Of course, it may seem ugly to pass the request, but it is advocated [in Pyramid docs](http://pyramid.readthedocs.org/en/1.5-branch/narr/threadlocals.html#why-you-shouldn-t-abuse-thread-locals):
> get_current_request function should never be called because it’s “easier” or “more elegant” to think about calling it **than to pass a request through a series of function calls when creating some API design**. Your application should instead almost certainly pass data derived from the request around rather than relying on being able to call this function to obtain the request in places that actually have no business knowing about it.

We could pass a derived info like `request.bound_data`, but I like the fact to pass the request object. 

Especially because it gives a lot of freedom in possible backend implementation (*imagine a box.com storage backend where original request headers are manipulated before being sent to third party service...*)

Before going on with that, what do you think ?


ametaireau:
While I'm not sold to the idea to pass the request to all of the backends, I see where you're coming from and I canot think about any other way to do solve this in a cleaner way.

Natim:
This is something we will need if we want to plug the Cloud Storage backend at some point.

leplatrem:
I take your feedbacks as a green light :)

Natim:
If we could find a way not to pass the request to each method but rather have a set_current_request tween that configures the backend.

leplatrem:
> have a set_current_request tween that configures the backend.

Since the registry is shared among threads, this would prevent cliquet applications to work with multithreaded workers.

Let me be clear : I am not a big fan of having those requests passed at every method! But I can't see any other way.

leplatrem:
@Natim was right. And @tarekziade suggested a solution to avoid the global registry limitation.

We should rely on a python `transaction` manager instead of passing request everywhere.

I'll investigate that :)

Closing :)

Issue #290, opened 2015-06-05
Uniformize the naming of getters in our classes


almet:
I've seen that the different contributors to the project don't have the same way to define *getters* on classes.

For instance, on `resource.ViewSet`, I've seen that there are getters prefixed by `get_`, whereas in the *permissions* backend, it's not the case (we have methods named `object_permission_principals` for instance).

In the case of the ViewSet, the design is such that the class can have class attributes (which aren't prefixed by get, because they're attributes) and then getters which start by `get`.

I would advocate we try to follow this pattern everywhere, or update how we do it for viewsets. Thoughts?

Natim:
For the permission backend now that we changed ``object_permission`` with ``ace`` I think it makes sense to add the ``get_``. At first they were there but the method name was too long so I had to remove them. But with ``ace`` it should work.

Also @ametaireau tells me that it may not be the only place where they are missing.

I like the fact that ``get_`` indicate that it is a call and that without the ``get_`` it is a property.

PR #470, opened 2015-09-30
Use project name as setting prefix (fixes #448)

### LINKED ISSUE #448: Settings names based on project name
One of the drawbacks of using cliquet in your app is that your settings names are prefixed with ``cliquet.`` (see [previous discussion](https://github.com/mozilla-services/cliquet/issues/318)).

Same for the management command, called ``cliquet`` to migrate ([see related kinto issue](https://github.com/Kinto/kinto/issues/129))

What if, in Cliquet, we would read the settings with a *dynamic* namespace based on the project name (this only one would be ``cliquet.project_name`` though) ?
Instead of having to define ``cliquet.http_host`` in the application, we would define ``readinglist.http_post``...

Didn't look about possible implementation so far. Just thinking loudly.

leplatrem:
* [x] Refactor usage of settings
* [ ] Document it

@ametaireau @Natim r?


cliquet/errors.py
```diff
@@ -176,7 +176,7 @@ def send_alert(request, message=None, url=None, code='soft-eol'):
     :param url: The URL for more information, default to the documentation url.
     """
     if url is None:
-        url = request.registry.settings['cliquet.project_docs']
+        url = request.setting('project_docs')
```
ametaireau: `request.get_setting()`?

ametaireau:
LGTM otherwise.

ametaireau:
I actually wonder if that would make sense to store the settings without their prefix (at initialization time) so there is no need to ask for prefix + setting then?

leplatrem:
> I actually wonder if that would make sense to store the settings without their prefix

Very good point indeed. Even though that represents a lot more work in the tests code... I'll have a look at what it could be. Especially because we already have some code that manipulates settings in `cliquet.initialize()`

cliquet/errors.py
leplatrem: I liked it short, but I know what you mean :) This is indeed a broader question : #290

PR #471, opened 2015-09-30
Allow override 405 response error message

leplatrem:
Ref https://github.com/mozilla-services/syncto/pull/35

@Natim r?


Natim:
Ok, I didn't noticed that you wanted to raise the new http_error but yes sure.

PR #472, opened 2015-10-01
Use project name as setting prefix (fixes #448)

### LINKED ISSUE #448: Settings names based on project name

leplatrem:
As suggested by @ametaireau, this re-implementation of #470 is ~~much more sweet and light~~ not after discussing :)

Thoughts ? /cc @Natim

### DIFF
--- a/cliquet/initialization.py
+++ b/cliquet/initialization.py
@@ -419,19 +419,19 @@ def handle_project_name_prefix(settings):
     settings = settings.copy()
 
     # First, override cliquet.* settings with <project_name>.* values.
-    for key, value in settings.items():
+    for key, value in list(settings.items()):
         if key.startswith(project_name + '.'):
             prefixed_key = re.sub(r"^%s" % project_name, "cliquet", key)
             settings[prefixed_key] = value
 
     # Then, complete <project_name>.* with cliquet.* values.
-    for key, value in settings.items():
+    for key, value in list(settings.items()):
         if key.startswith('cliquet.'):
             prefixed_key = re.sub(r"^cliquet", project_name, key)
             settings.setdefault(prefixed_key, value)
 
     # Then, use <project_name>.* for unprefixed settings.
-    for key, value in settings.items():
+    for key, value in list(settings.items()):
         if key.startswith(project_name + '.'):
             unprefixed_key = re.sub(r"^.+\.", "", key)
             settings[unprefixed_key] = value
--- a/cliquet/initialization.py
+++ b/cliquet/initialization.py
@@ -381,7 +381,7 @@ def load_default_settings(config, default_settings):
         cliquet_prefix = 'cliquet.' + unprefixed
         return unprefixed, project_prefix, cliquet_prefix
 
-    # # Fill settings with default values if not defined.
+    # Fill settings with default values if not defined.
     for key, default_value in sorted(default_settings.items()):
         unprefixed, project_prefix, cliquet_prefix = keys = _prefixed_keys(key)
         is_defined = len(set(settings.keys()).intersection(set(keys))) > 0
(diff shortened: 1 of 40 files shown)


cliquet/initialization.py
```diff
@@ -367,6 +368,65 @@ def on_new_response(event):
...
+                is_already_set = 'basicauth' in settings['multiauth.policies']
+                if asbool(settings.pop(old)) and not is_already_set:
+                    settings['multiauth.policies'] += ' basicauth'
+            else:
+                settings[new] = settings.pop(old)
+
+    config.add_settings(settings)
+
+
+def handle_project_name_prefix(settings):
```
ametaireau: a description here would be very welcome :)

cliquet/initialization.py
```diff
@@ -367,6 +368,65 @@ def on_new_response(event):
...
+    for key, value in settings.items():
+        if key.startswith('cliquet.'):
+            prefixed_key = re.sub(r"^cliquet", project_name, key)
+            settings.setdefault(prefixed_key, value)
+
+    # Then, use <project_name>.* for unprefixed settings.
+    for key, value in settings.items():
+        if key.startswith(project_name + '.'):
+            unprefixed_key = re.sub(r"^.+\.", "", key)
+            settings[unprefixed_key] = value
```
ametaireau: Is there a way to not loop on the settings three times?

leplatrem: Maybe. But I could not find it.

leplatrem:
After discussing : 

* [ ] Remove prefix from default settings
* [ ] Remove prefix when reading settings at init
* [ ] Remove prefix when reading in app
* [ ] Warn if both `cliquet.*` and `readinglist.*` are defined

cliquet/initialization.py
```diff
@@ -367,6 +368,43 @@ def on_new_response(event):
...
+
+    project_name = settings['project_name']
+
+    def _prefixed_keys(key):
+        unprefixed = key.split('.', 1)[1] if '.' in key else key
+        project_prefix = project_name + '.' + unprefixed
+        cliquet_prefix = 'cliquet.' + unprefixed
+        return unprefixed, project_prefix, cliquet_prefix
+
+    # # Fill settings with default values if not defined.
```
Natim: nit: Remove one #

cliquet/storage/memory.py
```diff
@@ -154,7 +154,7 @@ class Memory(MemoryBasedStorage):
 
     Enable in configuration::
 
-        cliquet.storage_backend = cliquet.storage.memory
+        myapp.storage_backend = cliquet.storage.memory
```
Natim: project_name instead of myapp?

leplatrem:
> Note: See Pyramid source. Not very convincing : https://github.com/Pylons/pyramid/blob/master/pyramid/config/settings.py#L64-L149

cliquet/storage/memory.py
leplatrem: or maybe leave cliquet like in other backends :)

ametaireau:
Looks good to me. r+

PR #474, opened 2015-10-06
Fix multiauth prefix being removed.

Natim:
r? @leplatrem


leplatrem:
r+ 

Thanks @Natim... #myfault

PR #476, opened 2015-10-09
Expose public settings the way they are listed (according to their prefixes).

Natim:
Refs https://github.com/mozilla-services/syncto/issues/48

r? @leplatrem @n1k0

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -7,7 +7,7 @@ This document describes changes between each past release.
 2.9.0 (unreleased)
 ------------------
 
-- Expose public settings without prefix, expect if we explicitely want
+- Expose public settings without prefix, except if we explicitely want
   to expose cliquet.
 
 
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -7,8 +7,9 @@ This document describes changes between each past release.
 2.9.0 (unreleased)
 ------------------
 
-- Expose public settings without prefix, except if we explicitely want
-  to expose cliquet.
+- Expose public settings without prefix, except if we explicitely
+  configure public_settings to expose them (with ``cliquet.`` or
+  ``project_name.``) (refs #476)
 
 
 2.8.0 (2015-10-06)
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -7,7 +7,8 @@ This document describes changes between each past release.
 2.9.0 (unreleased)
 ------------------
 
-- Nothing changed yet.
+- Expose public settings without prefix, expect if we explicitely want
+  to expose cliquet.
 
 
 2.8.0 (2015-10-06)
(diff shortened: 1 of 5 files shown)


leplatrem:
Without more argumentation, I am kind of `r-` on this.

Syncto is a particular use-case. It intends to mimic the Kinto API. It could therefore expose public settings using an explicit prefix.

I remember that @ametaireau didn't like this idea of prefixes in the hello view either :) We can get rid of them if everybody agrees.

Natim:
Yes this is just to be backward compatible. It provide both with prefix and without them.

leplatrem:
In my opinion (may be wrong), the backward compatibility of `kinto.js` should not be handled in cliquet, but in `Kinto`. That's why we have this:
https://github.com/Kinto/kinto/blob/master/kinto/__init__.py#L47-L48

And when I look at the modifications in the tests, it looks more like a breaking change. Please help :)

Natim:
Ok, I can do it in Syncto as well then.

Natim:
Also nevertheless I would like to have them without the app prefix

CHANGELOG.rst
```diff
@@ -7,7 +7,8 @@ This document describes changes between each past release.
 2.9.0 (unreleased)
 ------------------
 
-- Nothing changed yet.
+- Expose public settings without prefix, expect if we explicitely want
```
Natim: except

n1k0:
Can anybody explain me what are the benefits of having prefixes at all? Is it to allow sharing a same name in two different namespaces? Is it likely to ever happen? So, if they're actually useful, let's keep them; if they don't, deprecating them seems fine.

Just my opinion though, I'm not too much into the details — thinking out loud here :)

Natim:
Yes it is also my point, I cannot see the benefit of having the prefix but I can see the benefit of removing them.

Then clients knows that they can assume some settings if the app is cliquet based.
All cliquet apps will eventually use the batch endpoint at some point.

cliquet/tests/test_views_hello.py
```diff
@@ -53,8 +53,8 @@ def test_public_settings_can_be_set_with_and_without_prefix(self):
         response = self.app.get('/')
         settings = response.json['settings']
         self.assertIn('cliquet.paginate_by', settings)
-        self.assertIn('myapp.paginate_by', settings)
-        self.assertIn('myapp.project_version', settings)
+        self.assertIn('paginate_by', settings)
```
Natim: Because the project_name is explicit for ``myapp.paginate_by`` we should probably allow it to be prefixed as we do for cliquet.

Natim:
Updated to keep prefixes if they are present and not adding it if they aren't.

leplatrem:
For the reference, the original idea of prefixes was introduced here : 
https://github.com/mozilla-services/cliquet/issues/318#issuecomment-117709623

I like it this way, it looks consistent to me and compatible with the retro-compability written here:
https://github.com/Kinto/kinto/blob/e079660ee5ac6e4a9e7134d9dc496457149fa8a0/kinto/__init__.py#L48

r+ \o/ :)

PR #481, opened 2015-10-15
Some settings are list but cannot be added to a set.

Natim:
r? @leplatrem

### DIFF
--- a/cliquet/initialization.py
+++ b/cliquet/initialization.py
@@ -395,7 +395,8 @@ def load_default_settings(config, default_settings):
 
         # Fail if not only one is defined.
         defined = set(settings.keys()).intersection(set(keys))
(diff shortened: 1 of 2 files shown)


cliquet/initialization.py
```diff
@@ -395,7 +395,8 @@ def _prefixed_keys(key):
 
         # Fail if not only one is defined.
         defined = set(settings.keys()).intersection(set(keys))
-        distinct_values = set([settings[d] for d in defined])
+        distinct_values = set([str(settings[d]) for d in defined])
```
leplatrem: Won't this cause a problem with settings that contain unicode values ?

Shouldn't we use something like `repr()` instead ?

Natim: We should probably use ``six.text_type``

leplatrem:
r+

PR #484, opened 2015-10-19
Document JSON patch format (fixes #482)

### LINKED ISSUE #482: Document the format of supported patch
Hello,

the documentation doesn't explicitly tell which patch format it supports. Does it support only key-value replace?

Is json-patch format (http://jsonpatch.com/) (http://tools.ietf.org/html/rfc6902) support planned in the future?

leplatrem:
review? @Lothiraldan


Lothiraldan:
:+1:

PR #487, opened 2015-10-19
Replace DROP INDEX IF EXISTS by a DO block

rodo:
It's often a bad idea to DROP an index and recreate it just after to ensure the script is idempotent.
Even if the CREATE INDEX IF NOT EXISTS will be available soon in PostgreSQL 9.5 I suggest to replace the DROP INDEX instructions

### DIFF
--- a/cliquet/storage/postgresql/schema.sql
+++ b/cliquet/storage/postgresql/schema.sql
@@ -1,4 +1,8 @@
 --
+-- Automated script, we do not need NOTICE and WARNING
+--
+SET client_min_messages TO ERROR;
+--
 -- Convert timestamps to milliseconds epoch integer
 --
 CREATE OR REPLACE FUNCTION as_epoch(ts TIMESTAMP) RETURNS BIGINT AS $$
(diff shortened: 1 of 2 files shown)


cliquet/storage/postgresql/schema.sql
```diff
@@ -26,13 +30,53 @@ CREATE TABLE IF NOT EXISTS records (
...
-DROP INDEX IF EXISTS idx_records_parent_id_collection_id_last_modified;
-CREATE UNIQUE INDEX idx_records_parent_id_collection_id_last_modified
+  IF NOT EXISTS (
+    SELECT 1 FROM pg_indexes
+       WHERE indexname = 'idx_records_parent_id_collection_id_last_modified'
+       AND tablename = 'records'
+  ) THEN
+
+  CREATE UNIQUE INDEX idx_records_parent_id_collection_id_last_modified
     ON records(parent_id, collection_id, last_modified DESC);
```
leplatrem: /me wondering: Since we are in a `DO` block, couldn't we catch the creation exception instead (less verbose) ?

(It might fill the server logs with meaningless errors though)

leplatrem:
\o/ Thanks for this awesome contribution !

You can add your name to CONTRIBUTORS.rst :) We maintain it by hand.

cliquet/storage/postgresql/schema.sql
rodo: What do you mean exactly, by now if the index is not exists it will be created, and if it exists nothing will happen. I fixed an error in https://github.com/rodo/cliquet/commit/1477c25e6c8cbeefb0951a9a86a593dccac4e694 suppose it's related on what you mean.

leplatrem: Well, what I mean is that there are two ways to write the same behaviour.
For example, in python:

```python
if key in mydict:
    print mydict[key]
```
is equivalent (and less performant btw) to: 
```python
try:
    print mydict[key]
except KeyError:
    pass
```

So for the index creation I guess it's the same.
Either : 
```sql
DO $$
BEGIN
    IF NOT EXISTS ( ... )
    THEN
         CREATE INDEX
    END IF;
END$$;
```

or 


```sql
DO $$
BEGIN
    BEGIN
         CREATE INDEX ...
    EXCEPTION WHEN <error> THEN
            RAISE NOTICE 'Index exists with this name';
    END;
END$$;
```

Indeed, it would have prevented the typo you add in the commit you mentionned :)

leplatrem:
I realize that we should mimic the same improvements in the other backends `cliquet/cache/postgresql/schema.sql` and `cliquet/permissions/postgresql/schema.sql`

Idempotence for the win!

cliquet/storage/postgresql/schema.sql
rodo: Yes we can do that, but the log will be **more verbose** as at each time you'll execute the file on your DB you'll have an error, I think we have a different goal. As a DBA I always want to reduce log amount, I prefer script that run without raising error, even if I have less information, here you want to Raise an error if you try to create an already existing index, the opposite side of view.
We have to define the goal before choosing a solution.
Be careful of the line number **4** that set min_loglevel to **ERROR** you won't see the **NOTICE**

leplatrem: >  log will be more verbose

Yes, that what I thought in my first comment :) 

This works for me ! I just wanted to challenge your choice :) 

Thanks for the details !

Issue #32, opened 2015-02-26
Add custom Pyramid events


leplatrem:
* At endpoint level
* At storage level
* At authentication level

ametaireau:
If I understand this correctly, the aim here is to be able to trigger some code when specific events arise.

Let's try to list them.

### At storage level:

- a new resource is created (send resource id)
- a resource had been deleted (resource id)
- a new record had been added (resource id + record id)
- a record had been modified (resource id + record id)
- a record had been deleted (resource id + record id)

### At authentication level:

- A user had failed to authenticate (user id)
- A user has successfully logged in (user id)
- Access had been denied to a resource (resource id + verb + user id)
- Access had been granted to a resource (resource id + verb + user id)

I don't understand what "endpoint level" is, and as such I fail to see which events should be triggered there.

Natim:
For me some of your storage level event are in fact endpoint level events.

You could have event on batch and for most of the event, I think the full record should be present (not only its id).

ametaireau:
I've updated my previous comment to send records rather than record ids.

leplatrem:
> the full record should be present
+1

Indeed, I didn't explicit what would be an "endpoint" :) I meant like incoming request, but Pyramid provides something similar...

Instead, @ametaireau suggested `pre_` and `post_` on storage, which makes more sense

Natim:
How do you handle ``pre_`` and ``past_batch``?

Natim:
The problem I see with event on storage is that you cannot have separate hooks for separate collections.

ametaireau:
It really depends how you implement the event subscriber.

For instance, Let's say you subscribe to all "pre update" events, for all collections.

Then, when your function is triggered, you're given the resource id + the record and you can do routing from there. Your application logic can do anything it feels good to trigger a different hook depending on a routing logic it defines (be it the collection id or anything else).

leplatrem:
Implemented in #488

PR #488, opened 2015-10-19
Notification system

tarekziade:
Stuff to add before it's ready :
- [x] Agree on the payload v1 in cliquet/events.py
- [x] test : No event when POST on collection that returns 200 (existing id provided)
- [x] test : PUT that create
- [x] test : PUT that replace
- [x] test : PATCH whose fields are identical to existing ones
- [x] Update documentation (add new section?)
- [x] Update the roadmap, we now have notifications !
- [x] Add some logging during the setup of listeners


cliquet/tests/resource/test_views.py
```diff
@@ -678,3 +683,31 @@ def test_next_page_url_relies_on_headers_information(self):
         resp = self.app.get(self.collection_url + '?_limit=1',
                             headers=headers)
         self.assertIn('https://server.name:443', resp.headers['Next-Page'])
+
+
+class EventsTest(BaseWebTest, unittest.TestCase):
```
leplatrem: move to separate `test_events` file :)

cliquet/events.py
```diff
@@ -0,0 +1,15 @@
+from cliquet.utils import strip_uri_prefix
+
+
+class CollectionChanged(object):
+    """Triggered when a collection is changed.
+    """
+    def __init__(self, action, collection, request):
```
leplatrem: nit: collection is actually a `resource` here

cliquet/listeners/redis.py
```diff
@@ -0,0 +1,44 @@
...
+import json
+
+from cliquet.listeners import ListenerBase
+
+
+class RedisListener(ListenerBase):
+    def __init__(self, *args, **kwargs):
+        super(RedisListener, self).__init__(*args, **kwargs)
+        maxconn = kwargs.pop('max_connections')
+        self._listname = kwargs.pop('listname', 'cliquet.events')
```
leplatrem: nit: duplicated default with load_from_config

cliquet/tests/resource/test_events.py
```diff
@@ -0,0 +1,34 @@
...
+        return app
+
+    def test_events_triggered(self):
+        # let's subcribe to the CollectionChanged event
+        events = []
+
+        def listener(event):
+            events.append(event)
+
+        self.config.add_subscriber(listener, CollectionChanged)
```
leplatrem: you can use a mock instead I guess

cliquet/tests/resource/test_events.py
```diff
@@ -0,0 +1,34 @@
...
+        self.config.commit()
+
+        # creating a record should trigger an event
+        body = {'data': {'name': 'de Paris'}}
+        self.app.get(self.collection_url, headers=self.headers, status=200)
```
leplatrem: I know this is tempting to do it that way :)  But the assertion is hidden.

Please use setUp to plug your listener and add specific tests like `test_does_not_trigger_on_read` ... `triggers_on_write` etc.

cliquet/tests/resource/test_collection.py
```diff
@@ -1,5 +1,4 @@
 from pyramid import httpexceptions
-
```
leplatrem: We usually separate external from internal imports using a blank line

cliquet/tests/test_listeners.py
```diff
@@ -0,0 +1,101 @@
...
+        self.config = testing.setUp()
+        pool = redis.BlockingConnectionPool(max_connections=1)
+        self._redis = redis.StrictRedis(host='localhost', port=6739, db=0,
+                                        connection_pool=pool)
+        self._size = 0
+
+    def _save_redis(self):
+        self._size = self._redis.llen('cliquet.events')
+
+    def redis_changed(self):
```
leplatrem: nit: `has_redis_changed` ?

cliquet/tests/test_listeners.py
```diff
@@ -0,0 +1,101 @@
...
+    StrictRedis.lpush = old
+
+
+class ListenerCalledTest(unittest.TestCase):
+
+    def setUp(self):
+        self.config = testing.setUp()
+        pool = redis.BlockingConnectionPool(max_connections=1)
+        self._redis = redis.StrictRedis(host='localhost', port=6739, db=0,
+                                        connection_pool=pool)
```
leplatrem: I would suggest to use `load_from_config` here

cliquet/listeners/redis.py
leplatrem: make `self.listname` a public attribute

cliquet/tests/test_listeners.py
```diff
@@ -0,0 +1,101 @@
...
+        self._size = 0
+
+    def _save_redis(self):
+        self._size = self._redis.llen('cliquet.events')
+
+    def redis_changed(self):
+        return self._redis.llen('cliquet.events') > self._size
```
leplatrem: `self._redis.llen(self._redis.listname)`

cliquet/tests/test_listeners.py
```diff
@@ -0,0 +1,101 @@
...
+            class Request(object):
+                path = '/1/bucket/collection/'
+                unauthenticated_userid = 'tarek'
+                matchdict = {'bucket_id': 'bucket_id'}
+
+            # now let's trigger an event
+            event = CollectionChanged('create', Collection(), Request())
+            self.notify(event)
+            self.assertTrue(self.redis_changed())
+
```
leplatrem: Please split into independant tests using a common setup :)

cliquet/initialization.py
```diff
@@ -367,6 +367,21 @@ def on_new_response(event):
...
+                 settings['event_listeners'].split(',')
+                 if listener.strip()]
+
+    # XXX we should rename it to "RecordChanged" here...
+    from cliquet.events import CollectionChanged
+    #
+    for listener in listeners:
+        listener_mod = config.maybe_dotted(listener)
+        listener = listener_mod.load_from_config(config)
+        config.add_subscriber(listener, CollectionChanged)
```
leplatrem: I really like what you suggested in the design docs about kwargs being passed automatically from config...

Any reason you gave up ?

I think that's what Pyramid multiauth does, it would great to mimic it no  ?


```
cliquet.listeners = redis http

cliquet.listeners.redis.callable = cliquet.listeners.redis
cliquet.listeners.redis.url = http://localhost:6739
cliquet.listeners.redis.pool_size = 10

cliquet.listeners.http.callable = cliquet.listeners.http
cliquet.listeners.http.endpoint = http://balrog/api
```

cliquet/resource.py
```diff
@@ -541,10 +551,16 @@ def apply_changes(self, record, changes):
...
                 raise_invalid(self.request, name=field, description=error)
 
-    def postprocess(self, result):
+    def postprocess(self, result, action=ACTION_READ):
         body = {
             'data': result
         }
+
+        if action != ACTION_READ:
+            # the collection has been changed
```
leplatrem: Add a comment that later it should be integrated with transaction system

cliquet/events.py
```diff
@@ -0,0 +1,15 @@
...
+        self.request = request
+
+        self.payload = {'collection_id': collection.record_id,
+                        'timestamp': collection.timestamp,
+                        'action': action,
+                        'uri': strip_uri_prefix(request.path),
+                        'user_id': request.unauthenticated_userid,
+                        'bucket_id': request.matchdict.get('bucket_id')}
```
leplatrem: Oh I would not do that. There is no notion of bucket in cliquet !

I suggest:
```
matchdict = dict(request.matchdict)
matchdict[resource.name + '_id'] = matchdict.pop('id')
payload.update(**matchdict)
```

cliquet/tests/test_listeners.py
tarekziade: Which load_from_config ? This is not to load a RedisListener. This is a redis client to verify values in Redis.

cliquet/events.py
```diff
@@ -0,0 +1,15 @@
+from cliquet.utils import strip_uri_prefix
+
+
+class CollectionChanged(object):
+    """Triggered when a collection is changed.
+    """
+    def __init__(self, action, collection, request):
+        self.request = request
+
+        self.payload = {'collection_id': collection.record_id,
```
leplatrem: Nope, `collection.record_id` is the record id :)

cliquet/tests/test_listeners.py
tarekziade: self._redis is not a RedisListener class. It's a pure redis connector

cliquet/events.py
Natim: I would like to settle on the payload format first.

cliquet/events.py
```diff
@@ -0,0 +1,15 @@
...
+                        'timestamp': collection.timestamp,
+                        'action': action,
+                        'uri': strip_uri_prefix(request.path),
+                        'user_id': request.unauthenticated_userid,
```
leplatrem: I would suggest to use `request.prefixed_userid` that contains the auth policy name in addition to the user id

cliquet/initialization.py
Natim: :+1:

cliquet/listeners/redis.py
```diff
@@ -0,0 +1,44 @@
...
+        connection_pool = redis.BlockingConnectionPool(max_connections=maxconn)
+        self._client = redis.StrictRedis(connection_pool=connection_pool,
+                                         **kwargs)
+
+    def __call__(self, event):
+        try:
+            payload = json.dumps(event.payload)
+        except Exception:
```
Natim: What exception do you except here? If any, let it raise because cliquet will handle it at a upper level.

cliquet/listeners/redis.py
```diff
@@ -0,0 +1,44 @@
...
+            # XXX we should log this
+            return
+
+        try:
+            self._client.lpush(self._listname, payload)
+        except Exception:
+            # XXX we should log this
```
Natim: Here, we should probably use [the ``wrap_redis_error`` decorator](https://github.com/mozilla-services/cliquet/blob/master/cliquet/storage/redis.py#L14)

cliquet/resource.py
```diff
@@ -18,6 +18,14 @@
     COMPARISON, classname, native_value, decode64, encode64, json,
     current_service, encode_header, decode_header
 )
+from cliquet.events import CollectionChanged
+
+
+# event actions
+ACTION_CREATE = 'create'
```
Natim: Maybe we can consider using ACTIONS.CREATE as we do for errors here: https://github.com/mozilla-services/cliquet/blob/master/cliquet/errors.py#L8-L28

cliquet/tests/resource/test_events.py
leplatrem: Using a mock it would like (totally untested):
```
listener = mock.Mock()
self.config.add_subscriber(listener, CollectionChanged)
...
self.app.get(..)
self.assertFalse(listener.called)
```
or
```
self.app.post(..)
self.assertTrue(listener.called)
```

and if you want to inspect events:
```
args, kwargs = listener.callargs_list[-1]
self.assertEqual(args[0], 'create')
```

cliquet/resource.py
```diff
@@ -282,7 +290,8 @@ def collection_delete(self):
         filters = self._extract_filters()
         deleted = self.collection.delete_records(filters=filters)
 
-        return self.postprocess(deleted)
+        action = len(deleted) > 0 and ACTION_DELETE or ACTION_READ
```
Natim: Do we really want to trigger the even only if something was deleted? I think this should be up to the even handler to decide with the related data what to do in that case.

Anyway why would you like to send a read action in that case?

cliquet/resource.py
```diff
@@ -367,7 +376,8 @@ def put(self):
         timestamp = record[self.collection.modified_field]
         self._add_timestamp_header(self.request.response, timestamp=timestamp)
 
-        return self.postprocess(record)
+        action = existing and ACTION_UPDATE or ACTION_CREATE
```
Natim: :+1:

cliquet/resource.py
```diff
@@ -541,10 +551,16 @@ def apply_changes(self, record, changes):
...
-    def postprocess(self, result):
+    def postprocess(self, result, action=ACTION_READ):
         body = {
             'data': result
         }
+
+        if action != ACTION_READ:
+            # the collection has been changed
+            event = CollectionChanged(action, self, self.request)
+            self.request.registry.notify(event)
```
Natim: I think we should notify the even only if the configuration state that we want to receive notification for this kind of even on this kind of resource.

cliquet/resource.py
```diff
@@ -541,10 +551,16 @@ def apply_changes(self, record, changes):
             for field, error in e.asdict().items():
                 raise_invalid(self.request, name=field, description=error)
 
-    def postprocess(self, result):
+    def postprocess(self, result, action=ACTION_READ):
         body = {
             'data': result
         }
+
+        if action != ACTION_READ:
```
Natim: When do you notify for the ``ACTION_READ`` event?

cliquet/resource.py
Natim: Reading further, I feel a bit embarrassed that you consider ACTION_READ as being: not sending event.

cliquet/tests/resource/test_events.py
Natim: Well there are two things here: 
 - Unittesting the EventListener -> In that case Mock is a good idea
 - Functionnality Testing Listener (Redis / HTTP) -> Apparently this is the case in a following file.

cliquet/tests/test_listeners.py
```diff
@@ -0,0 +1,101 @@
...
+        config = testing.setUp()
+        with mock.patch.dict(config.registry.settings,
+                             [('event_listeners', listener)]):
+            with redis_class as redis_mocked:
+                initialization.setup_listeners(config)
+                self.assertTrue(redis_mocked.called)
+
+
+@contextmanager
+def broken_redis():
```
Natim: Why not use

```
with mock.patch('cliquet.listeners.redis.StrictRedis') as mocked_redis:
    mocked_redis.return_value.lpush.side_effect = Exception('boom')
```

cliquet/tests/resource/test_events.py
tarekziade: @leplatrem What is the benefit of using a mock here ? I find the code you're showing is very complicated and convulted compared to the 3 simple lines I have. 

I find mocks very cool when we need to mock an object which methods that might be called by external code, but in the present case it seems overkill unless I am missing something

cliquet/initialization.py
tarekziade: > I really like what you suggested in the design docs about kwargs being passed automatically from config

Not sure what you are talking about. Are you talking about having several sections in the ini ? I am still waiting for feedback on https://github.com/Kinto/kinto/issues/220

cliquet/listeners/redis.py
tarekziade: any exception here should be catched so we are sure that the data is stored and the end user gets back a valid response. 

by design, a notification that fails should not interfere with the request that changes the data.

cliquet/resource.py
tarekziade: Do we really want to send events everytime a record is read ? what is the use case ?

cliquet/resource.py
tarekziade: we don't, only if the data is changed because that's the use case

cliquet/tests/test_listeners.py
leplatrem: Ok got it, sorry for the noise

cliquet/tests/resource/test_events.py
leplatrem: Ok maybe it is matter of taste :) Forget about mock here then !

cliquet/tests/resource/test_events.py
leplatrem: You should also assert that:
* create action on PUT when record does not exist
* no create action on POST when record already exists
* update on PATCH and PUT with existing
* delete on DELETE :)

cliquet/initialization.py
leplatrem: Let's imagine that we don't have the several sections. 

Pyramid multiauth has this pattern of using prefixes to simulate them.  I saw this convention in other pyramid projects too.

> still waiting for feedback on Kinto/kinto#220

Let's say both approaches behave the same way about kwargs being passed from config, it is just a matter of format/syntax IMO

cliquet/events.py
leplatrem: Yes, what you suggested in the initial document was too tight to original http request IMO (verb, status code, etc.)

cliquet/listeners/redis.py
leplatrem: > a notification that fails should not interfere

Then maybe the catchall should be around the notify in resource then ?

cliquet/resource.py
Natim: My point is to not send any notificaton except if they are activated. And if you want to ping a board at each collection_get you would want to activate notification on ACTION.READ.

Because we are building a generic solution we want this to be possible if needed.

cliquet/resource.py
leplatrem: nit: `action = ACTION_UPDATE if existing else ACTION_CREATE `

cliquet/listeners/redis.py
tarekziade: That'd be yet another error: the registry failed to notify the event in Kinto. The exception I am catching here are different: the event was notified, but the listener that listens to them failed to work.

cliquet/resource.py
tarekziade: this use case is very cool but I think we should narrow the scope of this PR to the use case, and keep in mind your proposal for the next iterations

cliquet/initialization.py
tarekziade: sounds cool

cliquet/listeners/redis.py
leplatrem: Yes but since the listeners are executed synchronously it would be equivalent and would apply to every listeners. I guess I prefer your solution, since we may to let listeners trigger some httpexceptions for example

cliquet/resource.py
leplatrem: I agree with tarek here, let's start with something straightforward, and let the listeners do the filtering.

In the future we can imagine to introduce what you suggest from the configuration (per resource, per action, etc.)

cliquet/tests/test_listeners.py
tarekziade: I tried but that did not work. Happy to simplify that part if there's a way.

cliquet/events.py
tarekziade: resource.name <-- what is the resource object in that case ? 

cliquet.collection.Collection and cliquet.resource.Resource  don't have a name attribute

cliquet/listeners/redis.py
tarekziade: > Yes but since the listeners are executed synchronously it would be equivalent and would apply to every listeners.

Not equivalent. 

For instance if I have a redis listener and an http listener (in that order) and if only the redis listener fails before the redis server is down or something, catching the error at the upper level means that the http listener will never be called. 

That said, adding a try..catch at the upper level is also a good idea

cliquet/initialization.py
```diff
@@ -367,6 +367,20 @@ def on_new_response(event):
     config.add_subscriber(on_new_response, NewResponse)
 
 
+def setup_listeners(config):
+    settings = config.get_settings()
+    listeners = [listener.strip() for listener in
+                 settings['event_listeners'].split(' ')
+                 if listener.strip()]
```
leplatrem: nit: I think there is a `aslist` in `pyramid.settings`

cliquet/tests/resource/test_events.py
```diff
@@ -0,0 +1,87 @@
...
+        resp = self.app.post_json(self.collection_url, self.body,
+                                  headers=self.headers, status=201)
+        record = resp.json['data']
+        record_url = self.get_item_url(record['id'])
+
+        self.app.delete(record_url, headers=self.headers, status=200)
+        self.assertEqual(len(self.events), 2)
+        self.assertEqual(self.events[0].payload['action'], ACTIONS.CREATE)
+        self.assertEqual(self.events[1].payload['action'], ACTIONS.DELETE)
+
```
leplatrem: Three missing test cases IHMO:
* No event when POST on collection that returns 200 (existing id provided)
* PUT that create
* PUT that replace

:)

cliquet/events.py
```diff
@@ -0,0 +1,22 @@
...
+               READ='read',
+               UPDATE='update')
+
+
+class ResourceChanged(object):
+    """Triggered when a resource is changed.
+    """
+    def __init__(self, action, resource, request):
+        self.request = request
+        self.payload = {'timestamp': resource.timestamp(),
```
leplatrem: `resource.timestamp` is supposed to be an attribute ! oO !

cliquet/tests/resource/test_events.py
leplatrem: Fourth one:

* A PATCH whose fields are identical to existing ones

cliquet/events.py
tarekziade: what is "resource" ? please give me the name of the class I should have here. because below you are using resource.name and I can't find it in the code. timestamp is a function in collection

cliquet/events.py
```diff
@@ -0,0 +1,22 @@
...
+                        'action': action,
+                        'uri': strip_uri_prefix(request.path),
+                        'user_id': request.prefixed_userid}
+
+        # matchdict = dict(request.matchdict)
+        # matchdict[resource.name + '_id'] = matchdict.pop('id')
+        # self.payload.update(**matchdict)
```
leplatrem: I would vote for that :)

cliquet/tests/resource/test_events.py
tarekziade: > No event when POST on collection that returns 200 (existing id provided)

This is not tested anywhere for now in the views.  I can't test this unless we set unique_fields to ('id',) I guess - so I am wondering, why this is not the case by default ? 

In the current test setup the default storage (redis) is not checking unicity outside test_storage - should I change it globally ? or should I do this change only in test_events ?

tarekziade: Tests:
- [x] No event when POST on collection that returns 200 (existing id provided)
- [ ] PUT that create
- [ ] PUT that replace
- [ ] PATCH whose fields are identical to existing ones

Natim:
I have created #492 and #493 to handle some more stuff after this PR.

cliquet/events.py
```diff
@@ -0,0 +1,28 @@
...
+               UPDATE='update')
+
+
+class ResourceChanged(object):
+    """Triggered when a resource is changed.
+    """
+    def __init__(self, action, resource, request):
+        self.request = request
+        service = current_service(request)
+        resource_id = service.viewset.get_name(resource.__class__) + '_id'
```
leplatrem: With what @Natim suggested in #492 it becomes clear that we will want the resource name in the payload.

In the case of kinto, they are `bucket`, `group`, `collection`, `record`.

It can still be shipped as is, but if there is still some room for last-minute changes :)

leplatrem:
Among the remaining tasks, I think of : 

* [ ] Update documentation (add new section?)
* [ ] Update the roadmap, we now have notifications !
* [ ] Add some logging during the setup of listeners
* [ ] Add some statsd for listeners ?
* [ ] Add impacted records to payload

Do you agree with that ? Should I create individual issues for each of them ?

The original issue for this was https://github.com/mozilla-services/cliquet/issues/32. But it mentions «pre»/«post» storage hooks. Should we close it with this PR ? And open another issue for hooks where records can be manipulated before/after interactions with storage backend ?

tarekziade:
I would like to merge the PR asap before it rots, so I guess the most important stuff (besides the small tasks you added) is to agree on the payload v1. Let's talk about this today and decide

tarekziade:
I am adding https://github.com/mozilla-services/cliquet/issues/501 where I am listing stuff to add in the payload later

leplatrem:
FYI the roadmap I had in mind was this : http://cliquet.readthedocs.org/en/latest/rationale.html#roadmap

tarekziade:
I can't see any statsd used anywhere in cliquet. I am create a new bug for adding statsd not only in notifications

Natim:
It looks good to me apart from the failing test.

PR #490, opened 2015-10-20
Organize code into a resource module

leplatrem:
* [x] Group resource, viewset, schema into a `resource` module. Since all this is optional to use.
* [x] Rename `Collection` to `Model`, since it is what it is :)
* [x] Document
* [x] Support deprecation

@Natim @ametaireau  thoughts ?


cliquet/resource/__init__.py
```diff
@@ -0,0 +1 @@
+from .base import *
```
Natim: We should not import * there, either move the base code here or define the ``__all__``  variable.

leplatrem: yes ok, but what do you think of the overall move ?

leplatrem:
Ready to review !

Natim:
It looks good to me.

PR #491, opened 2015-10-21
Use composition instead of inheritance for PostgreSQL client

leplatrem:
@Natim r?


Natim:
It looks great thanks :+1:

PR #497, opened 2015-10-22
Add Makefile rule build-requirements

Natim:
r? @leplatrem

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -28,6 +28,12 @@ $(PYTHON):
 	$(VIRTUALENV) $(VENV)
 	$(VENV)/bin/pip install -U pip
 
+build-requirements:
+	@rm -fr /tmp/cliquet
+	$(VIRTUALENV) /tmp/cliquet
+	/tmp/cliquet/bin/pip install -Ue ".[monitoring,postgresql]"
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -38,6 +38,12 @@ clean:
 	find . -name '*.pyc' -delete
 	find . -name '__pycache__' -type d -exec rm -fr {} \;
 
+distclean: clean
+	rm -fr *.egg *.egg-info/
+
+maintainer-clean: distclean
+	rm -fr .venv/ .tox/ dist/ build/
```
Natim: Maybe dist and build should be in the distclean command.

cliquet_docs/contributing.rst
```diff
@@ -143,7 +143,7 @@ Step 1
      $ git checkout -b prepare-X.Y.Z
      $ prerelease
      $ vim docs/conf.py
-     $ rm -rf .venv
+     $ make maintainer-clean
```
leplatrem: I am not a big fan of this to update the requirements.

Since this removes the virtualenv, we have to re-install `pip install "zest.releaser[recommended]"` otherwise the wheel won't be published on pypi.

leplatrem:
:+1:  :)

PR #502, opened 2015-10-22
Rename resource collection attribute to model

leplatrem:
Follow-up of #490 

This way there will be no confusion. The term «collection» will be only used for one thing : the list of records for a resource. 

@Natim r?


Natim:
r+

PR #504, opened 2015-10-26
Add cliquet.includes setting

leplatrem:
With `pyramid.includes` the plugins are loaded before the main application.

As a consequence, when a cliquet plugin is loaded, some Cliquet setup is not ready yet.

So far, we handled this with safety and double-initialization. But it had some drawbacks:

- the URL prefix mecanism was not setup yet, thus preventing plugins registering their URLs with or without prefixes
- the settings prefixes were not handled yet, preventing plugins to rely on general settings properly (e.g. `cliquet.cache_backend` vs. `kinto.cache_backend`)
etc.
- common libraries like cornice had to be included twice (in plugin to scan views and in cliquet in case no plugin was used)

This PR add a simple include mecanism, just as `pyramid.includes`, that will include plugin once basic cliquet stuff is initialized.

It does not prevent `pyramid.includes` to be used, but when a cliquet plugin relies on cliquet objects or mecanism, it shall be included with `cliquet.includes` (or `kinto.includes`)

@Natim r?

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -10,6 +10,8 @@ This document describes changes between each past release.
 **New features**
 
 - Added notifications (#488)
+- Added ``cliquet.includes`` setting allowing loading of plugins once Cliquet
+  is initialized (unlike ``pyramid.includes``).
 
 **Protocol**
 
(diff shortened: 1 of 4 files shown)


Natim:
Yes it sounds good to me :+1:

Issue #480, opened 2015-10-14
Add a statsd helper.


Natim:
For now, it is quite verbose to use the statsd client:

```python
statsd = request.registry.statsd
if statsd:
    statsd.count("syncclient.status_code.200")
```

In https://github.com/mozilla-services/syncto/pull/49 @ametaireau suggested that we add a helper that let us do that:

```python
statsd(request, "count", "syncclient.status_code.200")
```

It could be good to also replace the use of this helper anywhere it is needed.

greeshmab:
the helper function takes request and the status code as parameters.

leplatrem:
> the helper function takes request and the status code as parameters. 

From what I read in the original description, it takes any kind of string. 

As a first version, I would suggest the helper to be `statsd_count` for example.

The count is called with any string passed:

```python
statsd_count(request, 'record-created')
statsd_count(request, 'error.found')
``` 

In a second iteration, if it is relevant (not clear to me to be honest), it could be interesting to have a single helper that receives the method name (`count`, `timer`, ...) as a parameter:

```python
statsd(request, 'count', 'record-created')
statsd(request, 'count', 'error.found')

with statsd(request, 'timer', 'computation.prime.numbers'):
    compute_prime_number()
```

leplatrem:
Fixed with #574

PR #508, opened 2015-10-27
Add a statsd helper. #480

greeshmab:
Adding a statsd helper function to get the status code


cliquet/statsd.py
```diff
@@ -33,6 +33,10 @@ def count(self, key, unique=None):
         else:
             return self._client.set(key, unique)
 
+def status_code(request,str):
+	statsd=request.registry.statsd
+	if statsd:
+		statsd.count(str);
```
Natim: Please use 4 spaces rather than tabulation in your code.

cliquet/statsd.py
```diff
@@ -33,6 +33,10 @@ def count(self, key, unique=None):
         else:
             return self._client.set(key, unique)
 
+def status_code(request,str):
+	statsd=request.registry.statsd
```
Natim: nit missing space around ``=`` operator

cliquet/statsd.py
```diff
@@ -33,6 +33,10 @@ def count(self, key, unique=None):
         else:
             return self._client.set(key, unique)
 
+def status_code(request,str):
```
Natim: ``status_code`` should be rename ``statsd_count``

``str`` is a reserved keyword in python, please use ``count_key`` instead.

You need a space betwee ``,`` and ` str``

Natim:
Thank you for your work on this @greeshmab.
Could you please work on a test to make sure that it works if statsd is not defined and that if it is defined, statsd is called with the right key?

ametaireau:
Thanks for taking the time to write a patch. It still needs some work in order to be accepted!

Natim:
Superseeded by #574

PR #510, opened 2015-10-27
One transaction per request

leplatrem:
* [x] Switch to SQLA sessions
* [x] Plug SQLA session with pyramid transaction manager
* [x] Share session among backend instances
* [x] Change batch response content when one subrequest error occurs
* [x] Test that crash within request rollbacks global transaction
* [ ] ~~Two-phase commit accross backends ?~~ Session is shared accross backends if URL is the same

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -9,8 +9,17 @@ This document describes changes between each past release.
 
 **Protocol**
 
-* ``_since`` and ``_before`` now accepts an integer value between quotes ``"``,
+- ``_since`` and ``_before`` now accepts an integer value between quotes ``"``,
   as it would be returned in the ``ETag`` response header.
+- A batch request now fails if one of the subrequests fails (#510)
+  (*see new feature about transactions*)
+
+**New features**
+
+- A transaction now covers the whole request/response cycle (#510, Kinto/kinto#194).
+  If an error occurs during the request processing, every operation performed
+  is rolled back. **Note:** This is only enabled with *PostgreSQL* backends. In
+  other words, the rollback has no effect on backends like *Redis* or *Memory*.
 
 - Add the ``protocol_version`` to tell which protocol version is
   implemented by the service. (#324)
@@ -34,6 +43,7 @@ This document describes changes between each past release.
 - Switch to SQLAlchemy for smarter connections pools.
 - Added a simple end-to-end test on a *Cliquet* sample application, using
   `Loads <http://github.com/loads/>`_ (fixes #512)
+- Switched to SQLAlchemy sessions instead of raw connections and cursors (#510)
 
 
 2.10.2 (2015-11-10)
--- a/cliquet/__init__.py
+++ b/cliquet/__init__.py
@@ -79,6 +79,7 @@ DEFAULT_SETTINGS = {
     'storage_url': '',
     'storage_max_fetch_size': 10000,
     'storage_pool_size': 25,
+    'transaction_per_request': True,
     'userid_hmac_secret': '',
     'version_prefix_redirect_enabled': True,
     'trailing_slash_redirect_enabled': True,
@@ -109,6 +110,9 @@ def includeme(config):
 
     # Setup cornice.
     config.include("cornice")
+
+    # Per-request transaction.
+    config.include("pyramid_tm")
 
     # Include cliquet plugins after init, unlike pyramid includes.
     includes = aslist(settings['includes'])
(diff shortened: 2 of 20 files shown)


CHANGELOG.rst
```diff
@@ -9,13 +9,22 @@ This document describes changes between each past release.
...
+- ``_since`` and ``_before`` now accepts an integer value between quotes ``"``,
   as it would be returned in the ``ETag`` response header.
+- A batch request now fails if one of the subrequests fails (#510)
+  (*see new feature about transactions*)
+
+**New features**
+
+- A transaction now covers the whole request/response cycle (#510, Kinto/kinto#194).
+  If an error occurs during the request processing, every operation performed
+  is rolled back. **Note:** This is only enabled with *PostgreSQL* backends.
```
tarekziade: what happens when I have a postgres backend for the storage and redis for the cache ? can you confirm cache changes are also rollbacked ?

cliquet/__init__.py
```diff
@@ -30,6 +30,7 @@
     'cache_backend': '',
     'cache_url': '',
     'cache_pool_size': 25,
+    'cache_transaction_request': True,
```
tarekziade: I am a bit confused by the granularity of this configuration. it seems to me that it's a recipe for failure to allow transaction on storage but not on cache and/or permissions -- since their data is interdependant. imho, we should use transaction or not.

leplatrem: :+1: I did it for *consistency* with other parameters, but it's stupid.

cliquet/storage/postgresql/client.py
```diff
@@ -17,41 +21,44 @@ def __init__(self, engine):
...
-                trans.commit()
-            # Give back to pool.
-            connection.close()
+                # Mark session as dirty.
+                self.invalidate(session)
+            # Success
+            if with_transaction:
+                session.commit()
+                # Give back to pool.
+                session.close()
```
tarekziade: shouldn't we close() whatever happens here ? e.g. try..finally

leplatrem: :+1: good catch

leplatrem:
This is now ready to review/merge

@almet @tarekziade r?

cliquet/storage/postgresql/client.py
```diff
@@ -17,41 +21,44 @@ def __init__(self, engine):
...
 
         A COMMIT is performed on the current transaction if everything went
         well. Otherwise transaction is ROLLBACK, and everything cleaned up.
-
-        XXX: Committing should not happen here but using a global transaction
-        manager like `pyramid-tm`.
         """
-        connection = None
-        trans = None
+        with_transaction = (not readonly and self.commit_manually)
```
tarekziade: you don't need parentheses

cliquet/storage/postgresql/client.py
```diff
@@ -62,25 +69,38 @@ def create_from_config(config, prefix=''):
                    "Refer to installation section in documentation.")
         raise ImportWarning(message)
 
-    settings = config.get_settings()
-    url = settings[prefix + 'url']
+    from zope.sqlalchemy import ZopeTransactionExtension, invalidate
```
tarekziade: any reason we want local imports ? I see zope.sqlalchemy in the regular reqs

leplatrem: no it's in `POSTGRESQL_REQUIRES`  (`requirements.txt` is more like a good old *known good set*)

cliquet/storage/postgresql/client.py
leplatrem: I know, it's a custom we have for boolean exprs. Do you want me to get rid of it ?

tarekziade: My reference is always the (recent) Python stdlib for style. It's not considered "pythonic" to have superfluous characters.  The typical example is for tuple.

```python
one = two, three, four     # do this
one = (two, three, four)     # not this
```
One exception of avoiding \ to break a line.

e.g. do:

```python
import (one, two,
            three)     # yes!!
import one, two, \
           three   # boooo!
```

So I guess my question is: what's the benefit for our code base to do this differently ?  If none, I am +1 to use Python stdlib conventions. (We should check of course that it's wlays done like that there :))

cliquet/storage/postgresql/client.py
tarekziade: ok cool!

cliquet/storage/postgresql/client.py
leplatrem: :+1: agree

tarekziade:
r+

Issue #512, opened 2015-10-28
Add minimalist load test check for testapp


leplatrem:
It happened several times that issues are discovered through Kinto load tests running on a full stack, using PostgreSQL.

I suggest we add a small load test in the Cliquet code base, ran on TravisCI, for the provided `cliquet.tests.testapp`.

This would mean:

* [ ] Add a `testapp` configuration file that uses postgresql
* [ ] Import the simulation [load test of Kinto](https://github.com/Kinto/kinto/blob/684c31c0aefc131b39430f2efcd089c940f71ac9/loadtests/loadtest/simulation.py) for the testapp `mushroom` resource
* [ ] Add a Makefile target [like this one](https://github.com/Kinto/kinto/blob/684c31c0aefc131b39430f2efcd089c940f71ac9/Makefile#L67-L73)
* [ ] Run it on Travis [like this](https://github.com/Kinto/kinto/blob/684c31c0aefc131b39430f2efcd089c940f71ac9/.travis.yml#L20)

PR #513, opened 2015-10-28
Fix pool size for batch

leplatrem:
This fixes the Kinto load tests with current master.

Since discovering issues through Kinto is unacceptable, I opened #512.

@Natim r?


Natim:
Works for me.

leplatrem:
Thanks for answering at the speed of light @Natim !

PR #514, opened 2015-10-28
Follow redirections in batch subrequests (fixes #511)

### LINKED ISSUE #511: Follow redirects in batch subrequests
If a batch subrequest returns a response with a  `30X` status code, reissue a request with same verb, headers and payload, using the URL provided in the `Location` header.

This would prevent clients to parse batch JSON response and reissue those requests.

leplatrem:
@tarekziade or @Natim r?


tarekziade:
r+

PR #517, opened 2015-10-28
Improved the 503 error message

tarekziade:
No, a 503 does not necessarely mean that the server is under high load, so the message we return here can be puzzling when it's just a backend that's down or misconfigured.

I changed the message so we're closer to the HTTP RFC for the 503


leplatrem:
r+ (*tests fail because of pypi network error*)

nit: Would be nice to squash second commit

PR #519, opened 2015-10-28
Fix crash with empty body for PATCH (fixes #477, fixes #516)

### LINKED ISSUE #477: PATCH without JSON raises a 500 error.
```
File "cliquet/resource.py", line 398, in patch
    changes = self.request.json.get('data', {})  # May patch only perms.
  File "webob/request.py", line 725, in _json_body__get
    return json.loads(self.body.decode(self.charset))
ValueError: Expected object or value
```

### LINKED ISSUE #516: Patch request to bucket/{id} returns 500 error
During testing for updating on a existing bucket as per:
http://kinto.readthedocs.org/en/latest/api/buckets.html#updating-an-existing-bucket

I created a bucket called "test_bucket" using the bucket create endpoint, then tried updating that bucket using the patch endpoint. This was done through a python script, but the equivalent curl command is:

curl --request patch https://kinto.stage.mozaws.net/v1/buckets/test_bucket --user testuser:abc123

which returns:

{"info":"https:\/\/github.com\/mozilla-services\/cliquet\/issues\/","errno":999,"message":"A programmatic error occured, developers have been informed.","code":500,"error":"Internal Server Error"}

cc @tarekziade

leplatrem:


cliquet/resource/viewset.py
```diff
@@ -179,6 +179,11 @@ def get_record_schema(self, resource_cls, method):
                                              permissions=allowed_permissions,
                                              name='permissions')
         schema.add(permissions_node)
+
+        # XXX: If Cornice wouldn't recreate the schema on the fly.
```
tarekziade: Why not an issue for this future change, rather than commiting a XXX comment ?

leplatrem: I can update the comment and link to those https://github.com/mozilla-services/cornice/labels/schema-validation

We should probably invest some time on this though :/

PR #520, opened 2015-10-28
Use StaticPool in unit tests only, not in loadtests on Travis (fixes #515)

### LINKED ISSUE #515: Prevent static pool to be used outside unit tests on Travis
Currently we check for env var `TRAVIS` to fallback to a basic static pool.

This prevents load tests to be ran properly in Kinto (on Travis).

We should restrict the test to env var `TESTS` (set by tox for example) instead.

Related #512

leplatrem:


tarekziade:
r+ as long as we have somewhere some realistic postgresql setups when doing load tests and unittest

Natim:
@tarekziade this is to test with realistic configuration for loadtest and keep StaticPool for unittest only.

Issue #449, opened 2015-09-11
Deactivate Cache Headers when Authenticated.


Natim:
When updating a collection, we should disable cache headers.

leplatrem:
Regarding single record updates (PATCH/PUT), the response body will return the last version of the records.

So naively I would say that it concerns the `GET record` more than the collection.

Could you explain why you thought of collection specifically?

Natim:
I suspect that when using KWAC to update the collection it can be difficult to sync your local records against a cached collection because each time you will call the sync you may get cached values.
Also if the ``_since`` changed inbetween, it will probably re-cache correctly but not necessarily with the right values for the public.

We may delay this until we see a problem to update cached collection but we should start thinking about problems when updating cached-collections.

PR #522, opened 2015-10-28
Add no-cache headers by default

leplatrem:
(ref Kinto/kinto#231)

### DIFF
--- a/cliquet/resource/__init__.py
+++ b/cliquet/resource/__init__.py
@@ -616,13 +616,28 @@ class BaseResource(object):
     def _add_cache_header(self, response):
         """Add Cache-Control and Expire headers, based a on a setting for the
         current resource.
+
+        Cache headers will be set with anonymous requests only.
+
+        .. note::
+
+            The ``Cache-Control: no-cache`` response header does not prevent
+            caching in client. It will indicate the client to revalidate
+            the response content on each access. The client will send a
+            conditional request to the server and check that a
+            ``304 Not modified`` is returned before serving content from cache.
         """
         resource_name = self.context.resource_name if self.context else ''
         setting_key = '%s_cache_expires_seconds' % resource_name
         collection_expires = self.request.registry.settings.get(setting_key)
         is_anonymous = self.request.prefixed_userid is None
-        if collection_expires is not None and is_anonymous:
+        if collection_expires and is_anonymous:
             response.cache_expires(seconds=int(collection_expires))
+        else:
+            # Since `Expires` response header provides an HTTP data with a
+            # resolution in seconds, do not use Pyramid `cache_expires()` in
+            # order to omit it.
+            response.cache_control.no_cache = True
 
     def _raise_400_if_invalid_id(self, record_id):
         """Raise 400 if specified record id does not match the format excepted
--- a/cliquet/tests/resource/test_views.py
+++ b/cliquet/tests/resource/test_views.py
@@ -611,8 +611,9 @@ class CacheControlTest(BaseWebTest, unittest.TestCase):
     def get_app_settings(self, extras=None):
         settings = super(CacheControlTest, self).get_app_settings(extras)
         settings['toadstool_cache_expires_seconds'] = 3600
-        settings['psilo_cache_expires_seconds'] = 0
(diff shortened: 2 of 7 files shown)


cliquet/resource/__init__.py
```diff
@@ -611,6 +611,11 @@ def _add_cache_header(self, response):
         collection_expires = self.request.registry.settings.get(setting_key)
         if collection_expires is not None:
             response.cache_expires(seconds=int(collection_expires))
+        else:
+            # Since Expires has resolution in seconds, do not use cache_expires
+            # in order to omit it.
+            response.cache_control.no_cache = True
+            response.pragma = 'no-cache'
```
michielbdejong: I don't think `pragma` is a response header, even in HTTP/1.0 it was only a request header, right? See http://stackoverflow.com/a/15050018/680454

cliquet/resource/__init__.py
```diff
@@ -611,6 +611,11 @@ def _add_cache_header(self, response):
         collection_expires = self.request.registry.settings.get(setting_key)
         if collection_expires is not None:
             response.cache_expires(seconds=int(collection_expires))
+        else:
+            # Since Expires has resolution in seconds, do not use cache_expires
+            # in order to omit it.
```
michielbdejong: Isn't `Expires: 0` equivalent to `Cache-Control: no-cache`? The `no-cache` directive does *not* mean don't cache (despite its name), it just means 'revalidate each time', where 'revalidate' means send a conditional request to the server and check that you get a 403 and not a 401 or 200 before serving the data from cache. So this `if / else` statement creates a situation where adding an `collection_expires` will cause the client to not revalidate. Revalidation is useful for making sure that each client receives updates to the resource contents asap, but also for revoking access. So setting `collection_expires` to a date in the future now has the side-effect that you cannot revoke access to the resource until that date.

michielbdejong: Might be useful to add a comment about "`no-cache` directive does not mean don't cache (despite its name)", since this is so confusing.

cliquet/resource/__init__.py
leplatrem: Ah, oh. To be honest, I just had a look at the underlying http library of the Pyramid framework, and copied this :) https://github.com/Pylons/webob/blob/master/webob/response.py#L953

I will remove it, and update the tests, thanks!

cliquet/resource/__init__.py
leplatrem: Thanks for clarifying this pitfall!

I think the confusion comes also from the fact that theses headers can be used by a reversed-proxy that will cache the responses on the server-side.

One aspect of the problem you described will be covered by #449 

We'll have to clean and consolidate the specifications (i.e. test suite) about all this.

Thanks for reviewing thoroughly! <3

leplatrem:
@michielbdejong I updated the PR with your suggestion (getting rid of Pragma response header, and rebasing on the cache restricted to anonymous requests)

@Natim any feedback ?

cliquet/tests/resource/test_views.py
```diff
@@ -612,15 +613,21 @@ def test_cache_control_headers_are_set_if_anonymous(self):
...
-        self.assertIn('Pragma', resp.headers)
+        # Check that not set by Pyramid.cache_expires()
+        self.assertNotIn('Expires', resp.headers)
+        self.assertNotIn('Pragma', resp.headers)
+
+    def test_cache_control_provides_no_cache_by_default(self):
+        resp = self.app.get('/moistures')
+        self.assertIn('no-cache', resp.headers['Cache-Control'])
+        self.assertNotIn('Expires', resp.headers)
+        self.assertNotIn('Pragma', resp.headers)
```
Natim: I think we can remove this, since it is not a feature to not have them.

Natim:
It looks good to me apart from the Pragma thing.

leplatrem:
> apart from the Pragma thing.

Do you think we should put some Pragma response header ?

cliquet/tests/resource/test_views.py
leplatrem: Ok got it sorry for the noise

PR #523, opened 2015-10-28
Set cache headers only when anonymous (fixes #449)

### LINKED ISSUE #449: Deactivate Cache Headers when Authenticated.

leplatrem:


tarekziade:
r+

Issue #396, opened 2015-07-17
Document what to do on API changes


Natim:
Following #391 discussion, we should write down the process for an API change. (e.g. doc impact, deprecation process, etc.)

leplatrem:
Side note: we've just merged #324 which uses a constant for the current protocol version

tarekziade:
I think we have a good pattern now : the doc has an API page per major version.

@Natim if you feel we miss something, please start a fresh discussion on the other tracker.

PR #524, opened 2015-10-29
Extensible user info in hello view

leplatrem:
Ref https://github.com/Kinto/kinto/pull/233

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -7,6 +7,11 @@ This document describes changes between each past release.
 2.10.0 (unreleased)
 -------------------
 
+**Protocol**
+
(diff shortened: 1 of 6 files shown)


CHANGELOG.rst
```diff
@@ -7,6 +7,11 @@ This document describes changes between each past release.
 2.10.0 (unreleased)
```
Natim: Then it should probably become 3.0.0 if we changed the protocol.

Natim:
It looks good to me with nit.

CHANGELOG.rst
leplatrem: It makes sense, also because of the switch to sqlalchemy

leplatrem:
Related https://github.com/mozilla-services/cliquet/issues/396

PR #525, opened 2015-10-29
Add a read_only settings that deactivate write endpoint methods.

Natim:
Refs Kinto/kinto#235

 - [x] Test that migrate in read_only mode display warnings for readonly backends.
 - [x] Test that all write resource endpoint are deactivated
 - [x] Test that the read_only public settings is set accordingly
 - [x] Test that the heartbeat continue to works in ``read_only`` mode.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -22,7 +22,7 @@ This document describes changes between each past release.
   ``*_pool_recycle``, ``*_pool_timeout`` to control connections pool
   behaviour.
 - Follow redirections in batch subrequests (fixes #511)
-- Add a ``read_only`` setting to run the service in read-only mode. (#525)
+- Add a ``readonly`` setting to run the service in read-only mode. (#525)
 
 **Internal changes**
 
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -22,6 +22,7 @@ This document describes changes between each past release.
   ``*_pool_recycle``, ``*_pool_timeout`` to control connections pool
   behaviour.
 - Follow redirections in batch subrequests (fixes #511)
+- Add a ``read_only`` setting to run the service in read-only mode. (#525)
 
 **Internal changes**
 
--- a/Makefile
+++ b/Makefile
@@ -35,6 +35,7 @@ build-requirements:
 	/tmp/cliquet/bin/pip freeze | tail -n +2 > requirements.txt
 
 tests-once: install-dev
+	@rm -fr .coverage
 	$(VENV)/bin/nosetests -s --with-mocha-reporter --with-coverage --cover-min-percentage=100 --cover-package=cliquet
 
 tests:
--- a/cliquet/errors.py
+++ b/cliquet/errors.py
@@ -72,10 +72,6 @@ ERRORS = Enum(
 """
 
 
-class ReadOnlyError(Warning):
-    pass
-
-
 def http_error(httpexception, errno=None,
                code=None, error=None, message=None, info=None, details=None):
     """Return a JSON formated response matching the error protocol.
--- a/cliquet/errors.py
+++ b/cliquet/errors.py
@@ -72,6 +72,10 @@ ERRORS = Enum(
 """
 
 
+class ReadOnlyError(Warning):
+    pass
+
+
 def http_error(httpexception, errno=None,
                code=None, error=None, message=None, info=None, details=None):
     """Return a JSON formated response matching the error protocol.
--- a/cliquet/permission/postgresql/__init__.py
+++ b/cliquet/permission/postgresql/__init__.py
@@ -55,15 +55,16 @@ class Permission(PermissionBase):
 
     :noindex:
     """
-    def __init__(self, client, *args, **kwargs):
+    def __init__(self, client, read_only=False, *args, **kwargs):
         super(Permission, self).__init__(*args, **kwargs)
         self.client = client
+        self.read_only = read_only
 
     def initialize_schema(self):
         # Create schema
         here = os.path.abspath(os.path.dirname(__file__))
         schema = open(os.path.join(here, 'schema.sql')).read()
-        with self.client.connect() as conn:
+        with self.client.connect(readonly=self.read_only) as conn:
             conn.execute(schema)
         logger.info('Created PostgreSQL permission tables')
 
@@ -72,7 +73,7 @@ class Permission(PermissionBase):
         DELETE FROM user_principals;
         DELETE FROM access_control_entries;
         """
-        with self.client.connect() as conn:
+        with self.client.connect(readonly=self.read_only) as conn:
             conn.execute(query)
         logger.debug('Flushed PostgreSQL permission tables')
 
@@ -86,7 +87,7 @@ class Permission(PermissionBase):
             WHERE user_id = %(user_id)s
               AND principal = %(principal)s
         );"""
-        with self.client.connect() as conn:
+        with self.client.connect(readonly=self.read_only) as conn:
             conn.execute(query, dict(user_id=user_id, principal=principal))
 
     def remove_user_principal(self, user_id, principal):
@@ -94,7 +95,7 @@ class Permission(PermissionBase):
         DELETE FROM user_principals
          WHERE user_id = %(user_id)s
            AND principal = %(principal)s;"""
-        with self.client.connect() as conn:
+        with self.client.connect(readonly=self.read_only) as conn:
             conn.execute(query, dict(user_id=user_id, principal=principal))
 
     def user_principals(self, user_id):
@@ -102,7 +103,7 @@ class Permission(PermissionBase):
         SELECT principal
           FROM user_principals
          WHERE user_id = %(user_id)s;"""
-        with self.client.connect() as conn:
+        with self.client.connect(readonly=self.read_only) as conn:
             result = conn.execute(query, dict(user_id=user_id))
             results = result.fetchall()
         return set([r['principal'] for r in results])
@@ -118,7 +119,7 @@ class Permission(PermissionBase):
                AND permission = %(permission)s
                AND principal = %(principal)s
         );"""
-        with self.client.connect() as conn:
+        with self.client.connect(readonly=self.read_only) as conn:
             conn.execute(query, dict(object_id=object_id,
                                      permission=permission,
                                      principal=principal))
@@ -129,7 +130,7 @@ class Permission(PermissionBase):
          WHERE object_id = %(object_id)s
            AND permission = %(permission)s
            AND principal = %(principal)s;"""
-        with self.client.connect() as conn:
+        with self.client.connect(readonly=self.read_only) as conn:
             conn.execute(query, dict(object_id=object_id,
                                      permission=permission,
                                      principal=principal))
@@ -140,7 +141,7 @@ class Permission(PermissionBase):
           FROM access_control_entries
          WHERE object_id = %(object_id)s
            AND permission = %(permission)s;"""
-        with self.client.connect() as conn:
+        with self.client.connect(readonly=self.read_only) as conn:
             result = conn.execute(query, dict(object_id=object_id,
                                               permission=permission))
             results = result.fetchall()
@@ -166,7 +167,7 @@ class Permission(PermissionBase):
           FROM required_perms JOIN access_control_entries
             ON (object_id = column1 AND permission = column2);
         """ % perms_values
-        with self.client.connect() as conn:
+        with self.client.connect(readonly=self.read_only) as conn:
             result = conn.execute(query)
             results = result.fetchall()
         return set([r['principal'] for r in results])
@@ -209,7 +210,7 @@ class Permission(PermissionBase):
         """ % dict(perms=perms_values,
                    principals=principals_values)
 
-        with self.client.connect() as conn:
+        with self.client.connect(readonly=self.read_only) as conn:
             result = conn.execute(query, placeholders)
             results = result.fetchall()
         return set([r['object_id'] for r in results])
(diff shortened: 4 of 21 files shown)


cliquet/scripts/cliquet.py
```diff
@@ -15,6 +17,12 @@ def deprecated_init(env):
 
 def init_schema(env):
     registry = env['registry']
+    settings = registry.settings
+
+    if settings['read_only']:
+        message = 'You cannot migrate the database with a read_only setup.'
+        warnings.warn(message, ReadOnlyError)
+        return
```
Natim: We may want to return an error code there.

cliquet/errors.py
```diff
@@ -72,6 +72,10 @@
 """
 
 
+class ReadOnlyError(Warning):
```
tarekziade: Since you are using this as a warning, and you inherit from Warning, I would call it ReadOnlyWarning instead

cliquet/scripts/cliquet.py
```diff
@@ -15,6 +17,12 @@ def deprecated_init(env):
 
 def init_schema(env):
     registry = env['registry']
+    settings = registry.settings
+
+    if settings.get('read_only', False):
+        message = 'You cannot migrate the database with a read_only setup.'
+        warnings.warn(message, ReadOnlyError)
```
tarekziade: Warnings should be used just to warn about something, and let us run python in a mode where a warning is turned into an exception.

Here, you are doing a warning and exiting no matter what. If this is the only use case for ReadOnlyError, you should turn it into an Exception and simply raise it. And remove the explicit sys.exit(1) code, because raising an error exits with 1.

cliquet/errors.py
tarekziade: Per my comment below, I think you want a plain exception here.

leplatrem:
I realize we had not the same idea in mind :]

Could you please detail what motivated you to do it in Cliquet rather that in Kinto ?

In my opinion: 

* schema migration should not be ran on a read-only instance, it sounds obvious (and `make serve` is not used in production anyway)
* the `kinto.readonly_mode` is just an alias for a set of settings to disable endpoints
* in case of read only, the storage heartbeat can be replaced from Kinto init by a simpler one
* the readonly can be provided in kinto's hello view (if ever necessary)
* with the chosen strategy here, the readonly feature is spread among cliquet and kinto (*because implicit creation in default bucket will also have to be disabled there*)

I'm sure you had a good reason I don't see here :)

cliquet/permission/postgresql/__init__.py
```diff
@@ -55,15 +55,16 @@ class Permission(PermissionBase):
...
         super(Permission, self).__init__(*args, **kwargs)
         self.client = client
+        self.read_only = read_only
 
     def initialize_schema(self):
         # Create schema
         here = os.path.abspath(os.path.dirname(__file__))
         schema = open(os.path.join(here, 'schema.sql')).read()
-        with self.client.connect() as conn:
+        with self.client.connect(readonly=self.read_only) as conn:
```
leplatrem: Why do you pass `readonly=True` if it is to execute a schema ?

cliquet_docs/api/utilities.rst
```diff
@@ -15,6 +15,7 @@ The returned value is a JSON mapping containing:
 - ``documentation``: The URL to the service documentation. (this document!)
 - ``settings``: a mapping with the values of relevant public settings for clients
     - ``batch_max_requests``: Number of requests that can be made in a batch request.
+    - ``read_only``: Tells client if the service is in read_only mode.
```
leplatrem: nit: `Tells client if` is not necessary here

cliquet/scripts/cliquet.py
Natim: Yes you are totally right, also raising an exception add a traceback where I would like to just write down a message, maybe printing to stderr and exiting would be enough, what do you think?

CHANGELOG.rst
```diff
@@ -17,6 +17,7 @@ This document describes changes between each past release.
   ``*_pool_recycle``, ``*_pool_timeout`` to control connections pool
   behaviour.
 - Follow redirections in batch subrequests (fixes #511)
+- Add a ``read_only`` setting to run the service in read-only mode. (#525)
```
leplatrem: nit: `readonly` works well in one word too. Don't we to add some suffix, like `mode` or `resources` ?

cliquet/scripts/cliquet.py
```diff
@@ -15,6 +16,12 @@ def deprecated_init(env):
 
 def init_schema(env):
     registry = env['registry']
+    settings = registry.settings
+
+    if settings.get('read_only', False):
+        message = 'You cannot migrate the database with a read_only setup.'
```
leplatrem: nit: I don't think we use `You` in the error messages.

cliquet/scripts/cliquet.py
```diff
@@ -15,6 +16,12 @@ def deprecated_init(env):
...
+        print(message, file=sys.stderr)
+        sys.exit(1)
```
leplatrem: Side note:  the cache backend requires some write access, but we expect it to have been migrated with the two other backends

cliquet_docs/api/utilities.rst
```diff
@@ -15,6 +15,7 @@ The returned value is a JSON mapping containing:
 - ``documentation``: The URL to the service documentation. (this document!)
 - ``settings``: a mapping with the values of relevant public settings for clients
     - ``batch_max_requests``: Number of requests that can be made in a batch request.
+    - ``read_only``: read_only mode service status.
```
leplatrem: What about `Only request with read operations are allowed` ?

cliquet_docs/reference/configuration.rst
```diff
@@ -73,6 +73,22 @@ following setting should be declared in the ``.ini`` file:
     cliquet.record_mushroom_patch_enabled = false
 
 
+Setting the service in read_only mode
+:::::::::::::::::::::::::::::::::::::
+
+It is also possible to deploy a cliquet service in read_only mode.
+
+Instead of listing all the endpoint, you can use the ``read_only`` settings::
```
leplatrem: of having to specify every resources endpoint and verbs, the `<read only>` setting allows to make sure the services only allows read operations.

cliquet_docs/reference/configuration.rst
```diff
@@ -73,6 +73,22 @@ following setting should be declared in the ``.ini`` file:
...
+
+    cliquet.read_only = true
+
+This will disable all resources endpoints that are not accessed with one of
+``GET``, ``OPTIONS``, or ``HEAD`` method.
+
+This settings will also make sure the heartbeat does not fails to
+write into the database.
```
leplatrem: This sentence is too related to the code IMO. 

However it could be worth mentionning here the expected setup for backend. For exemple, that the cache differs from storage and permissions...

cliquet/resource/viewset.py
```diff
@@ -151,6 +151,9 @@ def is_endpoint_enabled(self, endpoint_type, resource_name, method,
 
         Uses the settings to tell so.
         """
+        if settings.get('readonly') and method not in self.readonly_methods:
```
leplatrem: nit: use `asbool()` in case you case `readonly == "false"`

cliquet_docs/reference/configuration.rst
Natim: Does this sentence fits better:

    The readonly mode also register a readonly heartbeat for the storage and permission backend.

cliquet/scripts/cliquet.py
```diff
@@ -15,10 +17,18 @@ def deprecated_init(env):
...
+    readonly_backends = ('storage', 'permission')
+    readonly_mode = asbool(settings.get('readonly', False))
 
     for backend in ('cache', 'storage', 'permission'):
         if hasattr(registry, backend):
-            getattr(registry, backend).initialize_schema()
+            if readonly_mode and backend in readonly_backends:
+                message = ('Cannot migrate the %s backend while '
+                           'in readonly mode.' % backend)
+                warnings.warn(message, Warning)
```
tarekziade: Don't pass the base Warning class here, warn() uses a UserWarning by default. If you want a specific Warning class you need to pass a XXXWarning or create your own

Makefile
```diff
@@ -35,6 +35,7 @@ build-requirements:
 	/tmp/cliquet/bin/pip freeze | tail -n +2 > requirements.txt
 
 tests-once: install-dev
+	@rm -fr .coverage
```
leplatrem: :+1:

cliquet/resource/viewset.py
```diff
@@ -151,9 +152,13 @@ def is_endpoint_enabled(self, endpoint_type, resource_name, method,
 
         Uses the settings to tell so.
         """
+        if (asbool(settings.get('readonly')) and
+                method not in self.readonly_methods):
```
leplatrem: nit: I would recommend an intermediary variable here. Readability over brevity :)

cliquet/tests/resource/test_viewset.py
```diff
@@ -268,6 +268,42 @@ def test_is_endpoint_enabled_returns_true_if_disabled(self):
                                                  config)
         self.assertTrue(is_enabled)
 
+    def test_is_endpoint_enabled_returns_false_for_put_if_readonly(self):
+        viewset = ViewSet()
+        config = {
+            'readonly': True
+        }
```
leplatrem: micronit: would also work on a single line, no?

cliquet/tests/test_permission.py
```diff
@@ -117,7 +132,10 @@ def test_can_remove_a_principal_for_a_user(self):
     def test_can_remove_a_unexisting_principal_to_a_user(self):
         user_id = 'foo'
         principal = 'bar'
+        principal2 = 'foobar'
+        self.permission.add_user_principal(user_id, principal2)
         self.permission.remove_user_principal(user_id, principal)
+        self.permission.remove_user_principal(user_id, principal2)
```
leplatrem: Why this change ?

cliquet/tests/test_permission.py
```diff
@@ -82,12 +82,27 @@ def test_backend_error_is_raised_anywhere(self):
...
-        self.assertTrue(self.permission.ping(self.request))
+        ping = heartbeat(self.permission)
+        self.assertTrue(ping(self.request))
+
+    def test_ping_returns_false_if_unavailable_in_readonly_mode(self):
+        self.request.registry.settings['readonly'] = 'true'
+        ping = heartbeat(self.permission)
+        with mock.patch.object(self.permission, 'user_principals',
+                               side_effect=exceptions.BackendError(
+                                   "Segmentation fault.")):
```
leplatrem: nit: This indentation is not very python-friendly

Look what happens if we leave that :D https://twitter.com/tarek_ziade/status/659466245322985472

cliquet_docs/reference/configuration.rst
```diff
@@ -73,6 +73,26 @@ following setting should be declared in the ``.ini`` file:
     cliquet.record_mushroom_patch_enabled = false
 
 
+Setting the service in readonly mode
+::::::::::::::::::::::::::::::::::::
+
+It is also possible to deploy a cliquet service in readonly mode.
```
leplatrem: nit: I think we use `*Cliquet*` elsewhere

cliquet_docs/reference/configuration.rst
```diff
@@ -73,6 +73,26 @@ following setting should be declared in the ``.ini`` file:
...
+
+Instead of listing all the endpoint, you can use the ``readonly`` settings::
+
+    cliquet.readonly = true
+
+This will disable all resources endpoints that are not accessed with one of
```
leplatrem: every resources (not sure)

cliquet_docs/reference/configuration.rst
```diff
@@ -73,6 +73,26 @@ following setting should be declared in the ``.ini`` file:
...
+``GET``, ``OPTIONS``, or ``HEAD`` method.
```
leplatrem: method or HTTP verb?

cliquet_docs/reference/configuration.rst
```diff
@@ -73,6 +73,26 @@ following setting should be declared in the ``.ini`` file:
...
+
+This settings will also activate readonly heartbeat checks for the
```
leplatrem: setting (singular)

cliquet_docs/reference/configuration.rst
```diff
@@ -73,6 +73,26 @@ following setting should be declared in the ``.ini`` file:
...
+permission and the storage backend.
+
+In case you need to use the cache backend in readonly mode, it will
+still needs read-write capabilities. Do not configure it if you do not
```
leplatrem: still need (infinitive) read-write permissions/privileges?

cliquet_docs/reference/configuration.rst
```diff
@@ -73,6 +73,26 @@ following setting should be declared in the ``.ini`` file:
...
+
+    cliquet.readonly = true
+
+This will disable all resources endpoints that are not accessed with one of
+``GET``, ``OPTIONS``, or ``HEAD`` method.
+
+This settings will also activate readonly heartbeat checks for the
+permission and the storage backend.
+
+In case you need to use the cache backend in readonly mode, it will
```
leplatrem: *In case you need to use* does not fit here, nor with the rest of the paragraph

cliquet_docs/api/utilities.rst
```diff
@@ -15,6 +15,7 @@ The returned value is a JSON mapping containing:
 - ``documentation``: The URL to the service documentation. (this document!)
 - ``settings``: a mapping with the values of relevant public settings for clients
     - ``batch_max_requests``: Number of requests that can be made in a batch request.
+    - ``readonly``: Only request with read operations are allowed.
```
leplatrem: requests (plural)

leplatrem:
LGTM (only nits and english)

Please squash into a single commit though :)

cliquet/tests/test_permission.py
Natim: Because of the coverage with the memory backend :(

Issue #367, opened 2015-07-08
Collection creation returns a 200 on a deleted collection instead of a 201


Natim:
When running the loadtest, this always occurs 3 times in a raw with the memory backend.

    echo '{"data": {}}' | http PUT http://localhost:8888/v1/buckets/bucket-9c323df5-856a-427c-993c-3a38c0260289\
        --auth 'd9ad49c48db646b5a4c2511cfe855bc7:secret'
    echo '{"data": {}, "permissions": {"record:create": ["system.Authenticated"]}}' | \
        http PUT http://localhost:8888/v1/buckets/bucket-9c323df5-856a-427c-993c-3a38c0260289/collections/tasks \
        --auth 'd9ad49c48db646b5a4c2511cfe855bc7:secret'

returns a 200 instead of a 201

Same results with:

  - ``d0eff5fedff946e6802cbbd6feb22b68:secret`` and ``/buckets/bucket-77419427-128f-4d8f-b02a-8a2105fddcac/collections/tasks``
  - ``972fc9c63b214b8c90093c1e7efc5ca0:secret`` and ``/buckets/bucket-2d9388fa-82a6-4774-b0f8-6ca67b8583db/collections/tasks``

Natim:
- [ ] Update the kinto loadtests - https://github.com/mozilla-services/kinto/blob/118-test-tutorial/loadtests/loadtest/tutorial.py#L182

Natim:
I am using this to try to debug it:

```
diff --git a/kinto/views/buckets.py b/kinto/views/buckets.py
index b0a7371..6dd6921 100644
--- a/kinto/views/buckets.py
+++ b/kinto/views/buckets.py
@@ -37,6 +37,8 @@ def create_collection(request, bucket_id):
         collection_id = subpath.split('/')[2]
         collection_put = (request.method.lower() == 'put' and
                           request.path.endswith(collection_id))
+        if request.method.lower() == "put" and collection_id in request.path and not collection_put:
+            import pdb; pdb.set_trace()
         if not collection_put:
             subrequest = build_request(request, {
                 'method': 'PUT',
diff --git a/loadtests/loadtest/tutorial.py b/loadtests/loadtest/tutorial.py
index bd72c5e..d35cdae 100644
--- a/loadtests/loadtest/tutorial.py
+++ b/loadtests/loadtest/tutorial.py
@@ -179,10 +179,9 @@ class TutorialLoadTestMixin(object):
             auth=self.auth,
[...]

Issue #365, opened 2015-07-08
``If-None-Match: *`` should not raise a 412 on a deleted record.


Natim:
Today, we are using ``If-None-Match: *`` to create the record if it doesn't exists yet.

The thing is, if the record have been deleted, then ``If-None-Match: *`` fails because the record exists as deleted.

It might be a feature also it means we should describe a way to create a record if it doesn't exists or have been deleted.

Natim:
I think we should not consider deleted for the specific case of ``If-None-Match: *``

Natim:
- Notification for deleted object makes sense :+1: 
 - Conflict handling with deleted items makes sense :+1: 
 - Conflict handling on creation with deleted items doesn't makes sense. :-1:

Natim:
@michielbdejong do you have any insights for us here?

leplatrem:
This was described in https://github.com/mozilla-services/cliquet/issues/249

michielbdejong:
> Conflict handling on creation with deleted items doesn't makes sense.

I agree. An example of this situation would be:
* the resource does not exist
* client1 successfully creates the record, then successfully deletes it again, before client2 discovers it ever existed.
* client2 "thinks" the resource never existed, and tries to create it, possibly with the same content as what client1 put into it, possibly with different content.
* I agree client2's action should succeed in this case.

A way to see it is that the versioning of a resource should always be a linear step-by-step path, so that there is only one clear sequence of versions the resource goes through on the server, from its creation to its deletion. But once the resource has been deleted, this chain is forgotten by the server and it's back to its default state of non-existence.

At least that's the way we did it in remoteStorage.

The reasoning in #249 is different because that's based on modified-timestamps, not versions. There tombstones made more sense, maybe. But if you just look at ETags, then my intuition is to consider the lifespan of a resource from creation to deletion, and once it's gone, its lifespan is over and it can be forgotten.

Natim:
Here as the steps to reproduce:

```
$ http PUT https://kinto.dev.mozaws.net/v1/buckets/default/collections/tasks/records/9055850d-3038-4979-b1dd-0e5084fbc0f3 --auth natim:
$ http DELETE https://kinto.dev.mozaws.net/v1/buckets/default/collections/tasks/records/9055850d-3038-4979-b1dd-0e5084fbc0f3 --auth natim:
[...]

PR #531, opened 2015-10-30
Fix PUT not using create() method in storage backend when tombstone exists (fixes #530)

### LINKED ISSUE #530: PUT does not use create() on backend when a tombstone exists
And thus revamps bug #518 (and its fix #521)

leplatrem:
Fixes the api-test as expected :)

@Natim r?


leplatrem:
Possibly related #367 #365

cliquet/tests/resource/test_views.py
```diff
@@ -40,7 +40,7 @@ def test_record_operations_are_authorized_if_authenticated(self):
         self.app.get(record_url, headers=self.headers, status=200)
         self.app.patch_json(record_url, body, headers=self.headers, status=200)
         self.app.delete(record_url, headers=self.headers, status=200)
-        self.app.put_json(record_url, body, headers=self.headers, status=200)
+        self.app.put_json(record_url, body, headers=self.headers, status=201)
```
Natim: I love that thanks a lot :+1:

PR #533, opened 2015-11-02
Add simple loadtest on PostgreSQL (fixes #512)

### LINKED ISSUE #512: Add minimalist load test check for testapp

leplatrem:
See #512 

* [x] Add a `testapp` configuration file that uses postgresql
* [x] Import the simulation [load test of Kinto](https://github.com/Kinto/kinto/blob/684c31c0aefc131b39430f2efcd089c940f71ac9/loadtests/loadtest/simulation.py) for the testapp `psilos` resource (protected resource)
* [x] Add a Makefile target [like this one](https://github.com/Kinto/kinto/blob/684c31c0aefc131b39430f2efcd089c940f71ac9/Makefile#L67-L73)
* [x] Run it on Travis [like this](https://github.com/Kinto/kinto/blob/684c31c0aefc131b39430f2efcd089c940f71ac9/.travis.yml#L20)

@almet r?


loadtests/simulation.py
```diff
@@ -0,0 +1,318 @@
...
+import os
+import random
+import uuid
+
+from konfig import Config
+from loads.case import TestCase
+from requests.auth import HTTPBasicAuth
+
+
+ACTIONS_FREQUENCIES = [
```
Natim: I am wondering if we could write it in a way it does try each of them at least once so that we can be confident about the stability of the loadtest.

I'd like to run it once as a smoketest here while making sure each call was tried at least once.

leplatrem: I just took it from kinto and readinglist. Would it be ok to improve it in another iteration?

Natim: Yes probably.

leplatrem:
@almet r?

cliquet/tests/testapp/__init__.py
```diff
@@ -14,4 +14,6 @@ def main(settings=None, config=None, *args, **additional_settings):
         config = Configurator(settings=settings)
     cliquet.initialize(config, version='0.0.1')
     config.include(includeme)
-    return config.make_wsgi_app()
+    app = config.make_wsgi_app()
+    # Install middleware (idempotent if disabled)
```
almet: I don't understand this comment: what is idempotent if disabled? Can you enlighten me? (and ideally update the comment with an easier to grasp sentence)

cliquet/tests/testapp/views.py
```diff
@@ -16,6 +16,7 @@ class Toadstool(resource.ProtectedResource):
     mapping = MushroomSchema()
 
     def get_parent_id(self, request):
+        # Shared collection.
```
almet: Same here: when reading the source code, I don't understand why this comment lies here. Since it's an application that could be viewed as an example application, I would rather add some more details in here (yes, it's off-topic with this pull request)

cliquet_docs/contributing.rst
```diff
@@ -84,6 +84,41 @@ to speed-up the execution:
     nosetests -s --with-mocha-reporter cliquet.tests.test_views_hello:HelloViewTest.test_returns_info_about_url_and_version
 
 
+Load tests
+----------
+
+A load test is provided in order to run end-to-end tests on *Cliquet* through a sample application,
+or prevent regressions in terms of performance.
```
almet: nit: replace "or" with "and".

cliquet_docs/contributing.rst
```diff
@@ -84,6 +84,41 @@ to speed-up the execution:
...
+
+The following ``make`` command will check briefly the overall sanity of the API,
+by running a server and running a very few random HTTP requests on it.
+
+::
+
+    make loadtest-check-simulation
+
+It is possible to customize this load test, by focusing on a single action,
```
almet: nit: I believe that we are trying to use the english form elsewhere in the documentation (for the "z" versus "s"). This should be "customise", then.

cliquet_docs/contributing.rst
```diff
@@ -84,6 +84,41 @@ to speed-up the execution:
...
+First, run the test application manually in a terminal: ::
+
+    cd loadtests/
+    pserve loadtests/server.ini
+
+And then, run the test suite against it: ::
+
+    SERVER_URL=http://localhost:8888 make test -e
+
+With a specific action: ::
```
almet: condider replacing with "To run a specific action, specify it with"

loadtests/server.ini
```diff
@@ -0,0 +1,22 @@
+[app:main]
```
almet: should we name this "testapp.ini" like it's called elsewhere?

loadtests/simulation.py
```diff
@@ -0,0 +1,318 @@
...
+
+    def collection_url(self, prefix=True):
+        if prefix:
+            return self.api_url(self.collection)
+        return '/' + self.collection
+
+    def record_url(self, record_id, prefix=True):
+        return self.collection_url(prefix) + '/%s' % record_id
+
+    def init_record(self, *args, **kwargs):
```
almet: not sure why this is in a separate method: it's not called by anything else than the constructor.

loadtests/simulation.py
```diff
@@ -0,0 +1,318 @@
...
+        """Choose a random action among available, if not frequent enough,
+        try again recursively.
+
+        :note:
+
+            This method is called as many times as number of hits.
+        """
+        action, percentage = random.choice(ACTIONS_FREQUENCIES)
+
+        forced_action = os.getenv('LOAD_ACTION')
```
almet: I don't think this will work when being run in a cluster (I don't think the env is passed along). Should we read from the configuration instead?

almet: (Or consider that this option is only available when running locally)

almet:
After a review, this looks like a correct way to exercise some load locally. It might be worth it to try running it in a cluster (even if no server is running with the demoapp) to check that all the configuration files are properly included; I've had issues with this in the past.

r+ with comments. Most of them are just quick remarks, I'll let you chose what you want to do with them.

cliquet/tests/testapp/__init__.py
Natim: It means that if install_middleware actually doesn't install any middleware it will just return the app.

cliquet/tests/testapp/views.py
Natim: Related to #550

loadtests/simulation.py
Natim: It is the exact same simulation test that we have in Kinto we are using this functional test locally only to validate the overall API wasn't broken with cliquet modification.

almet: OK. I have the same concerns for the kinto loadtest then ;)

Natim: Why would you need this option for a remote loadtest?

PR #324, opened 2015-11-02
Expose the version protocol in the / endpoint

Natim:
Currently, the version exposed in the / endpoint isn't the protocol version but the version of the application. As such, it's important to also provide the protocol version alongside the other information.

Protocol version could be the major version of cliquet.

### DIFF
--- a/cliquet/tests/test_views_hello.py
+++ b/cliquet/tests/test_views_hello.py
@@ -8,7 +8,7 @@ class HelloViewTest(BaseWebTest, unittest.TestCase):
     def test_returns_info_about_url_and_version(self):
         response = self.app.get('/')
         self.assertEqual(response.json['version'], "0.0.1")
-        self.assertEqual(response.json['api_version'], "2")
+        self.assertEqual(response.json['protocol'], "v2")
         self.assertEqual(response.json['url'], 'http://localhost/v0/')
         self.assertEqual(response.json['hello'], 'myapp')
         self.assertEqual(response.json['documentation'],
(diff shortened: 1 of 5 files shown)


cliquet/tests/test_views_hello.py
```diff
@@ -8,6 +8,7 @@ class HelloViewTest(BaseWebTest, unittest.TestCase):
     def test_returns_info_about_url_and_version(self):
         response = self.app.get('/')
         self.assertEqual(response.json['version'], "0.0.1")
+        self.assertEqual(response.json['api_version'], "2")
```
Natim: I find it a bit disturbing to have ``v1`` in the kinto url and ``api_version: 2`` in its hello page.
Shall we call it ``cliquet_api_version``?

cliquet/views/hello.py
```diff
@@ -13,6 +13,7 @@ def get_hello(request):
     data = dict(
         hello=project_name,
         version=settings['project_version'],
+        protocol_version=API_VERSION,
```
leplatrem: That is exactly what I had in mind. What if we break the Cliquet python API, but keep the HTTP API identical? I think it could a separated constant, no?

cliquet_docs/api/utilities.rst
```diff
@@ -9,6 +9,7 @@ GET /
 The returned value is a JSON mapping containing:
 
 - ``hello``: the name of the service (e.g. ``"reading list"``)
+- ``protocol_version``: the cliquet protocol version (``"2"``)
 - ``version``: complete version (``"X.Y.Z"``)
```
leplatrem: it would be worth here to specify *complete application/project version*

cliquet/views/hello.py
Natim: Would you like me to rename ``API_VERSION`` in ``PYTHON_API_VERSION`` and ``HTTP_API_VERSION``? Or should we wait until the aren't the same?

leplatrem: I would be in favor of having two constants, it will clarify to the reader that they can differ. That's a nitpick though :)

PR #537, opened 2015-11-02
Upgrade ujson to 1.35

Natim:
Refs https://github.com/mozilla-services/readinglist/issues/224

### DIFF
--- a/cliquet/initialization.py
+++ b/cliquet/initialization.py
@@ -52,7 +52,7 @@ def setup_json_serializer(config):
     requests.models.json = utils.json
 
     # Override json renderer using ujson
-    renderer = JSONRenderer(serializer=lambda v, **kw: utils.json.dumps(v))
+    renderer = JSONRenderer(serializer=utils.json_serializer)
     config.add_renderer('json', renderer)
 
 
--- a/cliquet/utils.py
+++ b/cliquet/utils.py
@@ -9,11 +9,18 @@ import warnings
 from base64 import b64decode, b64encode
 from binascii import hexlify
 from six.moves.urllib import parse as urlparse
-# ujson is not installable with pypy
+
+# ultrajson is not installable with pypy
 try:
-    import ujson as json  # NOQA
+    import ultrajson as json  # NOQA
+
+    def json_serializer(v, **kw):
+        return json.dumps(v, escape_forward_slashes=False)
+
 except ImportError:  # pragma: no cover
     import json  # NOQA
+
+    json_serializer = json.dumps
 
 try:
     # Register psycopg2cffi as psycopg2
--- a/cliquet/utils.py
+++ b/cliquet/utils.py
@@ -11,7 +11,7 @@ from binascii import hexlify
 from six.moves.urllib import parse as urlparse
 
 # ujson is not installable with pypy
-try:
+try:  # pragma: no cover
     import ujson as json  # NOQA
 
     def json_serializer(v, **kw):
--- a/cliquet/utils.py
+++ b/cliquet/utils.py
@@ -10,9 +10,9 @@ from base64 import b64decode, b64encode
 from binascii import hexlify
 from six.moves.urllib import parse as urlparse
 
-# ultrajson is not installable with pypy
+# ujson is not installable with pypy
 try:
-    import ultrajson as json  # NOQA
+    import ujson as json  # NOQA
 
     def json_serializer(v, **kw):
         return json.dumps(v, escape_forward_slashes=False)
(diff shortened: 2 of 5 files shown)


Natim:
AFAIK it doesn't seems to fix the problem.

Natim:
Or maybe it is because of missing ``, escape_forward_slashes=False`` ?

Natim:
I added the option but I have no idea if it will actually works with normal json.

Natim:
```TypeError: __init__() got an unexpected keyword argument 'escape_forward_slashes'; uid=None; errno=None; agent=curl/7.35.0; authn_type=None```

@leplatrem any idea of how to handle this?

File: cliquet/initialization.py [hunk 1]
```diff
@@ -50,7 +50,8 @@ def setup_json_serializer(config):
     requests.models.json = utils.json
 
     # Override json renderer using ujson
-    renderer = JSONRenderer(serializer=lambda v, **kw: utils.json.dumps(v))
+    renderer = JSONRenderer(serializer=lambda v, **kw: utils.json.dumps(
+        v, escape_forward_slashes=False))
```
Natim: The escape_forward_slashes only works with ultrajson.

leplatrem:
See https://github.com/esnme/ultrajson/issues/185

Natim:
@leplatrem well it doesn't really change the matter on this PR but ok I understand that.

cliquet/initialization.py
leplatrem: The authors could not publish the version with the escape option on Pypi yet

Natim: > The authors could not publish the version with the escape option on Pypi yet

The ujson people couldn't but a fork ultrajson was made available with it.

leplatrem: > The escape_forward_slashes only works with ultrajson.

What if we do `renderer = JSONRenderer(serializer=utils.json_serializer)` ?

Natim:
With my last commit I've got two test failing on this kind of JSON:

```u'{"responses":[{"status":201,"path":"/v0/mushrooms/e93e0961-0e09-4d8f-9ee0-8b05cdf18f41","body":{"data":{"last_modified":1446547536607,"name":"Trompette de la mort","id":"e93e0961-0e09-4d8f-9ee0-8b05cdf18f41"}},"headers":{"Last-Modified":"Tue, 03 Nov 2015 10:45:36 GMT","Content-Length":"114","ETag":""1446547536607"","Content-Type":"application/json; charset=UTF-8","Access-Control-Expose-Headers":"Retry-After, Content-Length, Alert, Backoff"}}]}'```

File: cliquet/utils.py [hunk 2]
```diff
@@ -9,12 +9,19 @@
...
+    import ultrajson as json  # NOQA
+
+    def json_serializer(v, **kw):
+        return json.dumps(v, escape_forward_slashes=False)
+
 except ImportError:  # pragma: no cover
     import json  # NOQA
 
+    def json_serializer(v, **kw):
+        return json.dumps(v)
```
leplatrem: nit: I'm not 100% sure that it is better, but you can do `json_serializer = json.dumps`

cliquet/utils.py
```diff
@@ -9,12 +9,19 @@
...
 from six.moves.urllib import parse as urlparse
 
-# ujson is not installable with pypy
+# ultrajson is not installable with pypy
 try:
-    import ujson as json  # NOQA
+    import ultrajson as json  # NOQA
+
+    def json_serializer(v, **kw):
+        return json.dumps(v, escape_forward_slashes=False)
```
leplatrem: nit: I'm not 100% sure that it is better, but you could do `json_serializer = functools.partial(json.dumps, escape_forward_slashes=False)`

Natim: Telepathy for the win :+1:

Natim:
Well it is even worse actually :(

leplatrem:
I tried to update this, but found a problem in the JSON serialization of batch responses:

capture d ecran de 2015-11-03 17 19 12

Natim:
> It means that " are not escaped within strings :/

Yes it is what I also noticed, do you know if it is a bug with your ultrajson fix?

leplatrem:
I guess it is a bug, look : 

```
>>> import ujson
>>> ujson.dumps('"')
'"\\""'

>>> import json
>>> json.dumps('"')
'"\\""'

>>> ujson.dumps('"', escape_forward_slashes=False)
'"""'
```

leplatrem:
I updated this. Needs rebase.

Natim:
Let me rebase then :+1:

leplatrem:
@Natim final r+ ?

Natim:
GG mat r+

PR #539, opened 2015-11-02
Make sure read endpoints are activated.

Natim:
r? @leplatrem

### DIFF
--- a/cliquet/resource/viewset.py
+++ b/cliquet/resource/viewset.py
@@ -153,7 +153,9 @@ class ViewSet(object):
         Uses the settings to tell so.
         """
         readonly_enabled = asbool(settings.get('readonly'))
-        if readonly_enabled and method not in self.readonly_methods:
+        readonly_method = method.lower() in [m.lower() for m in
+                                             self.readonly_methods]
+        if readonly_enabled and not readonly_method:
             return False
 
         setting_enabled = '%s_%s_%s_enabled' % (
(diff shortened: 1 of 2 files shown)


leplatrem:
r+

cliquet/resource/viewset.py
```diff
@@ -153,7 +153,9 @@ def is_endpoint_enabled(self, endpoint_type, resource_name, method,
         Uses the settings to tell so.
         """
         readonly_enabled = asbool(settings.get('readonly'))
-        if readonly_enabled and method not in self.readonly_methods:
+        readonly_method = method.lower() in [m.lower() for m in
```
tarekziade: nitpick: shouldn't we normalize the other way ? e.g. the request method is always uppercase and readonly_methods could be normalize on startup

Natim: Actually the bug was that we though that the method was always uppercase, actually it is more always lowercase.

cliquet/resource/viewset.py
```diff
@@ -153,7 +153,9 @@ def is_endpoint_enabled(self, endpoint_type, resource_name, method,
...
+                                             self.readonly_methods]
+        if readonly_enabled and not readonly_method:
```
ametaireau: why the "not" here? I would have naively expected it to return False when the current method is readonly.

e.g.

```python
if readonly_enabled and readonly_method:
    return False
```

ametaireau: Please disregard :-)

PR #546, opened 2015-11-04
Make sure the proxies caches with regards to the Authorization header.

Natim:
Refs https://github.com/Kinto/kinto/pull/233
Refs https://github.com/mozilla-services/syncto/pull/62

review? @almet 
feedback? @michielbdejong @leplatrem

### DIFF
--- a/cliquet/initialization.py
+++ b/cliquet/initialization.py
@@ -142,6 +142,9 @@ def setup_authentication(config):
             event.request.prefixed_userid = '%s:%s' % (authn_type.lower(),
                                                        user_id)
 
+            # Add a vary Authorization header
+            event.request.response.vary = ['Authorization']
+
     config.add_subscriber(on_new_request, NewRequest)
 
 
--- a/cliquet/initialization.py
+++ b/cliquet/initialization.py
@@ -134,6 +134,9 @@ def setup_authentication(config):
     def on_new_request(event):
         authn_type = getattr(event.request, 'authn_type', None)
 
+        # Add a vary Authorization header
+        event.request.response.vary = ['Authorization']
+
         # Prefix the user id with the authn policy type name.
         user_id = event.request.authenticated_userid
 
@@ -141,9 +144,6 @@ def setup_authentication(config):
         if user_id and authn_type:
             event.request.prefixed_userid = '%s:%s' % (authn_type.lower(),
                                                        user_id)
-
-            # Add a vary Authorization header
-            event.request.response.vary = ['Authorization']
 
     config.add_subscriber(on_new_request, NewRequest)
 
--- a/cliquet/tests/resource/test_views.py
+++ b/cliquet/tests/resource/test_views.py
@@ -638,6 +638,25 @@ class CacheControlTest(BaseWebTest, unittest.TestCase):
         self.assertNotIn('Expires', resp.headers)
 
 
+class VaryHeaderTest(BaseWebTest, unittest.TestCase):
+    collection_url = '/toadstools'
+
+    def get_app_settings(self, extras=None):
+        settings = super(VaryHeaderTest, self).get_app_settings(extras)
+        settings['toadstool_read_principals'] = 'system.Everyone'
+        settings['moisture_read_principals'] = 'system.Everyone'
+        return settings
+
+    def test_vary_header_is_not_set_if_anonymous(self):
+        resp = self.app.get(self.collection_url)
+        self.assertNotIn('Vary', resp.headers)
+
+    def test_vary_header_is_set_if_authenticated(self):
+        resp = self.app.get(self.collection_url, headers=self.headers)
+        self.assertIn('Vary', resp.headers)
+        self.assertIn('Authorization', resp.headers['Vary'])
+
+
 class StorageErrorTest(BaseWebTest, unittest.TestCase):
     def __init__(self, *args, **kwargs):
         super(StorageErrorTest, self).__init__(*args, **kwargs)
--- a/cliquet/tests/resource/test_views.py
+++ b/cliquet/tests/resource/test_views.py
@@ -641,18 +641,35 @@ class CacheControlTest(BaseWebTest, unittest.TestCase):
 class VaryHeaderTest(BaseWebTest, unittest.TestCase):
     collection_url = '/toadstools'
 
+    def setUp(self):
+        super(VaryHeaderTest, self).setUp()
+
+        body = {'data': MINIMALIST_RECORD}
+        resp = self.app.post_json(self.collection_url, body,
+                                  headers=self.headers)
+        self.record = resp.json['data']
+        self.record_url = self.get_item_url()
+
     def get_app_settings(self, extras=None):
         settings = super(VaryHeaderTest, self).get_app_settings(extras)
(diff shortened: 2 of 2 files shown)


cliquet/initialization.py
```diff
@@ -142,6 +142,9 @@ def on_new_request(event):
             event.request.prefixed_userid = '%s:%s' % (authn_type.lower(),
                                                        user_id)
 
+            # Add a vary Authorization header
+            event.request.response.vary = ['Authorization']
```
leplatrem: /me wondering: don't we want this on resources only ?

michielbdejong:
We should probably test this with http://michielbdejong.com/kinto-buckets.html as well as with a proxy server (a local squid instance? or some CDN service?) and see if it actually works.

I'm not even sure if 'Authorization' is a valid value for a 'Vary' header, given the special role that the Authorization header already has in shared caches:
Point 2 of http://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.8 worries me a bit as well - even though 'no-cache' is different from 'must-revalidate', that statement does imply that shared caches can assume that the 'Authorization' header only influences Authorization, and does not influence the content of the resource.

Maybe we should add an 'X-User-Id' header and Vary on that? I think that could make the design a bit clearer. Feels like we're using the Authorization header in ways that caches don't expect.

cliquet/initialization.py
Natim: You are right, we need it for the /v1/ endpoint as well, let me add a test. (because of user_id and bucket_id)

Natim:
As far as I understand the Vary header works on Request headers and not Response ones, so X-User-Id won't help but I might be wrong about that.

cliquet/tests/resource/test_views.py
```diff
@@ -638,6 +638,42 @@ def test_cache_control_provides_no_cache_by_default(self):
...
+    def test_vary_header_is_not_set_if_anonymous_on_collection_url(self):
+        resp = self.app.get(self.collection_url)
+        self.assertNotIn('Vary', resp.headers)
+
+    def test_vary_header_is_set_if_authenticated_on_collection_url(self):
+        resp = self.app.get(self.collection_url, headers=self.headers)
+        self.assertIn('Vary', resp.headers)
+        self.assertIn('Authorization', resp.headers['Vary'])
+
+    def test_vary_header_is_not_set_if_anonymous_on_record_url(self):
```
leplatrem: Is this a prerequisite ? What happens if the server says `Vary: Authorization` and no authorization header is provided. I would expect it to silently compute *its cache key* using an empty value, which would make sense :)

cliquet/tests/resource/test_views.py
```diff
@@ -638,6 +638,42 @@ def test_cache_control_provides_no_cache_by_default(self):
...
+        resp = self.app.post_json(self.collection_url, body,
+                                  headers=self.headers)
+        self.record = resp.json['data']
+        self.record_url = self.get_item_url()
+
+    def get_app_settings(self, extras=None):
+        settings = super(VaryHeaderTest, self).get_app_settings(extras)
+        settings['toadstool_read_principals'] = 'system.Everyone'
+        return settings
+
```
leplatrem: Since your code adds the vary headers everywhere, I would add a specification saying «Vary is provided on hello view»

cliquet/tests/resource/test_views.py
Natim: We shall test that but I find cleaner to only add it if the Authorization header is present.

cliquet/tests/resource/test_views.py
Natim: Done here https://github.com/mozilla-services/cliquet/commit/95a408e55eb89bda6e104f4a6bef0e25898f9dea

michielbdejong:
My idea was that clients should add this header to make it uniquely defined which data they want to access, regardless of the Authorization header. Both the Kinto server and the caches could then:
* decide which data should be served (content negotiation) based on URL +  the X-User-Id header
* use Authorization only to decide whether the request is should be allowed or not.

Instead of the current situation, where Authorization and Content Negotiation are connected into one complex process (not so complex in itself, but it becomes complex due the complexities of how http caching evolved over the past decades)

Natim:
This PR doesn't seems to fix [@michielbdejong test ](https://michielbdejong.com/kinto-buckets.html) also I noticed that 304 do not have the Vary header.

cliquet/tests/resource/test_views.py
```diff
@@ -638,6 +638,51 @@ def test_cache_control_provides_no_cache_by_default(self):
...
+    def test_vary_header_is_not_set_if_anonymous_on_hello_page(self):
```
almet: Well actually it's not how it's supposed to work. The answer will vary, depending the authentication you put. If you don't add the Vary when you're anonymous, then the anonymous version will be cached and served even when you're authenticated.

cliquet/tests/resource/test_views.py
almet: See my previous answer. From what I understood, the Vary header should be added even if you're anonymous.

almet:
@michielbdejong that's an interesting point, and proposal. Actually, depending the authentication scheme you chose, it's value is not always the same for one same user accross requests. I like the idea of adding another header to identify the uniqueness of a user, and vary on it!

michielbdejong:
Ah, just thought of another situation: Suppose you want to use X509 client certificates to identify/authorize users on your Kinto server. Then there are no Authorization headers. But you could check the `X-User-Id` header (or whatever name you pick) against the client cert, and still have varying content depending on the current user.

Natim:
> I'm not even sure if 'Authorization' is a valid value for a 'Vary' header, given the special role that the Authorization header already has in shared caches.

What I understand is that it may cache the same thing once per user, maybe we should make the vary header activation a settings.

IMHO, asking clients to add a User-Id header on each request seems a lot to ask.

> Suppose you want to use X509 client certificates to identify/authorize users on your Kinto server.

Then the X509 validation will be made offload [on the nginx side](https://github.com/novapost/django-x509#nginx-configuration-for-certificate), the HTTP_SSL_CLIENT_SERIAL, HTTP_SSL_CLIENT_DN headers will be provided to the kinto server.

Depending where the proxy stands between the user and the kinto server is it may not have the User-ID information on the request (added by the SSL offload server)

With regards to how the user log in kinto and where the Proxy stands we may need to change the Vary header content.

cliquet/initialization.py
```diff
@@ -134,6 +134,9 @@ def on_policy_selected(event):
     def on_new_request(event):
         authn_type = getattr(event.request, 'authn_type', None)
 
+        # Add a vary Authorization header
+        event.request.response.vary = ['Authorization']
```
Natim: We may need to let people configure this from the settings.

almet:
The new header would be added in the responses, not requests (same for Vary).

almet:
After some quick chat, it seems that the `Vary` header won't be of any use here. If the goal is to avoid having the browser cache responses across users, then using `no-cache` is enough.

We don't actually *want* a mechanism to cache the responses, because we already have one with the Etags. Closing the issue for now.

PR #548, opened 2015-11-04
Fix notifications on protected resources

leplatrem:
@tarekziade r?

### DIFF
--- a/cliquet/resource/__init__.py
+++ b/cliquet/resource/__init__.py
@@ -1028,13 +1028,13 @@ class ProtectedResource(BaseResource):
 
         return annotated
 
-    def postprocess(self, result, action=ACTIONS.READ):
+    def postprocess(self, result, **kwargs):
         """Add ``permissions`` attribute in response body.
 
         In the protocol, it was decided that ``permissions`` would reside
         outside the ``data`` attribute.
         """
-        result = super(ProtectedResource, self).postprocess(result)
+        result = super(ProtectedResource, self).postprocess(result, **kwargs)
         if isinstance(result['data'], list):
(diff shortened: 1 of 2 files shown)


cliquet/resource/__init__.py
```diff
@@ -1028,13 +1028,13 @@ def process_record(self, new, old=None):
 
         return annotated
 
-    def postprocess(self, result, action=ACTIONS.READ):
+    def postprocess(self, result, **kwargs):
```
tarekziade: you are changing the signature from BaseResource. why ?

Natim: Shouldn't we also catch for *args in case action is not passed as a named argument?

cliquet/resource/__init__.py
```diff
@@ -1028,13 +1028,13 @@ def process_record(self, new, old=None):
...
         """Add ``permissions`` attribute in response body.
 
         In the protocol, it was decided that ``permissions`` would reside
         outside the ``data`` attribute.
         """
-        result = super(ProtectedResource, self).postprocess(result)
```
tarekziade: good catch!

tarekziade:
r+

cliquet/resource/__init__.py
leplatrem: do we want to support that?

Natim: I don't know, that's why I asked the question :)

leplatrem: :) Since I am given the absolute power to decide that, I will say no and push the big green button :D

Issue #559, opened 2015-11-09
Getting the exception when a heartbeat fails


tarekziade:
We're deploying the redis backend for the cache right now and the heartbeat reports that it fails.

But this we're swallowing the exception here: https://github.com/mozilla-services/cliquet/blob/master/cliquet/cache/__init__.py#L87

It's hard to make a diagnostic. We should at least do a logger.error() call there so we can have the traceback logged for analysis.

tarekziade:
fixed by https://github.com/mozilla-services/cliquet/commit/27c33e02b8ad72542984e6f743294b5a1390d6b5

PR #561, opened 2015-11-09
log any heartbeat exception - fixes issue #559

tarekziade:


leplatrem:
I would suggest, «ya que estamos», to do the same in storage and permission backends

cliquet/cache/__init__.py
```diff
@@ -85,6 +86,7 @@ def ping(request):
                 backend.set(_HEARTBEAT_KEY, 'alive', _HEARTBEAT_TTL_SECONDS)
             return True
         except:
+            logger.exception("Heartbeat Failure")
```
leplatrem: question: why not use the original exception?

tarekziade: logger.exception is dumping the full exception info. see https://docs.python.org/2/library/logging.html#logging.Logger.exception

What would you like to see changed ?

leplatrem: I was wondering why you didn't do `except Exception as e:  logger.exception(e)`

tarekziade: In both cases we get the same info about the error - But I like to use explicit messages because it makes it easier to grep in the logs. 

e.g. "Q: why is the heartbeat down. ?  A: let me grep "heartbeat failures" in the logs.


```python
>>> try:
...     raise Exception('pok')
... except Exception as e:
...     logging.exception(e)
...
ERROR:root:pok
Traceback (most recent call last):
  File "<stdin>", line 2, in <module>
Exception: pok

>>> try:
...     raise Exception('pok')
... except:
...     logging.exception("Heartbeat Error")
...
ERROR:root:Heartbeat Error
Traceback (most recent call last):
  File "<stdin>", line 2, in <module>
Exception: pok
```

leplatrem: Excellent :+1: ! I'm glad I insisted ;)

leplatrem:
r+ if tests pass

Issue #25, opened 2015-02-25
Timestamp uniqueness by parent_id under heavy load


leplatrem:
A DB constraint was added to guarantee uniqueness of timestamps, per user, per collection.

The current version of the ``bump_timestamp()`` SQL function does not resist to *(very very high)* load by **one user**. In other words, when more than **two requests of the same user** are handled in **less than 2 msecs**.

For example, running several instances of this script in parallel will raise integrity errors (*~4/10000 when running 3 instances on my devbox*):

```python
import psycopg2
import psycopg2.extensions

conn = psycopg2.connect(dbname='testdb', host='localhost', user='postgres', password='postgres')
cursor = conn.cursor()

nberrors = 0

for i in range(10000):
    try:
        cursor.execute("INSERT INTO records(user_id, resource_name, data) VALUES ('', '', '{}')")
        conn.commit()
        if i % 100:
            print '.',
    except psycopg2.Error as e:
        nberrors += 1
        conn.rollback()
        print e

print nberrors, 'errors'
````

tarekziade:
> when more than two requests of the same user are handled in less than 2 msecs

Can this really happen in the real world ? Also, since we are using transaction now, are we still triggering that function for every row ? or could we have the event triggered at the transaction level

leplatrem:
I posted a simplified version of this on http://stackoverflow.com/questions/34817760/how-to-generate-unique-timestamps-in-postgresql

We could do as it is done for Sync:

* add another column with a sequence
* use the sequence as fake decimals `"%s.%s" % (last_modified, sequence)` for timestamps

tarekziade:
moved to https://github.com/Kinto/kinto/issues/602

PR #563, opened 2015-11-09
Enhance error message for invalid If-Match headers

almet:


cliquet/resource/__init__.py
```diff
@@ -715,9 +715,11 @@ def _raise_412_if_modified(self, record=None):
                 assert if_match[0] == if_match[-1] == '"'
                 modified_since = int(if_match[1:-1])
             except (IndexError, AssertionError, ValueError):
+                message = ("Invalid value for If-Match. "
+                           "The value should be inside double quotes.")
```
leplatrem: should be integer between double quotes

leplatrem:
cliquet/tests/resource/test_preconditions.py:2:1: F401 'six' imported but unused

r+wc

leplatrem:
The test that fails is `test_purge_deleted_remove_with_before_remove_olders_exclusive`. This is related IMO to the race condition in #25. A solution is to slow down [those 4 operations](https://github.com/mozilla-services/cliquet/blob/master/cliquet/tests/test_storage.py#L710-L713)

PR #565, opened 2015-11-09
Fix Redis client instantiation (fixes #564)

### LINKED ISSUE #564: bad instanciation of Redis clients
We had an issue on stage with a misbehaving redis. Turns out the way we instanciate redis 2.10.3+ is wrong:

```python
class Cache(CacheBase):
    def __init__(self, *args, **kwargs):
        super(Cache, self).__init__(*args, **kwargs)
        maxconn = kwargs.pop('max_connections')
        connection_pool = redis.BlockingConnectionPool(max_connections=maxconn)
        self._client = redis.StrictRedis(connection_pool=connection_pool, **kwargs)
```

If you look at the client code, if you provide a connection_pool object, all connections options passed in StrictRedis kwargs are dropped. see  https://github.com/andymccurdy/redis-py/blob/2.10.3/redis/client.py#L445 - this is why stage absolutely wanted to use localhost....

The right code should be

```python
class Cache(CacheBase):
    def __init__(self, *args, **kwargs):
        super(Cache, self).__init__(*args, **kwargs)
        connection_pool = redis.BlockingConnectionPool(**kwargs)
        self._client = redis.StrictRedis(connection_pool=connection_pool)
```

Also, we have 6739 instead of 6379 as a default port there. doh!

Changes tried and validated on stage. Now working on a branch I'll link here.

cc @Micheletto

tarekziade:


leplatrem:
This looks good to me. I will add a small test that fails with previous code.

PR #568, opened 2015-11-10
Refactor Redis client instantiation (fixes #567)

### LINKED ISSUE #567: Add a settings property on backends
When a backend is instanciated, it gets some settings and creates a connection object with it. In some our tests we're checking the values of those settings to check that they are correctly passed along.

I propose a simple standard for all our backends to expose those settings: a "settings" read-only property that will let us read the values.

Implementation example for the redis backends:

```python

class RedisBackend:
    @property
    def settings(self):
        data = dict(self._client.connection_pool.connection_kwargs)
        data['max_connections'] = self._client.connection_pool.max_connections
        return data
```

leplatrem:
* [x] Refactor Redis client instantiation to avoid duplicated defaults
* [x] Add settings property to avoid introspection of private member in tests (fixes #567)

@tarekziade r?

### DIFF
--- a/cliquet/storage/redis.py
+++ b/cliquet/storage/redis.py
@@ -21,6 +21,24 @@ def wrap_redis_error(func):
     return wrapped
 
 
+def create_from_config(config, prefix=''):
+    """Redis client instantiation from settings.
+    """
+    settings = config.get_settings()
+    uri = settings[prefix + 'url']
+    uri = urlparse.urlparse(uri)
+    pool_size = int(settings[prefix + 'pool_size'])
+    kwargs = {
+        "max_connections": pool_size,
+        "host": uri.hostname or 'localhost',
+        "port": uri.port or 6379,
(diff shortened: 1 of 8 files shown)


cliquet/storage/redis.py
```diff
@@ -42,10 +60,13 @@ class Storage(MemoryBasedStorage):
...
-    def __init__(self, *args, **kwargs):
+    def __init__(self, client, *args, **kwargs):
         super(Storage, self).__init__(*args, **kwargs)
-        connection_pool = redis.BlockingConnectionPool(**kwargs)
-        self._client = redis.StrictRedis(connection_pool=connection_pool)
+        self._client = client
+
+    @property
+    def settings(self):
+        return self._client.connection_pool.connection_kwargs
```
tarekziade: let's return a copy with a dict() wrapping, so people can't change the settings

tarekziade:
r+

PR #569, opened 2015-11-11
Add pool with max_backlog option (fixes #509)

### LINKED ISSUE #509: SQLA Queue with max backlog
As stated #186, it could be nice to have a max_backlog option for the connection pool. Since it is not native in SQLAlchemy it was not part of #231 

This was implemented in Sync a while ago:
https://github.com/mozilla-services/server-syncstorage/commit/f93ca85a985364f123ad79cbdc2fcc29c5656586

We could: 
- discuss with SQLA to see it would be something relevant to contribute
- import and test the existing code otherwise 
- set it as default connection queue when creating the engine

leplatrem:
This code from imported from Mozilla Services Sync.
See mozilla-services/server-syncstorage@f93ca85

*Note: Cliquet is licensed under Apache and Sync is under MPL*

@Natim r?
@tarekziade, @rfk  comments?

### DIFF
--- a/cliquet/cache/postgresql/__init__.py
+++ b/cliquet/cache/postgresql/__init__.py
@@ -38,9 +38,15 @@ class Cache(CacheBase):
 
         cliquet.cache_pool_size = 10
         cliquet.cache_maxoverflow = 10
+        cliquet.cache_max_backlog = -1
         cliquet.cache_pool_recycle = -1
         cliquet.cache_pool_timeout = 30
-        cliquet.cache_poolclass = sqlalchemy.pool.QueuePool
+        cliquet.cache_poolclass = cliquet.storage.postgresql.pool.QueuePoolWithMaxBacklog
+
+    The ``max_backlog``  limits the number of threads that can be in the queue
+    waiting for a connection.  Once this limit has been reached, any further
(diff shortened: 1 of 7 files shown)


cliquet/cache/postgresql/__init__.py
```diff
@@ -53,7 +59,7 @@ class Cache(CacheBase):
         of connections used in a multi-process deployment.
 
     :noindex:
-    """
+    """  # NOQA
```
Natim: Why would you need ``# NOQA`` shouldn't we let an empty line before instead?

tarekziade:
Back to the original reason why this was done : https://bugzilla.mozilla.org/show_bug.cgi?id=750964

I am curious if we've reproduced this issue in our load tests ?

cliquet/cache/postgresql/__init__.py
leplatrem: Because of the 80 chars limit of the default ``cliquet.cache_poolclass`` line :]

leplatrem:
> I am curious if we've reproduced this issue in our load tests ?

Not that I am aware of :)

leplatrem:
@Natim bump r?

Natim:
r+

Issue #274, opened 2015-05-21
Add PATCH on collections


leplatrem:
```
PATCH /articles
{
    "items": [
        {"id": "56t7y8u9", "archived": true},
        {"id": "86r735ty", "title": "Yeah"}
    ]
}
```

Returns the list of modified records.

* ``400`` if ``id`` is missing in some provided record
* ``412`` if with preconditions headers etc.

Natim:
Isn't this already handled by the batch endpoint?

leplatrem:
> Isn't this already handled by the batch endpoint?

Yes, indeed, but this would be resource oriented, and more performant

PR #570, opened 2015-11-11
Add list of impacted records on ResourceChanged event (fixes #501)

### LINKED ISSUE #501: Payload v2 for notifications
Stuff to add later in the payload, as we need them

- [ ] Add impacted records

leplatrem:
This allowed me to implement a very nice demo of live sync on a map with cliquet-pusher :) 

This also allows listeners to react on particular field change, since old and new version of records is provided.

Note: the `permission` attribute is not provided, only the content of `data`.

Question: should the list of impacted records be in the event `payload` attribute instead of being a member?

Thoughts?

### DIFF
--- a/cliquet/resource/__init__.py
+++ b/cliquet/resource/__init__.py
@@ -579,30 +579,50 @@ class UserResource(object):
             'data': result
         }
 
-        # XXX this should be plugged with the transaction system
-        # when it's added
-        if action != ACTIONS.READ:
-            # the collection has been changed
-            if not isinstance(result, list):
-                result = [result]
-            impacted = {}
-            if action == ACTIONS.CREATE:
-                impacted = [{'new': r} for r in result]
-            elif action == ACTIONS.DELETE:
-                impacted = [{'old': r} for r in result]
-            elif action == ACTIONS.UPDATE:
-                impacted = [{'new': r, 'old': old} for r in result]
-            try:
-                event = ResourceChanged(action, self, impacted, self.request)
-                self.request.registry.notify(event)
-            except Exception:
-                logger.error("Unable to notify", exc_info=True)
+        self._notify_event(result, action, old)
 
         return body
 
     #
     # Internals
     #
+
+    def _notify_event(self, result, action, old):
+        """
+        Emit a :class:`cliquet.event.ResourceChanged` event with the
+        impacted records.
+
+        # XXX: this should be plugged with the transaction system.
+        When plugged with pyramid_tm, this method will have to be rewritten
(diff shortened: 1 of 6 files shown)


cliquet/resource/__init__.py
```diff
@@ -580,8 +580,17 @@ def postprocess(self, result, action=ACTIONS.READ):
...
             # the collection has been changed
+            if not isinstance(result, list):
+                result = [result]
+            impacted = {}
+            if action == ACTIONS.CREATE:
+                impacted = [{'new': r} for r in result]
+            elif action == ACTIONS.DELETE:
+                impacted = [{'old': r} for r in result]
+            elif action == ACTIONS.UPDATE:
+                impacted = [{'new': r, 'old': old} for r in result]
```
tarekziade: old is the same for every r here ?

cliquet/resource/__init__.py
```diff
@@ -580,8 +580,17 @@ def postprocess(self, result, action=ACTIONS.READ):
         # when it's added
         if action != ACTIONS.READ:
             # the collection has been changed
+            if not isinstance(result, list):
+                result = [result]
```
Natim: Why do we want it to be a list? It doesn't cope with @tarekziade comment (https://github.com/mozilla-services/cliquet/pull/570/files#r44639830)

cliquet/resource/__init__.py
Natim: I might be wrong but I think in that case result is a list with only one entry since we cannot update multiple records at once yet.

Natim: But it is something we eventually want with https://github.com/mozilla-services/cliquet/issues/274

tarekziade: maybe we should add a comment to explain here because it's a bit puzzling.

cliquet/resource/__init__.py
leplatrem: It should always be a list actually.

When we will plug the notifications with the transaction system, there will multiple creation/updates/deletes. So right now, when a single record is impacted, we put it in a list

cliquet/resource/__init__.py
leplatrem: Yes, I will document this. This will have to be rewritten when the notifications will be plugged with transactions. So far, Remy is right, only record can be modified at a time. But I wanted the event attribute to be unchanged when we go further with transactions

leplatrem:
@almet can you check that this would fit the needs of signing collections when a bulk of changes is sent?

PR #573, opened 2015-11-12
Fix principals of permission backend not being plugged by default

leplatrem:
@Natim r?


Natim:
Test aren't passing but I am happy with the implementation.

leplatrem:
Final r?

PR #577, opened 2015-11-16
Maintain pagination offset (fixes #366)

### LINKED ISSUE #366: Pagination loop when the number of records is a multiple of the limit
It is the case with the loadtest.

The before last page is:

    http://localhost:8888/v1/buckets/d31028b0-f297-125a-681a-91f99e3cb3b0/collections/articles/records?_limit=20&\ 
    _token=eyJsYXN0X21vZGlmaWVkIjoxNDM2MzYyNjI4ODQ5fQ%3D%3D

The last page is:

    http://localhost:8888/v1/buckets/d31028b0-f297-125a-681a-91f99e3cb3b0/collections/articles/records?_limit=20&\
    _token=eyJsYXN0X21vZGlmaWVkIjoxNDM2MzYyNjMwOTQxfQ%3D%3D

And the last page give the token of the before last one so we have a loop.

leplatrem:
@almet @Natim what do you think ?

### DIFF
--- a/cliquet/resource/__init__.py
+++ b/cliquet/resource/__init__.py
@@ -222,7 +222,7 @@ class UserResource(object):
         filter_fields = [f.field for f in filters]
         include_deleted = self.model.modified_field in filter_fields
 
-        pagination_rules = self._extract_pagination_rules_from_token(
+        pagination_rules, offset = self._extract_pagination_rules_from_token(
             limit, sorting)
 
         records, total_records = self.model.get_records(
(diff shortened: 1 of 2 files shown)


cliquet/resource/__init__.py
```diff
@@ -921,10 +924,13 @@ def _extract_pagination_rules_from_token(self, limit, sorting):
...
+        offset = 0
         if token:
             try:
-                last_record = json.loads(decode64(token))
-                assert isinstance(last_record, dict)
+                tokeninfo = json.loads(decode64(token))
+                assert isinstance(tokeninfo, dict)
+                last_record = tokeninfo['last_record']
+                offset = tokeninfo['offset']
             except (ValueError, TypeError, AssertionError):
```
leplatrem: KeyError missing

Natim:
I think I like it :+1:

almet:
Nice!

PR #578, opened 2015-11-17
Initalize Service class attributes before including plugins

leplatrem:
@Natim r?


leplatrem:
aaahah pypy now too slow with backlog tests `test_max_backlog_fails_when_reached()` `self.assertTrue(t2 - t1 < 0.9)`

Natim:
r+

PR #580, opened 2015-11-17
Fix Redis error traces not being logged (fixes #560)

### LINKED ISSUE #560: log every redis error
Let's log any error in wrap_redis_error, so we are sure to get a log of any problems

we should use logger.exception to log the TB

leplatrem:
@tarekziade r?


Natim:
:+1:

PR #584, opened 2015-11-18
Load plugins after setting up authn/authz

leplatrem:
While developing kinto-attachment, we realized that if plugins were included
before setting-up authn/authz, their views could not be protected by permissions.

For example, when a plugin defines a service like this in a view:

```python

    upload = Service(name="upload",
                     path="/upload",
                     description="",
                     factory=RouteFactory)
    @upload.get(permission="upload")
    def upload_get(request):
        return {"ok": True}

```

The factory is taken into account, but the authorization policy was never
called!

With this change, the plugins are included after ``cliquet.initialization.setup_authentication()``. Thus after ``config.include("pyramid_multiauth")`` thus after setting
up the authorization policy from conf ``multiauth.authorization_policy``.

The side-effet of this is that plugins cannot set default settings regarding auth.
It was the case of *cliquet-fxa* which defines an implicit setting ``multiauth.policy.fxa.use`` that allowed to just to specify ``fxa`` in the ``multiauth.policies`` setting.

The setting ``multiauth.policy.fxa.use`` must now be defined explicitly.

------------

There might be alternative within pyramid for an authorization policy to be applied on every view registered, but I could not find it. I thought it could also come from cornice, but could not find anything explicit in the code base.

@tarekziade @almet @Natim thoughts?

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -7,7 +7,16 @@ This document describes changes between each past release.
 2.12.0 (unreleased)
 -------------------
 
-- Nothing changed yet.
+**Breaking changes**
+
+- When using *cliquet-fxa*, the setting ``multiauth.policy.fxa.use`` must now
+  be explicitly set to ``cliquet_fxa.authentication.FxAOAuthAuthenticationPolicy``
+
+
+**Bug fixes**
+
+- Include plugins after setting up components (like authn/authz) so that plugins
+  can register views with permissions checking
 
 
 2.11.0 (2015-11-17)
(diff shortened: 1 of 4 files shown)


Natim:
We should probably update cliquet-fxa README to add the default settings that must now be explicit :)

Natim:
It looks good to me.

Also shall we add a test to make sure that plugins views can handle authentication?

almet:
I find it actually better to have to specify the authentication explicitly rather than with a `fxa` shortcut, so that's a good news.

After some thinking, it seems better to force the users of cliquet to load the plugins *after* the initialization (I was pondering with the idea of defining it as another init step but that sounds like a recipe for hard-to-debug setups).

r+ for me.

leplatrem:
Great to hear your positive feedback guyz! I will try to implement a test for this

leplatrem:
The above test reproduces the problem \o/ :)

PR #585, opened 2015-11-19
Improve configuration of listeners and  add read event (fixes #492, fixes #555, fixes #493)

### LINKED ISSUE #492: Deactivate notification by default and only activate selected one in the configuration

### LINKED ISSUE #493: Handle ACTION.READ notification activation.

### LINKED ISSUE #555: Improve configuration of listeners
When I implemented [cliquet-pusher](https://github.com/leplatrem/cliquet-pusher), I realized that currently we have only one set of configuration per type of listener.

My use-case would be to have several instantiations of the same type of listeners but with different settings (e.g. send `records` notifs to one channel and `bucket` to another)

So the only modification I would suggest is this (following pyramid-multiauth settings):

Instead of : 

```ini
cliquet.event_listeners = cliquet_pusher.listener
cliquet.event_listeners.pusher.resources = record
cliquet.event_listeners.pusher.channel = {bucket_id}-{collection_id}
```

Have this:

```ini
cliquet.event_listeners = pusher_records pusher_admin

cliquet.event_listeners.pusher_records.use = cliquet_pusher.listener
cliquet.event_listeners.pusher_records.resources = record
cliquet.event_listeners.pusher_records.channel = {bucket_id}-{collection_id}

cliquet.event_listeners.pusher_admin.use = cliquet_pusher.listener
cliquet.event_listeners.pusher_admin.resources = bucket group collection
cliquet.event_listeners.pusher_admin.channel = admin-notifs
```

We could even imagine to have general settings (like `actions` or `resources`), for example:

```ini
cliquet.event_listeners = redis_records redis_count

cliquet.event_listeners.redis_records.use = cliquet.listeners.redis
cliquet.event_listeners.redis_records.listname = asyncqueue
cliquet.event_listeners.redis_records.resources = record bucket collection
cliquet.event_listeners.redis_records.actions = create update delete

cliquet.event_listeners.redis_count.use = cliquet.listeners.redis
cliquet.event_listeners.redis_count.listname = hits
cliquet.event_listeners.redis_count.actions = read
```

Thoughts ?

leplatrem:
@tarekziade @Natim r?

### DIFF
--- a/cliquet/initialization.py
+++ b/cliquet/initialization.py
@@ -373,15 +373,42 @@ def setup_logging(config):
     config.add_subscriber(on_new_response, NewResponse)
 
 
-def setup_listeners(config):
-    listeners = aslist(config.get_settings()['event_listeners'])
-    from cliquet.events import ResourceChanged
+def _filter_events(callback, actions, resources):
+    def wrapped(event):
+        action = event.payload.get('action')
+        resource_name = event.payload.get('resource_name')
+        if action and action not in actions:
+            return
+        if resource_name and resources and resource_name not in resources:
+            return
+        return callback(event)
+    return wrapped
 
-    for listener in listeners:
+
+def setup_listeners(config):
+    from cliquet.events import ResourceChanged, ACTIONS
+
+    write_actions = (ACTIONS.CREATE, ACTIONS.UPDATE, ACTIONS.DELETE)
+    settings = config.get_settings()
+    listeners = aslist(settings['event_listeners'])
+
+    for name in listeners:
         logger.info('Setting up %r listener')
-        listener_mod = config.maybe_dotted(listener)
-        listener = listener_mod.load_from_config(config)
-        config.add_subscriber(listener, ResourceChanged)
+        prefix = 'event_listeners.%s.' % name
+
+        try:
+            listener_mod = config.maybe_dotted(name)
+            assert hasattr(listener_mod, 'load_from_config')
+            prefix = 'event_listeners.%s.' % name.split('.')[-1]
+        except (ImportError, AssertionError):
+            listener_mod = config.maybe_dotted(settings[prefix + 'use'])
+        listener = listener_mod.load_from_config(config, prefix)
+
+        actions = aslist(settings.get(prefix + 'actions', '')) or write_actions
+        resource_names = aslist(settings.get(prefix + 'resources', ''))
+        callback = _filter_events(listener, actions, resource_names)
+
+        config.add_subscriber(callback, ResourceChanged)
 
 
 def load_default_settings(config, default_settings):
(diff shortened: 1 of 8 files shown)


cliquet/initialization.py
```diff
@@ -373,15 +374,45 @@ def on_new_response(event):
...
 
+def _filter_events(callback, actions, resources):
+    def wrapped(event):
+        action = event.payload.get('action')
+        resource_name = event.payload.get('resource_name')
+        if action and action not in actions:
+            return
+        if resource_name and resources and resource_name not in resources:
+            return
+        return callback(event)
```
tarekziade: I am not sure to understand what _filter_events does. It seems to : 
- call the callback and return the result
- not call it and return implicitely None

But then in the code it;s used like this:

```python
callback = _filter_events(callback, actions, resources)
```
So I am confused about what is "callback" exactly. I think this can be clarified by:

- a better naming
- explicit returns everywhere in this filtering function

cliquet/initialization.py
```diff
@@ -373,15 +374,45 @@ def on_new_response(event):
     config.add_subscriber(on_new_response, NewResponse)
 
 
+def _filter_events(callback, actions, resources):
+    def wrapped(event):
```
tarekziade: use functools.wraps here to improve tracebacks

cliquet/initialization.py
```diff
@@ -373,15 +374,45 @@ def on_new_response(event):
...
+    for name in listeners:
         logger.info('Setting up %r listener')
-        listener_mod = config.maybe_dotted(listener)
-        listener = listener_mod.load_from_config(config)
-        config.add_subscriber(listener, ResourceChanged)
+        prefix = 'event_listeners.%s.' % name
+
+        try:
+            listener_mod = config.maybe_dotted(name)
+            assert hasattr(listener_mod, 'load_from_config')
```
tarekziade: why a plain assert here ? it means that if we run python using the -O option this line is stripped
I think you just want to catch AttributeError here

leplatrem: :+1:

cliquet/initialization.py
leplatrem: * `_filter_events` returns `wrapped`, the decorated function.
* the decorated function does execute the `callback` if the event is filtered
* `callback` could be renamed `listener` indeed
* the return value of `callback(event)` is useless, so we could just remove this `return`

tarekziade: > _filter_events returns wrapped, the decorated function.

Maybe the name of the function could make it clear that it's a decorator ?  (no strong opinion on this but since it's used inline )

> the return value of callback(event) is useless, so we could just remove this return 

Yeah, In that case I would be homogeneous and remove all the returns. It's a good practice to always return the same type from a function.

Natim:
Look good to me :+1: r+

tarekziade:
neat r+

PR #595, opened 2015-11-25
rename "hello" to "project_name"

Natim:
see https://github.com/mozilla-services/cliquet/blob/master/cliquet/views/hello.py#L14

it seems to be clearer to rename it to project_name

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -7,6 +7,10 @@ This document describes changes between each past release.
 2.12.0 (unreleased)
 -------------------
 
+**Protocol**
+
+- ``hello`` was renamed ``name`` on the hello page property (#595),
+
 **Breaking changes**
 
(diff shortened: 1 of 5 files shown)


tarekziade:
name vs project_name: what is the goal of having this value ?  I am not sure I understand its utility

Natim:
That's the hello page, there are no goal except being friendly with the people that end-up to that page.
That's the same utility as the documentation link. That's why we called it hello in the first place.

leplatrem:
project_name would match the setting name indeed.

Natim:
The rationale behind preferring ``name`` rather than ``project_name`` is that we are on the project resource, so we already know the context, we just want the project property name.

The same way that we would prefer ``project.name``over ``project.project_name``
The same way we declared ``project.documentation`` over ``project.project_documentation``
The same way that we would prefer ``article.title`` rather than ``article.article_title``

leplatrem:
Implemented in #600

PR #601, opened 2015-11-27
Implement partial responses. Fix #384.

### LINKED ISSUE #384: Partial responses
It would make sense to think about partial responses (some discussion [started here](https://github.com/mozilla-services/readinglist/issues/35) a while ago).

A simple approach would be to add something like ``?_fields=title,category.name``.

Source: http://blog.octo.com/en/design-a-rest-api/#partial

MrChoclate:


leplatrem:
Thanks!

As a general feedback, I would suggest to add some tests, especially to assert that the code behaves as expected by the specification of this new feature ([read more about tests as specs ](http://blog.mathieu-leplatre.info/your-tests-as-your-specs.html))

Please also don't forget to update the documentation and changelog to advertise this new cool feature!

cliquet/resource/__init__.py
```diff
@@ -240,6 +248,11 @@ def collection_get(self):
             next_page = self._next_page_url(sorting, limit, lastrecord, offset)
             headers['Next-Page'] = encode_header(next_page)
 
+        records = [
+            {field: val for field, val in record.items() if field in fields}
```
leplatrem: What if a *known field* is not present in the record ?

I would suggest to let extract_fields return `None` and don't perform the filtering if no `?_fields=` is specified (for performance reason also)

MrChoclate: > What if a known field is not present in the record ?

I don't undestand. `fields` is a subset of _get_known_fields. I assume that all fields of a record are in `known_fields` then I filter the fields by `?_filter=`. What could be wrong? What do you mean?

I agree that no filtering should be made if there is no `?_fields=`. I will also rename `fields` with `partial_fields`.

MrChoclate:
Todo:
- [ ] Partial response on get record.
- [ ] Add tests.
- [ ] Update documentation.
- [ ] Update changelog.

cliquet/resource/__init__.py
MrChoclate: Oh !
Do you prefer something like?
```python
records = [
    {field: record.get(field) for field in partial_fields}
    for record in records
]
```

I have no experience with optional fields so that why I hardly see the difference. Is the value of an optional field null or is the field missing? If a field is missing in the record and we ask for this field what should be the behavior?

cliquet/tests/resource/test_partial_response.py
```diff
@@ -0,0 +1,40 @@
+from pyramid import httpexceptions
+
+from cliquet.resource import ShareableResource
+from cliquet.tests.resource import BaseTest
+
+class PartialResponseBase(BaseTest):
+    def setUp(self):
+        super(PartialResponseBase, self).setUp()
+        self.resource._get_known_fields = lambda: ['field', 'other']
```
MrChoclate: I dislike this but I am not familiar with mocking.

cliquet/resource/__init__.py
```diff
@@ -1046,6 +1088,17 @@ def get_parent_id(self, request):
...
+    def _extract_fields(self):
+        """Override default fields extraction from QueryString to allow
+        the permission_field to exist.
+        """
+
+        fields = super(ShareableResource, self)._extract_fields()
+        if fields:
+            fields.append(self.model.permissions_field)
+
+        return fields
```
leplatrem: This reminds me that @Natim suggested a while ago to hide the permissions when the user has not the permission to write the record.

I suggest that we handle this part in another pull-request if you don't mind :)

cliquet/tests/resource/test_partial_response.py
leplatrem: This is acceptable IMO

cliquet/tests/resource/test_partial_response.py
```diff
@@ -0,0 +1,40 @@
...
+class BasicTest(PartialResponseBase):
+
+    def test_fields_parameter_do_projection_on_get(self):
+        self.resource.request.GET['_fields'] = 'field'
+        record = self.resource.get()
+        self.assertIn('field', record['data'])
+        self.assertNotIn('other', record['data'])
+
+    def test_fields_parameter_do_projection_on_get_all(self):
+        self.resource.request.GET['_fields'] = 'field'
```
leplatrem: You could probably also test with two field names (to test the split on `,`)

cliquet/resource/__init__.py
```diff
@@ -811,6 +829,30 @@ def _raise_400_if_id_mismatch(self, new_id, record_id):
...
+        """Extract the fields to do the projection from QueryString parameters.
+        """
+
+        fields = self.request.GET.get('_fields', None)
+        if fields:
+            fields = fields.split(',')
+            known_fields = self._get_known_fields()
+            if set(fields) - set(known_fields):
+                error_msg = "Fields {} do not exist".\
+                    format(set(fields) - set(known_fields))
```
leplatrem: nitpick: for better readabilty I suggest that you use an intermediary variable to store the intersection

leplatrem:
This is a very good work! It's a real pleasure to receive such contributions :+1:

cliquet/resource/__init__.py
MrChoclate: Without this there is no permission displayed when we get a record. There is no reason to hide it when we ask for partial_fields. It feels wrong when I test it and did not receive the permissions.

If you need write permission to display permissions, it should not be done here since it has nothing to do with partial response.

cliquet/tests/resource/test_partial_response.py
MrChoclate: Agreed

cliquet/resource/__init__.py
MrChoclate: Agreed

cliquet/resource/__init__.py
Natim: Well actually I hope we could ask the permission using partial responses on the collection_get endpoint.

MrChoclate: ```/v0/{{collection}}?_fields=permissions``` ?

What should it returns? permissions of the collection? permissions of each record? Wh?y would you use partial responses?

Natim: Permissions for each records, this would be useful to be able to sync permissions between two servers at once.
See https://github.com/Kinto/kinto.py/pull/26/files#diff-7d388f51c9ff65051463565ec3207ea8R32

MrChoclate: I think this is more than partial responses. I will open a new issue.

I will remove the code, then asking for partial response will hide permissions. I spotted a "bug" where asking for collection endpoint adds __permission__. This was not intended. Correct behavior should be clearly defined.

cliquet_docs/api/resource.rst
```diff
@@ -244,6 +244,51 @@ Pagination works with sorting, filtering and polling.
     ``If-None-Match``), in order to detect changes on collection while
     iterating through the pages.
 
+Partial response
+----------------
+
+If the ``_fields`` parameter is provided, only the fields specified are returned.
+Fields are separated with a comma. It is currently not possible to ask
+for nested field.
```
MrChoclate: fields*

MrChoclate:
Could someone review my documentation?
What's the next step? 
Is it possible that Travis CI build this?

cliquet_docs/api/resource.rst
```diff
@@ -244,6 +244,51 @@ Pagination works with sorting, filtering and polling.
...
+
+This is vital in mobile contexts (UMTS) where bandwidth usage must be optimized.
```
leplatrem: nitpick: I think UMTS is superflous here

leplatrem:
> Could someone review my documentation?

The documentation looks great!

> What's the next step?

I believe it is not fear from being merged! 

> Is it possible that Travis CI build this?

I don't know why, but it was unchecked in the settings!! It should be ok on the next push :)

MrChoclate:
Should I rebase it? Or you will do it on when merging?

Natim:
Yes please rebase it so that we can merge it

leplatrem:
@MrChoclate I can take care of that if you prefer :)

MrChoclate:
Please, do it

PR #604, opened 2015-11-30
Keep last modified

### LINKED ISSUE #587: Ability to specify last_modified value when updating records
One current limitation of the current protocol is that fact that the `last_modified` metadata (on buckets, collections and records) cannot be set from an external source.

For instance, if one wants replicate the data between two instances (origin and destination), when pushing the updates, the `last_modified` field from the origin data will be replaced by a new one defined by the destination.

As such, the data between the two instances cannot be considered equal.

One way to fix this would be to let the client define what the `last_modified` field is, checking for some heuristics: the new `last-modified` cannot be less than the `max(collection)` for instance.

almet:
Allow to keep the `last_modified` header when specified in the data.  Fixes #587.

### DIFF
--- a/cliquet/resource/__init__.py
+++ b/cliquet/resource/__init__.py
@@ -509,6 +509,7 @@ class UserResource(object):
         .. code-block:: python
 
             def process_record(self, new, old=None):
+                new = super(MyResource, self).process_record(new, old)
                 version = old['version'] if old else 0
                 new['version'] = version + 1
                 return new
@@ -520,6 +521,7 @@ class UserResource(object):
             from cliquet.errors import raise_invalid
 
             def process_record(self, new, old=None):
+                new = super(MyResource, self).process_record(new, old)
                 if new['browser'] not in request.headers['User-Agent']:
                     raise_invalid(self.request, name='browser', error='Wrong')
                 return new
--- a/cliquet/resource/__init__.py
+++ b/cliquet/resource/__init__.py
@@ -495,8 +495,7 @@ class UserResource(object):
         # Retreive the last_modified information from a querystring if present.
         last_modified = self.request.GET.get('last_modified')
 
-        deleted = self.model.delete_record(
-            record, last_modified=last_modified)
+        deleted = self.model.delete_record(record, last_modified=last_modified)
         return self.postprocess(deleted, action=ACTIONS.DELETE)
 
     #
@@ -537,8 +536,8 @@ class UserResource(object):
 
         # Drop the new last_modified if lesser or equal to the old one.
         new_last_modified = new.get(self.model.modified_field)
-        if (new_last_modified and
-                new_last_modified <= old[self.model.modified_field]):
+        is_greater = new_last_modified <= old[self.model.modified_field]
+        if (new_last_modified and is_greater):
             del new[self.model.modified_field]
 
         return new
--- a/cliquet/resource/__init__.py
+++ b/cliquet/resource/__init__.py
@@ -531,11 +531,13 @@ class UserResource(object):
         :returns: the processed record.
         :rtype: dict
         """
-        if old is None or self.model.modified_field not in old:
+        new_last_modified = new.get(self.model.modified_field)
+        not_specified = old is None or self.model.modified_field not in old
+
+        if new_last_modified is None or not_specified:
             return new
 
         # Drop the new last_modified if lesser or equal to the old one.
-        new_last_modified = new.get(self.model.modified_field)
         is_greater = new_last_modified <= old[self.model.modified_field]
         if (new_last_modified and is_greater):
             del new[self.model.modified_field]
--- a/cliquet/resource/__init__.py
+++ b/cliquet/resource/__init__.py
@@ -539,7 +539,7 @@ class UserResource(object):
 
         # Drop the new last_modified if lesser or equal to the old one.
         is_greater = new_last_modified <= old[self.model.modified_field]
-        if (new_last_modified and is_greater):
+        if new_last_modified and is_greater:
             del new[self.model.modified_field]
 
         return new
--- a/cliquet/resource/__init__.py
+++ b/cliquet/resource/__init__.py
@@ -492,7 +492,11 @@ class UserResource(object):
         record = self._get_record_or_404(self.record_id)
         self._raise_412_if_modified(record)
 
-        deleted = self.model.delete_record(record)
+        # Retreive the last_modified information from a querystring if present.
+        last_modified = self.request.querystring.get('last_modified')
+
+        deleted = self.model.delete_record(
+            record, last_modified=last_modified)
         return self.postprocess(deleted, action=ACTIONS.DELETE)
 
     #
--- a/cliquet/resource/__init__.py
+++ b/cliquet/resource/__init__.py
@@ -493,7 +493,7 @@ class UserResource(object):
         self._raise_412_if_modified(record)
 
         # Retreive the last_modified information from a querystring if present.
-        last_modified = self.request.querystring.get('last_modified')
+        last_modified = self.request.GET.get('last_modified')
 
         deleted = self.model.delete_record(
             record, last_modified=last_modified)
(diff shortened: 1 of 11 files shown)


Natim:
First of all I would like to say thank you for tackling this bug so fast :+1: I was kind of afraid it would block us on the long run.

almet:
Still misses (on the way) tests for creation and the migrations for the SQL schema.

cliquet/storage/postgresql/schema.sql
```diff
@@ -12,6 +12,13 @@ END;
 $$ LANGUAGE plpgsql
 IMMUTABLE;
 
+CREATE OR REPLACE FUNCTION from_epoch(epoch BIGINT) RETURNS TIMESTAMP AS $$
+BEGIN
+    RETURN TIMESTAMP WITH TIME ZONE 'epoch' + epoch * INTERVAL '1 millisecond';
```
Natim: Does this handle epoch being null?

Natim: Apparently it does :+1:

almet:
Just updated with the two missing bits.

cliquet/storage/redis.py
```diff
@@ -211,8 +234,13 @@ def delete(self, collection_id, parent_id, object_id,
 
         existing = self._decode(encoded_item)
 
+        # Need to delete the last_modified field of the record for it to not
+        # be kept (in no case we want to keep the old one).
+        del existing[modified_field]
+
```
Natim: nit: Need to remove whitespaces. for flake8

cliquet/resource/__init__.py
```diff
@@ -492,7 +492,11 @@ def delete(self):
         record = self._get_record_or_404(self.record_id)
         self._raise_412_if_modified(record)
 
-        deleted = self.model.delete_record(record)
+        # Retreive the last_modified information from a querystring if present.
+        last_modified = self.request.GET.get('last_modified')
+
+        deleted = self.model.delete_record(
+            record, last_modified=last_modified)
```
leplatrem: nit: I think elsewhere we do this:
```python
        deleted = self.model.delete_record(record,
                                           last_modified=last_modified)
```

cliquet/resource/__init__.py
```diff
@@ -492,7 +492,11 @@ def delete(self):
         record = self._get_record_or_404(self.record_id)
         self._raise_412_if_modified(record)
 
-        deleted = self.model.delete_record(record)
+        # Retreive the last_modified information from a querystring if present.
+        last_modified = self.request.GET.get('last_modified')
```
leplatrem: note: do not forget to update the API docs with a  `.. versionadded::` section

cliquet/resource/__init__.py
```diff
@@ -528,6 +532,15 @@ def process_record(self, new, old=None):
         :returns: the processed record.
         :rtype: dict
         """
+        if old is None or self.model.modified_field not in old:
+            return new
+
+        # Drop the new last_modified if lesser or equal to the old one.
+        new_last_modified = new.get(self.model.modified_field)
+        if (new_last_modified and
+                new_last_modified <= old[self.model.modified_field]):
```
leplatrem: nit: please use intermediary variable to avoid indenting in if conditions

cliquet/resource/__init__.py
```diff
@@ -528,6 +532,15 @@ def process_record(self, new, old=None):
         :returns: the processed record.
         :rtype: dict
         """
+        if old is None or self.model.modified_field not in old:
+            return new
```
leplatrem: This makes me realize that a `super()` from the examples of `process_record` in documention

cliquet/storage/memory.py
```diff
@@ -180,11 +184,28 @@ def _bump_timestamp(self, collection_id, parent_id):
...
-            current = previous + 1
-        self._timestamps[collection_id][parent_id] = current
+            collection_timestamp = previous + 1
+        else:
+            collection_timestamp = current
+
+        if not is_specified:
+            current = collection_timestamp
+
+        self._timestamps[collection_id][parent_id] = collection_timestamp
```
leplatrem: nit: Since the flow of this function is a bit complex, I would suggest putting a summary of the spec in the docstring (or a bit more comments).
It has been only a few days since we paired on this, and I already feel unsure about the intentions :D

cliquet/storage/memory.py
```diff
@@ -233,10 +254,15 @@ def delete(self, collection_id, parent_id, object_id,
                id_field=DEFAULT_ID_FIELD, with_deleted=True,
                modified_field=DEFAULT_MODIFIED_FIELD,
                deleted_field=DEFAULT_DELETED_FIELD,
-               auth=None):
+               auth=None, last_modified=None):
         existing = self.get(collection_id, parent_id, object_id)
+        # Need to delete the last_modified field of the record for it to not
+        # be kept (in no case we want to keep the old one).
```
leplatrem: nit: *for it to not be kept* is not really useful after *need to delete*

cliquet/storage/redis.py
```diff
@@ -96,11 +98,32 @@ def _bump_timestamp(self, collection_id, parent_id):
...
+                        collection_timestamp = int(previous) + 1
+                    else:
+                        collection_timestamp = current
+
+                    # Return the newly generated timestamp as the current one
+                    # only if nothing else was specified.
+                    if not is_specified:
+                        current = collection_timestamp
+
+                    print(is_specified, current, collection_timestamp, record)
```
leplatrem: print() left

cliquet/storage/redis.py
```diff
@@ -96,11 +98,32 @@ def _bump_timestamp(self, collection_id, parent_id):
...
-                        current = int(previous) + 1
-                    pipe.set(key, current)
+                        collection_timestamp = int(previous) + 1
+                    else:
+                        collection_timestamp = current
+
+                    # Return the newly generated timestamp as the current one
+                    # only if nothing else was specified.
+                    if not is_specified:
+                        current = collection_timestamp
```
leplatrem: Since `Redis` is a `MemoryBasedStorage` I believe this piece of code can be shared between the two

cliquet/tests/test_storage.py
```diff
@@ -253,6 +253,28 @@ def test_create_raise_unicity_error_if_provided_id_exists(self):
...
+        record[self.id_field] = RECORD_ID
+        record[self.modified_field] = last_modified
+        self.create_record(record=record)
+
+        retrieved = self.storage.get(object_id=RECORD_ID, **self.storage_kw)
+        self.assertIn(self.modified_field, retrieved)
+        self.assertEquals(retrieved[self.modified_field], last_modified)
+
+        # collection timestamp should not be modified.
+        collection_timestamp = self.storage.collection_timestamp(
```
leplatrem: nit: shortening the var name would be acceptable here to avoid wrapping args

cliquet/tests/test_storage.py
```diff
@@ -1083,3 +1170,27 @@ def test_warns_if_configured_pool_size_differs_for_same_backend_type(self):
...
+        self.assertIn(self.modified_field, retrieved)
+        self.assertEquals(retrieved[self.modified_field],
+                          record[self.modified_field])
+
+        # The collection timestamp should be bumped in this case.
+        collection_timestamp = self.storage.collection_timestamp(
+            **self.storage_kw)
+        self.assertEquals(
+            collection_timestamp,
+            first_record[self.modified_field])
```
leplatrem: :+1:

cliquet/tests/test_storage.py
```diff
@@ -284,6 +306,58 @@ def test_update_overwrites_record_id(self):
...
+        self.assertIn(self.modified_field, retrieved)
+        self.assertEquals(retrieved[self.modified_field],
+                          record[self.modified_field])
+
+        # The collection timestamp should be bumped in this case.
+        collection_timestamp = self.storage.collection_timestamp(
+            **self.storage_kw)
+        self.assertNotEquals(
+            collection_timestamp,
+            first_record[self.modified_field])
```
leplatrem: :+1:

Natim:
Youpi !!!!

PR #610, opened 2015-12-15
typo in RHEL package manager name.

FooBarQuaxx:


leplatrem:
Good catch :) Thanks a lot for taking the time to fix it!

PR #613, opened 2015-12-16
Refactor filtering of events to use Pyramid subscribers predicates (fixes #609)

### LINKED ISSUE #609: Use pyramid subscriber predicates rather than defining our own
Pyramid defines what they call [subscriber predicates](http://docs.pylonsproject.org/projects/pyramid/en/latest/narr/hooks.html#subscriber-predicates). We should use it rather than defining our own logic (also, it will be cleaner :-))

leplatrem:
@almet r?

### DIFF
--- a/cliquet/initialization.py
+++ b/cliquet/initialization.py
@@ -375,17 +375,35 @@ def setup_logging(config):
     config.add_subscriber(on_new_response, NewResponse)
 
 
-def _filter_events(listener_func, actions, resources):
-    def wrapped(event):
+class EventsFilterActions(object):
+    def __init__(self, actions, config):
+        self.actions = actions
+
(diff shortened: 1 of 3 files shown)


cliquet/initialization.py
```diff
@@ -375,17 +375,35 @@ def on_new_response(event):
     config.add_subscriber(on_new_response, NewResponse)
 
 
-def _filter_events(listener_func, actions, resources):
-    def wrapped(event):
+class EventsFilterActions(object):
```
almet: As it's a filter, what about naming them `ActionEventFilter`, and `ResourcesEventsFilter`?

almet:
Looks good to me, thanks!

PR #614, opened 2015-12-17
Support schemaless resources for _fields (ref #384)

leplatrem:
@MrChoclate @Natim review?

### DIFF
--- a/cliquet/tests/resource/test_views.py
+++ b/cliquet/tests/resource/test_views.py
@@ -706,13 +706,13 @@ class PaginationNextURLTest(BaseWebTest, unittest.TestCase):
         self.assertIn('https://server.name:443', resp.headers['Next-Page'])
 
 
-class PartialResponseTest(BaseWebTest, unittest.TestCase):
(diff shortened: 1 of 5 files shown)


cliquet/tests/resource/test_views.py
```diff
@@ -704,3 +704,33 @@ def test_next_page_url_relies_on_headers_information(self):
...
+        result = resp.json['data'][0]
+        self.assertNotIn('owner', result)
+
+    def test_specified_fields_are_included(self):
+        resp = self.app.get(self.collection_url + '?_fields=size,category')
+        result = resp.json['data'][0]
+        self.assertIn('size', result)
+        self.assertIn('category', result)
+
+    def test_unknown_fields_are_ignored(self):
```
MrChoclate: Is this for schemaless? I think so but cannot find where it is said.

MrChoclate:
LGTM

Natim:
I think we should always have ``id`` and ``last_modified`` even if they are not in ``_fields`` because they are part of the protocol.

leplatrem:
I merge like this and will take @Natim's into account in another PR

PR #615, opened 2015-12-17
Fix rst format in utilities API docs

leplatrem:
before

capture d ecran de 2015-12-17 22-01-46

after

capture d ecran de 2015-12-17 22-00-55

@n1k0 r?


n1k0:
Hmm, I preferred when the settings where listed as before the change. Reasoning?

leplatrem:
Good catch thanks! Fixed!

n1k0:
r+

PR #620, opened 2016-01-13
Refactor logger imports (fixes #102)

### LINKED ISSUE #102: Refactor structlog logger creation

leplatrem:


Natim:
GG r+ :)

PR #621, opened 2016-01-14
Do not log batch subrequests twice (fixes #450, Kinto/kinto#264)

### LINKED ISSUE #450: When answering to an OPTIONS request, the console shows the reponse twice.
I believe this happens because we print to the console as soon as a new request is created, but we would need to dig a but further to fix the problem.

leplatrem:
@almet r?


almet:
LGTM.

PR #628, opened 2016-01-26
API Capabilities

leplatrem:
I suggest we expose something like this on the Python side for plugins:

```python
config.add_api_capability("identifier", description="", documentation="")
```
> It would raise an error if already defined for example

And shown in the root URL as a mapping:

```js
"capabilities": {
  "auth-fxa": {
    "description": "Firefox Account authentication",
    "documentation": "http://github.com/mozilla-services/cliquet-fxa"
  }
}
```

@almet @n1k0 @Natim thoughts?

### DIFF
--- a/cliquet/__init__.py
+++ b/cliquet/__init__.py
@@ -119,9 +119,10 @@ def includeme(config):
 
     # Directive to declare arbitrary API capabilities.
     def add_api_capability(config, identifier, description="", url="", **kw):
-        if identifier in config.registry.api_capabilities:
-            error_msg = "API capability '%s' already exists." % identifier
-            raise ValueError(error_msg)
+        existing = config.registry.api_capabilities.get(identifier)
+        if existing:
+            error_msg = "The '%s' API capability was already registered (%s)."
+            raise ValueError(error_msg % (identifier, existing))
 
         capability = dict(description=description, url=url, **kw)
         config.registry.api_capabilities[identifier] = capability
--- a/cliquet/__init__.py
+++ b/cliquet/__init__.py
@@ -117,6 +117,18 @@ def includeme(config):
     # Public settings registry.
     config.registry.public_settings = {'batch_max_requests', 'readonly'}
 
+    # Directive to declare arbitrary API capabilities.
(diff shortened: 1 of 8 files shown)


Natim:
I totally like the idea :)

leplatrem:
This is now open for review/merge :)

cliquet/__init__.py
```diff
@@ -117,6 +117,18 @@ def includeme(config):
     # Public settings registry.
     config.registry.public_settings = {'batch_max_requests', 'readonly'}
 
+    # Directive to declare arbitrary API capabilities.
+    def add_api_capability(config, identifier, description="", url="", **kw):
+        if identifier in config.registry.api_capabilities:
+            error_msg = "API capability '%s' already exists." % identifier
```
almet: I think we can be more explicit here: `"The '%s' capability is already defined, with the values %s. Only one plugin can provide one capability" % (identifier, config.registry.api_capabilites[identifier])`.

cliquet_docs/api/utilities.rst
```diff
@@ -23,6 +23,10 @@ The returned value is a JSON mapping containing:
   - ``batch_max_requests``: Number of requests that can be made in a batch request.
   - ``readonly``: Only requests with read operations are allowed.
 
+- ``capabilities``: a mapping used by clients to detect optional features of the API.
+
+  - Example: ``{"auth-fxa": {"description": "Firefox Account authentication", "url": "http://github.com/mozilla-services/cliquet-fxa"}}``
```
almet: The example would be better if the json is expended.

cliquet_docs/quickstart.rst
```diff
@@ -200,6 +200,33 @@ Advanced initialization
 .. autofunction:: cliquet.initialize
 
 
+
+Declare API capabilities
```
almet: I would put this in the "plugins" section of the docs and not in the quickstart.

almet:
r+wc

michielbdejong:
Cool stuff!

Natim:
Yes great :+1:

leplatrem:
Thanks for your positive feedback!

Issue #629, opened 2016-01-27
Batch response status and transaction rollback


leplatrem:
Currently the `POST /batch` response status is:
* 200 if everything goes smooth
* 200 if a sub-response returns a 4XX status as expected by protocol (e.g. 400, 401, 403, 404, 412)
* 50X if a sub-response crashes (e.g 500, 503)

Current the transaction of the whole batch is:
* committed if everything goes smooth
* committed if a sub-response returns a 4XX status expected by protocol
* rolled back if a sub-response crashes

With the current implementation (based on `pyramid_tm` tween) the transaction final state depends of the final batch response status [(>=200 <400)](https://github.com/Pylons/pyramid_tm/blob/master/pyramid_tm/__init__.py#L28).
It is very easy to change.

But, we have to decide whether we want to:

1. return 200 and commit transaction when a sub-response is 4XX (*ie. keep current behaviour*)
1. return 4XX and rollback transaction when a sub-response is 4XX
1. return 200 and rollback transaction when a sub-response is 4XX

Thoughts ?

> Notes:
> * delete can generate 404 response if record was already deleted
> * https://github.com/mozilla-services/syncto/issues/78

leplatrem:
Note: @n1k0 would vote for `2.` for kinto.js needs.

Natim:
The second solution seems to make sense to me too.

PR #630, opened 2016-01-27
Allow custom views to handle errors in batch (fixes mozilla-services/syncto#78, rel #629)

leplatrem:
This is my proposition for the syncto fix. 

It does not change current protocol or transaction behaviour (decision for #629 can be made later)

@Natim r?


Natim:
r+ :+1:

PR #633, opened 2016-01-27
Use authn policy name in user prefix

leplatrem:
@Natim r?

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -27,6 +27,12 @@ This document describes changes between each past release.
 - Fix listener name logging during startup (#626)
 
 
+**New features**
+
+- The policy name used to configure authentication in settings is now used for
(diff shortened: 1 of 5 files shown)


Natim:
If tests are green it works for me :+1: (I am wondering if cliquet-fxa name will change in the process, same for basicauth)

leplatrem:
basicauth does not change.

cliquet-fxa will change indeed, unless we name the policy ``fxaoauth`` in the config file (instead of ``fxa``).
We can add a warning here and in Kinto (but let's be honest, this plugin has no critical use)

Natim:
Ok it works for me.

Natim:
Some test case have to be fixed: https://travis-ci.org/mozilla-services/cliquet/jobs/105195031#L193

PR #636, opened 2016-02-02
Output the cliquet version with --version.

almet:
r? @leplatrem 

Not sure how to test this. Insights?


leplatrem:
r+

If you want to add some tests you can have a look at the run command tests here : https://github.com/mozilla-services/cliquet/blob/2.15.0/cliquet/tests/test_scripts.py#L12-L17

and see how you could substitute `sys.stdout` by some `StringIO` or similar

PR #640, opened 2016-02-03
Comparison with pool timeout is too precise

leplatrem:
This test is usually unstable. We have a timeout of 1 and compare with 0.9. I give it a bit of margin here.


almet:
Oh, I hate these since they usually rely on the machine that runs the tests. I'll r+ though as I don't see any better way to do it.

PR #634, opened 2016-01-27
One event per resource per transaction

leplatrem:

PR #644, opened 2016-02-09
Listener must commit explicitly

leplatrem:
#634 introduced a breaking change. Since events are sent once the transaction is committed, if a listener wants to store/update/delete records from storage, an explicit *commit* must be performed.

This PR just adds a test to illustrate the situation in current master.

@tarekziade suggests that we add another Pyramid event, to distinguish the post-commit event from the one within the transaction.

If we want to avoid a breaking change, we could have:
- `ResourceChanged`: within transaction (like cliquet 2.15)
- `ResourcePostCommit` : once transaction is committed (would be used for listeners for example)

@Natim thouhts?


Natim:
Yes I would rather have to distinct events for that.

CHANGELOG.rst
```diff
@@ -7,6 +7,18 @@ This document describes changes between each past release.
 2.16.0 (unreleased)
 -------------------
 
+**Breaking changes**
+
+- When storage records are manipulated in events listeners, an explicit
+  transaction commit must be performed explicitly::
```
almet: that's two times explicit.

almet:
I'm not sure we want to have two kind of events here. We probably should issue a warning if the "commit()" method is not called.

leplatrem:
I close this PR and will open another one with a different strategy.

PR #645, opened 2016-02-09
Add a method to remove every principals

leplatrem:
There was no way to delete a principal from every users.

Here it is.

It will be useful in Kinto when a group is deleted.

@Natim r?

### DIFF
--- a/cliquet/permission/redis.py
+++ b/cliquet/permission/redis.py
@@ -55,6 +55,13 @@ class Permission(PermissionBase):
         if self._client.scard(user_key) == 0:
             self._client.delete(user_key)
 
+    def remove_principal(self, principal):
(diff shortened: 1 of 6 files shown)


cliquet/permission/redis.py
```diff
@@ -55,6 +55,13 @@ def remove_user_principal(self, user_id, principal):
         if self._client.scard(user_key) == 0:
             self._client.delete(user_key)
 
+    def remove_principal(self, principal):
+        with self._client.pipeline() as pipe:
+            user_keys = self._client.scan_iter(match='user:*')
+            for user_key in user_keys:
+                self._client.srem(user_key, principal)
```
Natim: You are not using the pipe here.

Natim:
r+wc

PR #652, opened 2016-02-15
Get rid of custom Enum (fixes #645)

leplatrem:
Isolated commit about Enum from #647 

@tarekziade r?

### DIFF
--- a/cliquet/errors.py
+++ b/cliquet/errors.py
@@ -1,31 +1,33 @@
 import six
 from pyramid import httpexceptions
+from enum import Enum
 
 from cliquet.logs import logger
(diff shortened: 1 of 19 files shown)


leplatrem:
The `.value` everywhere bother me a bit, but I think it's sane to use a standard `enum`.

@almet @Natim any strong thought against it before we merge?

Natim:
> The ``.value`` everywhere bother me a bit, but I think it's sane to use a standard ``enum``.

Same feeling, same conclusions.

almet:
yeah, it seems we're all there :-) Merging.

PR #654, opened 2016-02-16
pyramind config link updated

ayusharma:
Pyramid configuration is updated here : https://github.com/mozilla-services/cliquet/blob/master/cliquet_docs/reference/configuration.rst


leplatrem:
Good catch! Thanks for taking the time to fix it!

PR #660, opened 2016-02-18
Filter nested JSON objects fields.

Natim:
- [x] Add broken tests for Kinto/kinto#445

### DIFF
--- a/cliquet/resource/__init__.py
+++ b/cliquet/resource/__init__.py
@@ -224,7 +224,7 @@ class UserResource(object):
         filters = self._extract_filters()
         limit = self._extract_limit()
         sorting = self._extract_sorting(limit)
-        partial_fields = self._extract_fields()
+        partial_fields = self._extract_partial_fields()
 
         filter_fields = [f.field for f in filters]
         include_deleted = self.model.modified_field in filter_fields
@@ -249,7 +249,7 @@ class UserResource(object):
 
         if partial_fields:
             records = [
-                self._filter_fields(record, partial_fields)
+                dict_subset(record, partial_fields)
                 for record in records
             ]
 
@@ -343,9 +343,9 @@ class UserResource(object):
         self._raise_304_if_not_modified(record)
         self._raise_412_if_modified(record)
 
-        partial_fields = self._extract_fields()
+        partial_fields = self._extract_partial_fields()
         if partial_fields:
-            record = self._filter_fields(record, partial_fields)
+            record = dict_subset(record, partial_fields)
 
         return self.postprocess(record)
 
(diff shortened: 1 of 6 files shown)


cliquet/tests/resource/test_partial_response.py
```diff
@@ -46,6 +57,22 @@ def test_id_and_last_modified_are_not_filtered(self):
...
+        self.assertNotIn('bar', record['data']['original'])
+        self.assertNotIn('nested', record['data']['original'])
+
+    def test_nested_parameter_can_be_filtered_on_multiple_levels(self):
+        self.resource.request.GET['_fields'] = 'original.nested.size'
+        record = self.resource.get()
+        self.assertIn('nested', record['data']['original'])
+        self.assertIn('size', record['data']['original']['nested'])
+        self.assertNotIn('hash', record['data']['original']['nested'])
+
```
leplatrem: add test to filter multiple nested parameters: `original.foo, original.bar`

cliquet/resource/__init__.py
```diff
@@ -826,13 +827,9 @@ def _extract_fields(self):
                 }
                 raise_invalid(self.request, **error_details)
 
-        return fields
+            fields = fields + [self.model.id_field, self.model.modified_field]
```
almet: I don't understand why we're doing this?

almet: Could you add a little comment explaining the rationale behind this code?

cliquet/utils.py
```diff
@@ -147,7 +147,21 @@ def hmac_digest(secret, message, encoding='utf-8'):
 
 def dict_subset(d, keys):
     """Return a dict with the specified keys"""
-    return {k: d[k] for k in keys if k in d}
+    result = {}
+
+    for key in keys:
+        if '.' in key:
```
almet: Since python 2.6, the stdlib [`operator.attrgetter`](https://docs.python.org/2/library/operator.html#operator.attrgetter) has been added the capability of parsing dotted attributes, we might want to use this here.

cliquet_docs/api/resource.rst
```diff
@@ -260,8 +260,9 @@ Partial response
 ----------------
 
 If the ``_fields`` parameter is provided, only the fields specified are returned.
-Fields are separated with a comma. It is currently not possible to ask
-for nested fields.
+Fields are separated with a comma.
+
+Nested fields are specified using ``.`` (dot).
```
almet: Adding an example should be able to help users understand better.

cliquet_docs/api/resource.rst
```diff
@@ -322,7 +323,7 @@ List of available URL parameters
 - ``_fields``: filter the fields of the records
 
 
-Filtering, sorting, partial responsses and paginating can all be combined together.
+Filtering, sorting, partial responses and paginating can all be combined together.
```
almet: good catch!

cliquet/resource/__init__.py
Natim: This is because the id and last_modified field should always be return as part as the synchronisation protocol.
If the user forgot to add them in the _fields list we still want them in the record object.

almet: This comment would be a worthy addition to the current code :)

cliquet/utils.py
leplatrem: Since we handle dicts and not attributes, I'm not sure

almet: Too bad `operator.itemgetter` doesn't handle the `.` then.

PR #661, opened 2016-02-18
python installation command updated

ayusharma:
`brew install python3.4` gives the following error : 

```
brew install python3.4
Error: No available formula with the name "python3.4"
==> Searching for similarly named formulae...
Error: No similarly named formulae found.
==> Searching taps...
Error: No formulae found in taps.
```


Natim:
Thank you for this.

PR #663, opened 2016-02-22
Remove backward incompatibilities regarding Enums

leplatrem:
BC had been introduced in #652

* [x] Use COMPARISON Enum as constants instead of values
* [x] Use Enum in event subscriber predicate


Natim:
r+

almet:
Looks good to me!

PR #666, opened 2016-02-24
Troubleshooting section in docs

ayusharma:
Addition of troubleshooting section in cliquet docs.

### DIFF
--- a/cliquet_docs/troubleshooting.rst
+++ b/cliquet_docs/troubleshooting.rst
@@ -1,0 +1,40 @@
+Troubleshooting
+###############
+
+.. _troubleshooting:
+
+We are doing the best we can so you do not have to read this section.
+
+That said, we have included solutions (or at least explanations) for
+some common problems below.
+
+If you do not find a solution to your problem here, please
+:ref:`ask for help <communication_channels>`!
+
+
+ConnectionError: localhost:6379. nodename nor servname provided, or not known
+=============================================================================
+
+or
+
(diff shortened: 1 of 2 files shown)


cliquet_docs/troubleshooting.rst
```diff
@@ -0,0 +1,40 @@
...
+If you do not find a solution to your problem here, please
+:ref:`ask for help <communication_channels>`!
+
+
+ConnectionError: localhost:6379. nodename nor servname provided, or not known
+=============================================================================
+
+or
+
+IOError: [Errno 24] Too many open files
```
Natim: I would put both titles in the same paragraph title.

Natim:
Thank you for that. I am wondering if we should add a way to display these errors in Kinto troubleshooting since they can appear to our Kinto users too.

Maybe the Kinto troubleshooting part can include this troubleshooting file and add its own trouble after the inclusion?

@leplatrem @almet thoughts?

cliquet_docs/troubleshooting.rst
ayusharma: Okay, I'll make title in same paragraph.

leplatrem:
The duplication between Cliquet docs and the Kinto product docs is a major concern indeed.

I started something using templates, we can probably think of something for glossary, trouble docs and other things.

Anyway, thanks a lot @ayusharma for this contribution!!

PR #667, opened 2016-02-25
Service should only accept application/json content-types.

Natim:
Refs: Kinto/kinto#461

- [x] The check should only happens with POST, PUT and PATCH verbs.

### DIFF
--- a/cliquet/resource/viewset.py
+++ b/cliquet/resource/viewset.py
@@ -28,12 +28,26 @@ class ViewSet(object):
 
     readonly_methods = ('GET', 'OPTIONS', 'HEAD')
 
+    content_types = ["application/json"]
+
     service_arguments = {
         'description': 'Collection of {resource_name}',
     }
 
     default_arguments = {
         'permission': authorization.PRIVATE
+    }
+
+    default_post_arguments = {
+        "content_type": content_types,
+    }
+
+    default_put_arguments = {
+        "content_type": content_types,
+    }
+
+    default_patch_arguments = {
+        "content_type": content_types,
     }
 
     default_collection_arguments = {}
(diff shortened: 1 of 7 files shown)


Natim:
@ayusharma Can you validate that this branch fixes your issue?

There are tests [here](Kinto/kinto#461) to validate the integration with Kinto.

almet:
C'est bien.

Natim:
Few questions:

 - Do you think I should add a test for every verb on records and collections endpoints (get/option/head/put/patch/post/delete)?
 - Is it ok to have the errno INVALID_PARAMETER for the 406 and 415 http codes [See](https://github.com/mozilla-services/cliquet/pull/667/files#diff-40c8d42423b94fb5ae777dd830044a7cR152)

ayusharma:
Functionalities:  Woking.

Test Case: 1 Fail for python2.7 and Errors for Python 3

Take a look at error trace : http://pastebin.com/f4YfNeZP

cliquet/resource/viewset.py
```diff
@@ -28,12 +28,27 @@ class ViewSet(object):
 
     readonly_methods = ('GET', 'OPTIONS', 'HEAD')
 
+    content_types = ["application/json"]
```
leplatrem: Since this is not used at the class level, I would suggest not to define it here but in a constant (repeating is even acceptable IMO). 

What do you think?

leplatrem:
It looks good to me!

Do not hesitate to define new errors if no one matches (invalid parameter seems ok to me but I understand it's not very precise)

Theorically we should test very http method on both collection/record enpoints (motto *no code should be here if no test fails when it's removed*). But it's acceptable to leave it like you did. 

If you could go a bit further with:

- [ ] Update changelog (easier release)
- [ ] Update resources docs to mention the possible status for invalid accept/content-type on put/post/patch

would be great :)

cliquet/resource/viewset.py
Natim: One could want to accept xml for his service, that's the reason why I added this configuration here.

Natim:
> Do not hesitate to define new errors if no one matches (invalid parameter seems ok to me but I understand it's not very precise)

It seems to me that the http code is sufficient in that case. Do you think?

cliquet/resource/viewset.py
Natim: Oh you mean that it won't actually work if we update the content_type property.

leplatrem:
r+

amotl:
Didn't read the context here yet, but you referenced mozilla-services/cornice#343 here. Will have a look into that.

Natim:
At first I though that I could use function based content-type to return application/json in case it was post/put/patch and None in case it was the other methods.

Finally I did it differently in Cliquet.

@amotl Do you think you could add a failing test in cornice?

PR #674, opened 2016-03-04
Remove some imports at initialization

leplatrem:


leplatrem:
@magopian r? (moving the global import of `requests` to where they were needed could speed up CLI commands :))

magopian:
all good!

r+

PR #675, opened 2016-03-04
Do not always return 412 errors when request header `If-None-Match: *`` is sent on ``POST /collection`` (fixes #673)

### LINKED ISSUE #673: POST with If-None-Match: * and provided id in body always return 412
```
{
        "method": "POST",
        "path": "/buckets/default/collections/tasks/records",
        "headers": {
          "Authorization": "Basic dXNlcjpwYXNz",
          "If-None-Match": "*"
        },
        "body": {
          "data": {
            "id": "660d9409-241e-45cc-8b04-ba1ef72167bb",
            "title": "task3",
            "done": true
          },
          "permissions": {}
        }
      }
```
always return 412, instead of only when id already exists.

leplatrem:
@Natim r?

### DIFF
--- a/cliquet/resource/__init__.py
+++ b/cliquet/resource/__init__.py
@@ -276,16 +276,18 @@ class UserResource(object):
             Add custom behaviour by overriding
             :meth:`cliquet.resource.UserResource.process_record`
         """
-        self._raise_412_if_modified()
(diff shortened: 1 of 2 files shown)


Natim:
Good catch :+1:

PR #680, opened 2016-03-09
Add cache_prefix setting (fixes #227)

### LINKED ISSUE #227: Add key prefix option for cache
* Add a setting ``cliquet.cache_prefix`` (default: ``''``)
* Every cache entry key is prefixed by ``self.prefix`` in cache backend classes

lavish205:
issue #227

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -10,7 +10,7 @@ This document describes changes between each past release.
 **New features**
 
 - Default console log renderer now has colours (#671)
-- ``cache_prefix`` key option added for cache.
+- A ``cache_prefix`` setting was added for cache backends. (#680)
 
 
 3.0.0 (2016-02-26)
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -10,6 +10,7 @@ This document describes changes between each past release.
 **New features**
 
 - Default console log renderer now has colours (#671)
+- ``Prefix`` key option added for cache.
 
 
 3.0.0 (2016-02-26)
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -10,7 +10,7 @@ This document describes changes between each past release.
 **New features**
 
 - Default console log renderer now has colours (#671)
-- ``Prefix`` key option added for cache.
+- ``cache_prefix`` key option added for cache.
 
 
 3.0.0 (2016-02-26)
--- a/cliquet/cache/memory.py
+++ b/cliquet/cache/memory.py
@@ -41,8 +41,8 @@ class Cache(CacheBase):
     def get(self, key):
         current = utils.msec_time()
         expired = [k for k, v in self._ttl.items() if current > v]
-        for k in expired:
-            self.delete(self.prefix + k)
+        for expired_item_key in expired:
+            self.delete(expired_item_key.replace(self.prefix, ''))
         return self._store.get(self.prefix + key)
 
     def delete(self, key):
--- a/cliquet/cache/memory.py
+++ b/cliquet/cache/memory.py
@@ -51,4 +51,7 @@ class Cache(CacheBase):
 
 
 def load_from_config(config):
-    return Cache()
+    settings = config.get_settings()
+    prefix = settings['cache_prefix']
+    backend = Cache(prefix=prefix)
+    return backend
--- a/cliquet/cache/memory.py
+++ b/cliquet/cache/memory.py
@@ -52,6 +52,6 @@ class Cache(CacheBase):
 
 def load_from_config(config):
     settings = config.get_settings()
-    prefix = settings['cache_prefix']
+    prefix = settings.get('cache_prefix')
     backend = Cache(prefix=prefix)
     return backend
--- a/cliquet/cache/memory.py
+++ b/cliquet/cache/memory.py
@@ -13,6 +13,7 @@ class Cache(CacheBase):
     """
 
     def __init__(self, *args, **kwargs):
+        self.prefix = kwargs['prefix']
         super(Cache, self).__init__(*args, **kwargs)
         self.flush()
 
@@ -25,20 +26,24 @@ class Cache(CacheBase):
         self._store = {}
 
     def ttl(self, key):
+        key = self.prefix + key
         ttl = self._ttl.get(key)
         if ttl is not None:
             return (ttl - utils.msec_time()) / 1000.0
         return -1
 
     def expire(self, key, ttl):
+        key = self.prefix + key
         self._ttl[key] = utils.msec_time() + int(ttl * 1000.0)
 
     def set(self, key, value, ttl=None):
-        self._store[key] = value
         if ttl is not None:
             self.expire(key, ttl)
+        key = self.prefix + key
+        self._store[key] = value
 
     def get(self, key):
+        key = self.prefix + key
         current = utils.msec_time()
         expired = [k for k, v in self._ttl.items() if current > v]
         for key in expired:
@@ -46,12 +51,11 @@ class Cache(CacheBase):
         return self._store.get(key)
 
     def delete(self, key):
+        key = self.prefix + key
         self._ttl.pop(key, None)
         self._store.pop(key, None)
 
 
 def load_from_config(config):
     settings = config.get_settings()
-    prefix = settings.get('cache_prefix')
-    backend = Cache(prefix=prefix)
-    return backend
+    return Cache(prefix=settings['cache_prefix'])
--- a/cliquet/cache/memory.py
+++ b/cliquet/cache/memory.py
@@ -40,9 +40,9 @@ class Cache(CacheBase):
 
     def get(self, key):
         current = utils.msec_time()
-        expired = [k for k, v in self._ttl.items() if current > v]
+        expired = [k for k, v in self._ttl.items() if current >= v]
         for expired_item_key in expired:
-            self.delete(expired_item_key.replace(self.prefix, ''))
+            self.delete(expired_item_key[len(self.prefix):])
         return self._store.get(self.prefix + key)
 
     def delete(self, key):
--- a/cliquet/cache/memory.py
+++ b/cliquet/cache/memory.py
@@ -13,7 +13,6 @@ class Cache(CacheBase):
     """
 
     def __init__(self, *args, **kwargs):
-        self.prefix = kwargs['prefix']
         super(Cache, self).__init__(*args, **kwargs)
         self.flush()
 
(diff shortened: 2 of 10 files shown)


lavish205:
@leplatrem I'm getting `cache_prefix` KeyError as `config.get_settings()` is not returning settings defined in `DEFAULT_SETTINGS`

leplatrem:
It looks perfectly correct.

But when test are ran:
```
$ nosetests -s cliquet.tests.test_cache
```

I also get this error: 
```
Traceback (most recent call last):
  File "/home/mathieu/Code/Mozilla/cliquet/cliquet/tests/test_cache.py", line 40, in setUp
    self.cache = self.backend.load_from_config(self._get_config())
  File "/home/mathieu/Code/Mozilla/cliquet/cliquet/cache/memory.py", line 55, in load_from_config
    prefix = settings['cache_prefix']
KeyError: 'cache_prefix'
```

That's because the tests setup do not use DEFAULT_SETTINGS.
See, you should add something like this on `MemoryCacheTest` https://github.com/mozilla-services/cliquet/blob/master/cliquet/tests/test_cache.py#L162-L165

lavish205:
@leplatrem I have fixed this KeyError by `.get()` method, hope this works

lavish205:
#227

leplatrem:
> hope this works.

Nope actually. You should leave ``['cache_prefix']`` and set its value in the tests like I pointed out.

Here is a small list of things to do before this could be merged:

* [ ] Generalize the change to other backends (redis and postgresql)
* [ ] Add some tests in `test_cache.py` to ensure that the prefix value is used to fetch/delete the data associated to a key
* [ ] Add a test to ensure that the default prefix value is empty
* [ ] Update configuration documentation
* [ ] Update changelog
* [ ] Add yourself to contributors

leplatrem:
@lavish205 you could do something like that (I didn't test it):
```python
def test_get_uses_prefix(self):
    settings_prefix = self.settings.copy()
    settings_prefix['cache_prefix'] = 'prefix_'
    config_prefix = self._get_config(settings=settings_prefix)
    
    # Instantiate a cache backend with a prefix:
    backend_prefix = self.backend.load_from_config(config_prefix)

    # Set a value with a cache that has no prefix:
    self.cache.set('prefix_key', 'foo')

    # Obtain the value with the cache that has a prefix:
    obtained = backend_prefix.get('key')
    self.assertEqual(obtained, 'foo')

```

lavish205:
@leplatrem I think something is wrong with the code as test cases are not working as expected.

leplatrem:
@lavish205 yes, first you shouldn't set the prefix to `123`, so both `self.cache` and `backend_prefix` have a prefix.

Then, in order to make tests pass, you should use the `self.prefix` in `def get()` in the `Cache` backend class.

lavish205:
@leplatrem value of `backend_prefix` is `{'prefix': 'prefix_', '_store': {'prefix_key': 'bar'}, '_ttl': {}`, so can't access `prefix_key` on root of dict.

cliquet/tests/test_cache.py
```diff
@@ -156,12 +159,28 @@ def test_ping_returns_false_if_unavailable(self):
...
+        config_prefix = self._get_config(settings=settings_prefix)
+
+        # initiating cache backend with prefix:
+        backend_prefix = self.backend.load_from_config(config_prefix)
+
+        # set a value with a cache that has no prefix
+        backend_prefix.set('prefix_key', 'foo')
+
+        # obtain the value of cache that has prefix
+        obtained = backend_prefix.get('prefix_key')
```
leplatrem: This should be `get('key')` since this backend has prefix `prefix_` !

lavish205:
@leplatrem can you please check, if there is need to add any test case?

cliquet/cache/memory.py
```diff
@@ -25,20 +26,24 @@ def flush(self):
         self._store = {}
 
     def ttl(self, key):
+        key = self.prefix + key
```
Natim: I would rather use format here: ``key = "{}{}".format(self.prefix, key)`` or ``key = "%s%s" % (self.prefix, key)``

cliquet/cache/memory.py
```diff
@@ -25,20 +26,24 @@ def flush(self):
...
 
     def expire(self, key, ttl):
+        key = self.prefix + key
         self._ttl[key] = utils.msec_time() + int(ttl * 1000.0)
 
     def set(self, key, value, ttl=None):
-        self._store[key] = value
         if ttl is not None:
             self.expire(key, ttl)
+        key = self.prefix + key
```
Natim: You need to do that before calling expire because you should expire the key with the right prefix too,

cliquet/cache/memory.py
```diff
@@ -46,9 +51,13 @@ def get(self, key):
...
         self._ttl.pop(key, None)
         self._store.pop(key, None)
 
 
 def load_from_config(config):
-    return Cache()
+    settings = config.get_settings()
+    prefix = settings['cache_prefix']
+    backend = Cache(prefix=prefix)
+    return backend
```
Natim: You can probably do ``return Cache(prefix=settings['cache_prefix'])`` and save some intermediary variables.

cliquet/tests/test_cache.py
```diff
@@ -156,12 +159,86 @@ def test_ping_returns_false_if_unavailable(self):
     def test_ping_logs_error_if_unavailable(self):
         pass
 
+    def test_cache_prefix_is_set(self):
+        settings_prefix = self.settings.copy()
+        settings_prefix['cache_prefix'] = 'prefix_'
+        config_prefix = self._get_config(settings=settings_prefix)
+
+        # initiating cache backend with prefix:
+        backend_prefix = self.backend.load_from_config(config_prefix)
```
Natim: Why cannot you just do ``backend_prefix = self.backend.load_from_config(prefix="prefix_")`` here?

cliquet/tests/test_cache.py
```diff
@@ -156,12 +159,86 @@ def test_ping_returns_false_if_unavailable(self):
...
+
+        # set a value with a cache that has no prefix
+        backend_prefix.set('key', 'foo')
+
+        # obtain the value of cache that has prefix
+        obtained = backend_prefix.get('key')
+
+        self.assertEqual(obtained, 'foo')
```
Natim: Where do you check that redis was called with ``prefix_key``?

cliquet/tests/test_cache.py
```diff
@@ -156,12 +159,86 @@ def test_ping_returns_false_if_unavailable(self):
...
+        config_prefix = self._get_config(settings=settings_prefix)
+
+        # initiating cache backend with prefix:
+        backend_prefix = self.backend.load_from_config(config_prefix)
+
+        # set a value with a cache that has no prefix
+        backend_prefix.set('key', 'foo')
+
+        # deleting data
+        backend_prefix.delete('key')
```
Natim: You should try to get the key and make sure it has been deleted.

cliquet/tests/test_cache.py
```diff
@@ -156,12 +159,86 @@ def test_ping_returns_false_if_unavailable(self):
...
+
+    def test_prefix_value_used_with_ttl(self):
+        settings_prefix = self.settings.copy()
```
Natim: Since you are doing this code in all the test cases it probably worth to put in a ``def setUp(self)`` method.

cliquet/tests/test_cache.py
```diff
@@ -156,12 +159,86 @@ def test_ping_returns_false_if_unavailable(self):
...
+        settings_prefix['cache_prefix'] = 'prefix_'
+        config_prefix = self._get_config(settings=settings_prefix)
+
+        # initiating cache backend with prefix:
+        backend_prefix = self.backend.load_from_config(config_prefix)
+
+        # set a value with a cache that has no prefix
+        backend_prefix.set('key', 'foo', 10.0)
+        obtained = backend_prefix.ttl('key')
```
Natim: Same here, we should validate that redis is always called with the right key.

Natim:
That's a great job @lavish205 :+1: Thank you for working on it :)

cliquet/tests/test_cache.py
leplatrem: Because `load_from_config()` takes a config object :)

cliquet/tests/test_cache.py
leplatrem: Or add a helper `def get_backend_prefix()`

cliquet/cache/memory.py
leplatrem: Why ?

cliquet/cache/memory.py
lavish205: @Natim We can't do that because in `def expire()` we are again modifying the key. If we do it before calling the `self.expire()` then `key` will be modified to `prefix_prefix_key`

cliquet/tests/test_cache.py
lavish205: I didn't get you.

cliquet/cache/memory.py
Natim: Good point :+1:

cliquet/tests/test_cache.py
lavish205: @Natim can you tell me how to validate that redis is always called with the right key?

cliquet/tests/test_cache.py
```diff
@@ -15,7 +15,7 @@
 
 class CacheBaseTest(unittest.TestCase):
     def setUp(self):
-        self.cache = CacheBase()
+        self.cache = CacheBase(cache_prefix='')
```
leplatrem: Should be `prefix=''` here

cliquet/tests/test_cache.py
leplatrem: The idea is to use two instances, one with prefix and the other one without.

Natim: This idea is really good, but doesn't work with the MemoryBackend because they are not sharing the same store.

lavish205:
@leplatrem what to update in configuration documentation?

Natim:
@lavish205 I believe it is just about adding the ``cache_prefix`` configuration settings as you told me on IRC.

> 13:52:43 <lavish_> do I need to add prefix changes here?
> http://kinto.readthedocs.org/en/latest/configuration/settings.html#cache

lavish205:
@Natim what exactly I have write under "What does it do?"  column ?

Natim:
@lavish205 Something like: ``It add a prefix to cache key names so that multiple kinto server can use the same cache database without disturbing each other.``

CHANGELOG.rst
```diff
@@ -10,6 +10,7 @@ This document describes changes between each past release.
 **New features**
 
 - Default console log renderer now has colours (#671)
+- ``Prefix`` key option added for cache.
```
leplatrem: The setting is called `cache_prefix` no?

Natim:
I have added some documentation and a test to make sure ``cache.get`` added the prefix.

Thank you @lavish205 for your work on this.

leplatrem:
Congrats @lavish205 for keeping the faith and going through all this!

Thank you for getting involved with patience !

lavish205:
Thanks to both of you @leplatrem @Natim :)

PR #681, opened 2016-03-09
[POC] Generic resource docs

leplatrem:
I rebased this old branch where I transform the resource documentation into a Jinja template.

This way it can be inherited and instantiated (for example to document bucket and collections in Kinto docs)

This is a proof-of-concept and can be closed. We might obtain better results with docbox or whateveer proper api documentation tool...

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,5 @@
 .tox/
 loadtests/.env.install
 cliquet_docs/_build/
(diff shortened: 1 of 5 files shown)


Natim:
Now that kinto and cliquet have merged we don't need generic resource doc anymore.

PR #690, opened 2016-04-05
data-->'field' break with numeric comparison.

Natim:
Refs Kinto/kinto#534

This is a really important bug which means sorting and filtering doesn't work with numeric values and the postgresql backend.

We probably have to write new tests to cover this.

### DIFF
--- a/cliquet/storage/postgresql/__init__.py
+++ b/cliquet/storage/postgresql/__init__.py
@@ -590,24 +590,9 @@ class Storage(StorageBase):
                 # JSON operator ->> retrieves values as text.
                 # If field is missing, we default to ''.
                 sql_field = "coalesce(data->>:%s, '')" % field_holder
-
-                # Handle the special cases of numeric values
-                # Digits starting with a 0 are probably strings.
-                # (e.g phone numbers) and Boolean are not digits
-                if not six.text_type(value).startswith('0') and \
+                if isinstance(value, (int, float)) and \
                    value not in (True, False):
-                    try:
-                        value = int(value)
-                        sql_field = "(data->>:%s)::numeric" % field_holder
-                    except ValueError:
-                        try:
-                            value = float(value)
-                            sql_field = "(data->>:%s)::numeric" % field_holder
-                        except ValueError:
-                            pass
-                    except TypeError:
-                        # For list and tuples
-                        pass
+                    sql_field = "(data->>:%s)::numeric" % field_holder
 
             if filtr.operator not in (COMPARISON.IN, COMPARISON.EXCLUDE):
                 # For the IN operator, let psycopg escape the values list.
--- a/cliquet/storage/postgresql/__init__.py
+++ b/cliquet/storage/postgresql/__init__.py
@@ -586,17 +586,28 @@ class Storage(StorageBase):
                 # Safely escape field name
                 field_holder = '%s_field_%s' % (prefix, i)
                 holders[field_holder] = filtr.field
-                if filtr.operator in (COMPARISON.LT, COMPARISON.MAX,
-                                      COMPARISON.MIN, COMPARISON.GT):
+
+                # JSON operator ->> retrieves values as text.
+                # If field is missing, we default to ''.
+                sql_field = "coalesce(data->>:%s, '')" % field_holder
+
+                # Handle the special cases of numeric values
+                # Digits starting with a 0 are probably strings.
+                # (e.g phone numbers) and Boolean are not digits
+                if not six.text_type(value).startswith('0') and \
+                   value not in (True, False):
                     try:
-                        value = float(value)
-                        sql_field = "data->:%s" % field_holder
+                        value = int(value)
+                        sql_field = "(data->>:%s)::numeric" % field_holder
                     except ValueError:
-                        sql_field = "coalesce(data->>:%s, '')" % field_holder
-                else:
-                    # JSON operator ->> retrieves values as text.
-                    # If field is missing, we default to ''.
-                    sql_field = "coalesce(data->>:%s, '')" % field_holder
+                        try:
+                            value = float(value)
+                            sql_field = "(data->>:%s)::numeric" % field_holder
+                        except ValueError:
+                            pass
+                    except TypeError:
+                        # For list and tuples
+                        pass
 
             if filtr.operator not in (COMPARISON.IN, COMPARISON.EXCLUDE):
                 # For the IN operator, let psycopg escape the values list.
--- a/cliquet/storage/postgresql/__init__.py
+++ b/cliquet/storage/postgresql/__init__.py
@@ -578,12 +578,6 @@ class Storage(StorageBase):
         for i, filtr in enumerate(filters):
             value = filtr.value
 
-            try:
-                value = float(value)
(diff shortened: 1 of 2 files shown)


cliquet/storage/postgresql/__init__.py
```diff
@@ -578,6 +578,12 @@ def _format_conditions(self, filters, id_field, modified_field,
         for i, filtr in enumerate(filters):
             value = filtr.value
 
+            try:
+                value = float(value)
```
Natim: This autocast is not good enough because it fails with numeric values that aren't like phone numbers.

cliquet/tests/test_storage.py
```diff
@@ -369,6 +369,17 @@ def test_get_all_can_filter_with_list_of_values(self):
...
+    def test_get_all_can_filter_with_list_of_numeric_values(self):
+        for l in [1, 10, 6, 46]:
+            self.create_record({'code': l})
+        sorting = [Sort('code', 1)]
+        filters = [Filter('code', 10, utils.COMPARISON.MAX)]
+        records, _ = self.storage.get_all(sorting=sorting, filters=filters,
+                                          **self.storage_kw)
+        self.assertEqual(records[0]['code'], 1)
+        self.assertEqual(records[1]['code'], 6)
+        self.assertEqual(records[2]['code'], 10)
```
magopian: You're not testing that the length of the list of records is only 3 (and thus not returning 46)

cliquet/tests/test_storage.py
```diff
@@ -894,6 +905,20 @@ def test_support_sorting_on_deleted_field_groups_deleted_at_first(self):
         self.assertIn('deleted', records[1])
         self.assertNotIn('deleted', records[2])
 
+    def test_sorting_on_numeric_arbitrary_field(self):
```
magopian: you're already testing the sort on line 377 right?

magopian:
Do we run those tests on the postgresql backend at the moment? Do we need to test all the supported backend (for the unit and/or functional tests)?

Natim:
We run them on all backends at the moment.

cliquet/tests/test_storage.py
Natim: I one way here I am testing in the reverse order.

cliquet/tests/test_storage.py
Natim: Yes let me do that.

magopian:
r+, thanks!

cliquet/storage/postgresql/__init__.py
```diff
@@ -586,9 +586,17 @@ def _format_conditions(self, filters, id_field, modified_field,
...
-                # JSON operator ->> retrieves values as text.
-                # If field is missing, we default to ''.
-                sql_field = "coalesce(data->>:%s, '')" % field_holder
+                if filtr.operator in (COMPARISON.LT, COMPARISON.MAX,
+                                      COMPARISON.MIN, COMPARISON.GT):
+                    try:
+                        value = float(value)
+                        sql_field = "data->:%s" % field_holder
+                    except ValueError:
+                        sql_field = "coalesce(data->>:%s, '')" % field_holder
```
leplatrem: This is not tested :) (conversion to float and ValueError)

cliquet/storage/postgresql/__init__.py
```diff
@@ -586,9 +586,17 @@ def _format_conditions(self, filters, id_field, modified_field,
...
+                else:
+                    # JSON operator ->> retrieves values as text.
+                    # If field is missing, we default to ''.
+                    sql_field = "coalesce(data->>:%s, '')" % field_holder
```
leplatrem: I would rather move that part above. And do something like:

```python
+                    try:
+                        value = float(value)
+                        sql_field = "data->:%s" % field_holder
+                    except ValueError:
+                        pass
```

cliquet/storage/postgresql/__init__.py
```diff
@@ -586,9 +586,17 @@ def _format_conditions(self, filters, id_field, modified_field,
                 # Safely escape field name
                 field_holder = '%s_field_%s' % (prefix, i)
                 holders[field_holder] = filtr.field
-                # JSON operator ->> retrieves values as text.
-                # If field is missing, we default to ''.
-                sql_field = "coalesce(data->>:%s, '')" % field_holder
+                if filtr.operator in (COMPARISON.LT, COMPARISON.MAX,
+                                      COMPARISON.MIN, COMPARISON.GT):
```
leplatrem: Why do we restrict to those? Please add comment

cliquet/tests/test_storage.py
```diff
@@ -369,6 +369,29 @@ def test_get_all_can_filter_with_list_of_values(self):
                                           **self.storage_kw)
         self.assertEqual(len(records), 2)
 
+    def test_get_all_can_filter_with_list_of_numeric_values(self):
+        for l in [1, 10, 6, 46]:
+            self.create_record({'code': l})
+        sorting = [Sort('code', 1)]
+        filters = [Filter('code', 10, utils.COMPARISON.MAX)]
```
leplatrem: You might want to provide a number as string here then ?

leplatrem:
Very good catch ! Quite surprising we didn't cross it earlier (e.g. readinglist status filtering)

cliquet/storage/postgresql/__init__.py
Natim: It is here: https://github.com/mozilla-services/cliquet/pull/690/commits/6d6ca311abf7d64127504f1804f2943293a1fe63

cliquet/storage/postgresql/__init__.py
Natim: ok

cliquet/storage/postgresql/__init__.py
Natim: Because for the other comparison method it works the same with strings or number.

Also note that doing this ``value = float(value)`` adds problems for phone numbers for instance.

cliquet/storage/postgresql/__init__.py
Natim: Actually it is what I did before and I reverted because doing this ``value = float(value)`` adds problems for phone numbers like ``0123456789`` for instance.

cliquet/storage/postgresql/__init__.py
```diff
@@ -586,9 +586,13 @@ def _format_conditions(self, filters, id_field, modified_field,
                 # Safely escape field name
                 field_holder = '%s_field_%s' % (prefix, i)
                 holders[field_holder] = filtr.field
+
                 # JSON operator ->> retrieves values as text.
                 # If field is missing, we default to ''.
                 sql_field = "coalesce(data->>:%s, '')" % field_holder
+                if isinstance(value, (int, float)) and \
+                   value not in (True, False):
+                    sql_field = "(data->>:%s)::numeric" % field_holder
```
leplatrem: Much better than the previous version!

@Natim it makes me think: couldn't we have converted the `value` to string instead? (*not for sorting I know*)

Natim: I don't get your question, it was converted to string before right?

PR #694, opened 2016-04-12
Fix changelog about cache_prefix

leplatrem:
The changelog entry was misplaced because the PR #680 was not rebased before merging ;)

As a consequence the cliquet release 3.1.1 should have been tagged 3.2.0 instead :) 

r? @magopian 

capture d ecran de 2016-04-12 16-21-29


Natim:
r+

Issue #687, opened 2016-03-18
Move this repository to the Kinto github organisation and rename it kinto-core


Natim:
The more I think about it the more it makes sense to me to do that transition at some point.

However we are not in a hurry. I am just adding a ticket there as we already started to think about it.

glasserc:
I don't think I know Cliquet well enough to have a valid opinion, but cleaning up the documentation depends on this issue, so I'll offer my opinion.

I think that something like Cliquet could be a valuable abstraction/utility for Python services. Cliquet gives the Mozilla services team a "beachhead" from which to enforce common requirements/practices in Python-based services -- things like logging, monitoring (heartbeats? metrics?), exception handling (if we call another service and it fails, can we include information about the original failure in our failure message?), versioning. I have no idea if the services team is going to make other Python-based services, but I can't imagine that Kinto is going to be the only one.

I don't know if Cliquet, as it stands today, is the library/framework that I describe above. It may be that the current state of the project is too closely bound to Kinto's requirements.

To help figure out the relationship between Kinto and Cliquet: if Kinto needs something in Cliquet to change, does the Services team try to balance that against the needs of other hypothetical services? Or does it just make the change?

In my opinion, there are two good options:

1. Cliquet is really just part of Kinto. This might be true if Cliquet's been too infected by Kinto's requirements. In this case, I think we should just merge the Cliquet and Kinto repositories. There's no value in keeping them in separate repositories -- it just makes things more inconvenient.

2. Cliquet is a separate library. Even if Cliquet today isn't perfect, maybe it's possible to merge parts of it into Kinto and keep the rest as a standalone library, or even just a set of documentation describing the best practices that we think services should implement. In this case I think we should largely keep Cliquet as it is (as a separate repository), although of course there will be code that needs to move back and forth until we find the correct boundary.

My personal belief is that the second possibility is more likely, although it's hard to be sure since only one real project uses Cliquet so far, and it's impossible to be confident in a library without a lot of clients.

Natim:
The second possibility is what we aim for.

We actually have multiple projects that already uses Cliquet:

- [Readinglist](http://readinglist.readthedocs.io/en/latest/)
- [Kinto](http://kinto.readthedocs.io/en/latest/)
- [Syncto](http://syncto.readthedocs.io/en/latest/)

The idea is that you should use Kinto for your service until you have too many business rules that you need to process on server side and then you will use Cliquet to create this service.

glasserc:
[...]

PR #696, opened 2016-04-27
Fix some typos

glasserc:


cliquet_docs/api/resource.rst
```diff
@@ -876,7 +876,7 @@ In a response, ``permissions`` contains the current permissions of the record
 
 .. note::
 
-    When a record is created or modified, the current :term`user id`
+    When a record is created or modified, the current :term:`user id`
     **is always added** among the ``write`` principals.
 
-`Read more about leveraging resource permissions <resource-permissions>`.
+:ref:`Read more about leveraging resource permissions <permissions>`.
```
Natim: Just a warning that ``cliquet_docs`` is used in projects that uses it such as [Kinto](https://github.com/Kinto/kinto/blob/master/docs/conf.py#L31-L49) 

We need to make sure that if we change ``resource-permissions`` into ``permissions`` here, it doesn't breaks the references in the kinto docs as well.

I checked and it doesn't seems to be the case.

Natim:
LGTM

leplatrem:
Thanks Ethan!

cliquet_docs/api/resource.rst
glasserc: Well, the link is broken right now anyhow (see the end of [this page](http://kinto.readthedocs.org/en/latest/api/1.x/cliquet/resource.html#notes-on-permissions-attribute)), so this commit doesn't make things any worse.

I tried it both ways (after disabling the re-copy part of `docs/conf.py` by hand), and it's a dead link either way, because both `resource-permissions` and `permissions` point to the `reference` directory of the Cliquet docs (to `resource.rst` and `permission.rst`, respectively), which Kinto doesn't copy.

I think that the way Kinto copies Cliquet docs is confusing and strange and I think it needs to be rethought, but that's outside the scope of this PR.

leplatrem: > I think that the way Kinto copies Cliquet docs is confusing and strange and I think it needs to be rethought, but that's outside the scope of this PR.

Definitely agree!

A year ago we didn't know that Kinto was going to be the *star* product. This needs to be cleaned-up and rethought. (Related: https://github.com/mozilla-services/cliquet/issues/687)

PR #698, opened 2016-04-28
Fix migration of triggers in PostgreSQL storage backend …

leplatrem:
Migrations from cliquet<3.0.0 were broken.

Tests were testing that migrations did not loose any records. They now test create + delete.

Fixes Kinto/kinto#559

@Natim r?


Natim:
r+

PR #700, opened 2016-05-02
Missing end slash on hello page returns a 404. Fixes Kinto/kinto#544

Natim:
r? @leplatrem

### DIFF
--- a/cliquet/views/errors.py
+++ b/cliquet/views/errors.py
@@ -59,26 +59,26 @@ def page_not_found(request):
 
     querystring = request.url[(request.url.rindex(request.path) +
                                len(request.path)):]
-    resource_missing = (
-        request.path.startswith('/' + request.registry.route_prefix + '/') or
-        not trailing_slash_redirection_enabled
-    )
+
+    errno = ERRORS.MISSING_RESOURCE
+    error_msg = "The resource you are looking for could not be found."
 
     if not request.path.startswith('/' + request.registry.route_prefix):
         errno = ERRORS.VERSION_NOT_AVAILABLE
         error_msg = ("The requested protocol version is not available "
                      "on this server.")
-    elif request.path.endswith('/') and trailing_slash_redirection_enabled:
-        path = request.path.rstrip('/')
-        redirect = '%s%s' % (path, querystring)
-        return HTTPTemporaryRedirect(redirect)
-    elif resource_missing:
-        errno = ERRORS.MISSING_RESOURCE
-        error_msg = "The resource you are looking for could not be found."
     elif trailing_slash_redirection_enabled:
-        # Case for /v0 -> /v0/
-        redirect = '/%s/%s' % (request.registry.route_prefix, querystring)
-        return HTTPTemporaryRedirect(redirect)
+        redirect = None
+
+        if request.path.endswith('/'):
+            path = request.path.rstrip('/')
+            redirect = '%s%s' % (path, querystring)
+        elif request.path == '/' + request.registry.route_prefix:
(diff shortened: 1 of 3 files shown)


cliquet/views/errors.py
```diff
@@ -67,9 +67,14 @@ def page_not_found(request):
...
+    elif request.path.startswith('/' + request.registry.route_prefix + '/'):
         errno = ERRORS.MISSING_RESOURCE
         error_msg = "The resource you are looking for could not be found."
+    elif trailing_slash_redirection_enabled:
+        querystring = request.url[(request.url.rindex(request.path) +
+                                   len(request.path)):]
+        redirect = '/%s/%s' % (request.registry.route_prefix, querystring)
+        return HTTPTemporaryRedirect(redirect)
 
     response = http_error(httpexceptions.HTTPNotFound(),
```
glasserc: Is it possible to get through all these ifs without setting `errno` or `error_msg`?

Natim: There is apparently one case where you do not enter any ifs:

    request.path.startswith('/' + request.registry.route_prefix) and \
    not request.path.startswith('/' + request.registry.route_prefix + '/') and \
    not trailing_slash_redirection_enabled

In that specific case we do not define ``errno`` and ``error_msg``

Either I can change the ``elif`` on line 73 with a ``else`` making the redirection for ``/v1`` to ``/v1/`` even when trailing slash redirection is not enabled or I can add a else with ``errno`` et ``error_msg``.

Natim: Fixed with https://github.com/mozilla-services/cliquet/pull/700/commits/ad185c919ebf9046286777bc58c3025080024521

cliquet/views/errors.py
```diff
@@ -67,9 +67,15 @@ def page_not_found(request):
                                    len(request.path)):]
         redirect = '%s%s' % (path, querystring)
         return HTTPTemporaryRedirect(redirect)
-    elif request.path.startswith('/' + request.registry.route_prefix):
+    elif (request.path.startswith('/' + request.registry.route_prefix + '/') or
+          not trailing_slash_redirection_enabled):
```
leplatrem: nit: as a rule of thumb, never write a if on several lines. Please use intermediary variables :)

cliquet/views/errors.py
```diff
@@ -67,9 +72,15 @@ def page_not_found(request):
...
         redirect = '%s%s' % (path, querystring)
         return HTTPTemporaryRedirect(redirect)
-    elif request.path.startswith('/' + request.registry.route_prefix):
+    elif resource_missing:
         errno = ERRORS.MISSING_RESOURCE
         error_msg = "The resource you are looking for could not be found."
+    elif trailing_slash_redirection_enabled:
+        # Case for /v0 -> /v0/
+        querystring = request.url[(request.url.rindex(request.path) +
+                                   len(request.path)):]
```
leplatrem: querystring can defined only once

cliquet/views/errors.py
```diff
@@ -57,19 +57,28 @@ def page_not_found(request):
     redirect_enabled = request.registry.settings[config_key]
     trailing_slash_redirection_enabled = asbool(redirect_enabled)
```
leplatrem: this is unrelated, but `trailing_slash_redirect_enabled` is not documented in settings docs

Natim: I have added that here: https://github.com/Kinto/kinto/pull/583
I realized that a lot are missing in cliquet docs too.

leplatrem:
r+

PR #701, opened 2016-05-03
Fix crash when setting empty permission list with PostgreSQL permission backend (fixes Kinto/kinto#575)

leplatrem:
@Natim r?


Natim:
r+

PR #703, opened 2016-05-09
Rename to kinto-core

glasserc:
This PR represents the first step in addressing #687. It's basically a giant search-and-replace for the string "cliquet".

The badges in README.rst and links to related tools like Travis and Coveralls remain with the old name until we have them set up under the new name. The changelog also uses the name Cliquet to refer to versions before the rename.

I'm especially soliciting feedback for 60f3514dcd0446fe48b41feb37e4b83365593d9b.

r? @almet @Natim @leplatrem


glasserc:
Based on https://github.com/mozilla-services/cliquet/issues/687#issuecomment-217961544, I have also pushed 69c87600436a7e1ee854c5c59837dc1a4b0cc3f. It obviously looks very strange, but maybe it's correct?

Natim:
I think you should replace cliquet with kinto in there.

kinto_core/__init__.py
```diff
@@ -38,29 +38,29 @@
...
-        'cliquet.initialization.setup_cache',
-        'cliquet.initialization.setup_requests_scheme',
-        'cliquet.initialization.setup_version_redirection',
-        'cliquet.initialization.setup_deprecation',
-        'cliquet.initialization.setup_authentication',
-        'cliquet.initialization.setup_backoff',
-        'cliquet.initialization.setup_statsd',
-        'cliquet.initialization.setup_listeners',
-        'cliquet.events.setup_transaction_hook',
+        'kinto_core.initialization.setup_request_bound_data',
```
Natim: I think all this should be ``kinto`` rather than ``kinto_core`` or ``cliquet``

leplatrem: These are python paths, not setting names

glasserc:
OK, so I've been looking at the history of the code and I'm more confused than ever. First, N.B. that we expose a `cliquet_protocol_version` which is marked as "internal", so the fact that we change that to `kinto_core_protocol_version` is fine and doesn't require any special effort. Originally we exposed a `batch_max_requests` setting as `cliquet.batch_max_requests` (https://github.com/mozilla-services/cliquet/issues/318), but then we decided that it would be better if the namespace was determined by the configured project name (https://github.com/mozilla-services/cliquet/issues/448). We expose these "public settings" in the `hello` view with this "backwards compatibility" hack, but it isn't part of the specification, so I think it's meant as a backwards compatibility hack for old client libraries.

We have to release a major version update anyhow, so I think breaking compatibility is OK. I'd even vote for getting rid of the hack entirely, but that might be outside the scope of this PR. If we can't do that, then I'm OK with either changing it or not changing it.

Natim:
The work about that is happening here: https://github.com/glasserc/kinto/pull/1

PR #704, opened 2016-05-10
Fix crash when type of value in exclude/include is wrong (fixes Kinto/kinto#587)

leplatrem:
@Natim r?


glasserc:
Isn't this silent failure? I almost prefer the original behavior..

Natim:
Yes why not doing a cast to the correct type in that case?

leplatrem:
> Isn't this silent failure? I almost prefer the original behavior..

You're right, we shall return a 400. I'll rework the patch.

Natim:
Would you add a test to make sure the description of the error is the one you are looking for?

Natim:
It looks good to me

PR #709, opened 2016-05-18
If-None-Match:* with GET and existing record should raise a 412 Precondition Failed

Natim:
Fixes Kinto/kinto#592.

### DIFF
--- a/cliquet/resource/__init__.py
+++ b/cliquet/resource/__init__.py
@@ -723,14 +723,14 @@ class UserResource(object):
                 raise ValueError()
             modified_since = int(if_none_match[1:-1])
         except (IndexError, ValueError):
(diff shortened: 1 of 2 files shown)


cliquet/resource/__init__.py
```diff
@@ -729,6 +729,8 @@ def _raise_304_if_not_modified(self, record=None):
                     'description': "Invalid value for If-None-Match"
                 }
                 raise_invalid(self.request, **error_details)
+            else:
+                return
```
leplatrem:
```py
            if if_none_match == '*':
                return
            error_details = {
                'location': 'headers',
                'description': "Invalid value for If-None-Match"
            }
```

PR #711, opened 2016-05-18
Indicate that Cliquet is now deprecated

glasserc:
Following what I saw at https://github.com/citusdata/pg_shard, I added a deprecation notice to the README. Also, following http://stackoverflow.com/questions/24085400/how-to-formally-deprecate-a-pip-package, add a warning when the package is imported.

We can merge this once the tests pass on my fork of Kinto and I merge that into master.

### DIFF
--- a/cliquet/__init__.py
+++ b/cliquet/__init__.py
@@ -17,7 +17,7 @@ from cliquet.logs import logger
 
 
 import warnings
-warnings.warn("cliquet is now deprecated. Please update to kinto.core at "
+warnings.warn("Cliquet is now deprecated. Please update to kinto.core at "
               "https://github.com/Kinto/kinto. See "
               "https://github.com/mozilla-services/cliquet/issues/687 "
               "for more information.", DeprecationWarning)
--- a/cliquet/__init__.py
+++ b/cliquet/__init__.py
@@ -17,11 +17,16 @@ from cliquet.logs import logger
 
 
 import warnings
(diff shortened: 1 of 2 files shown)


cliquet/__init__.py
```diff
@@ -16,6 +16,13 @@
 from cliquet.logs import logger
 
 
+import warnings
+warnings.warn("cliquet is now deprecated. Please update to kinto.core at "
```
Natim: nit: Cliquet with a capital C.

Natim:
r+ (I won't ask for a test to make sure the warning is displayed.)

cliquet/__init__.py
glasserc: I wasn't sure because the package name is `cliquet`, but the library's name is `Cliquet`.

Natim:
Actually while running cliquet with this code, I do not see the warning.

cliquet/__init__.py
```diff
@@ -16,6 +16,13 @@
 from cliquet.logs import logger
 
 
+import warnings
+warnings.warn("Cliquet is now deprecated. Please update to kinto.core at "
```
Natim: It seems that ``DeprecationWarning`` are ignored by default so we need to run ``python -Walways`` to see them.
Maybe we could use ``logger.warn`` instead or both?

glasserc:
Deprecation warnings are apparently silenced by default :(. Added a call to logger.warn in 8d20e3ba62de7eaa3e5787b1902ff0090325bf66.

